Issue #480, opened 2015-11-06
Support full system backup and bare metal restore


dinoboy197:
It would be immensely helpful to include the ability to have Back In Time backup absolutely everything (grub, system files, programs, user files, etc) and enable bare metal restore.

Maybe its something I could help out with?

Germar:
Help is always very welcome! :+1: 

BIT only works on file level. So to backup grub we'd need an image first (e.g. with `dd`). This can be done with a `user-callback` script. Take a look at [bit-team/user-callback](https://github.com/bit-team/user-callback) for examples.

All the rest can be done with BIT already. For sure, there is no One-Click restore feature yet. You'd need to create partitions and manually write back grub.

There is also bug #58 making troubles when restoring from LiveCD.

dinoboy197:
> BIT only works on file level. So to backup grub we'd need an image first (e.g. with dd). This can be done with a user-callback script. Take a look at bit-team/user-callback for examples.

Got it, makes sense. I'm going to try to use BIT for the first time in a couple of days (excited!) and then will report back. Maybe we can come up with some steps involved in putting together some bare metal restore ability.

Germar:
That would be great!

Here's a list on what would need to be done for full featured bare metal restore (feel free to extend that):
1. boot-loader backup and restore
2. partition-table backup and recreation of partitions
3. format partitions (optional)
4. rebuild `/etc/fstab` if `UUID`s where in use
5. ability to load different `/etc/passwd` and `/etc/group` to fix bug #58

dinoboy197:
Awesome, thanks. I've been busy with other things to start on this - but look forward to taking a swing sometime. Your list of items sounds perfect - that's what I was thinking. Will circle back soon.

dinoboy197:
OK, I've been doing some brainstorming and realize that there are many. many potential features that could be included in such a project.

So that this does not become a huge single herculean effort (with final value delivery only at the very end), I thought it might be good to break the work up into phases, delivering some user value with each feature. I'm putting together a quick phase document now; will share in the next day or two for you to review. Stay tuned!

Germar:
Sounds great! :+1:

dinoboy197:
OK, [take a look at this](https://docs.google.com/document/d/14ZjPnf_lwxY11Me4OZzt8sV4Tlm45jMdXhxpqsgcVoM/edit#). Hopefully this will start to bring whatever ideas we have in both of our heads into alignment (they might be quite different at the moment!)

[...]

PR #506, opened 2015-12-18
Add button to settings panel to modify profile for full system backup #480

dinoboy197:
As a first step of [the bare metal backup / restore plan](https://docs.google.com/document/d/14ZjPnf_lwxY11Me4OZzt8sV4Tlm45jMdXhxpqsgcVoM/edit#), I've added a button to the BIT settings GUI to adjust the current profile (whatever that may be) to have settings appropriate for performing a full system backup of all files. Note that this button simply changes existing values of the profile at the moment it is tapped; the user can still modify the changed values afterwards in case they want to tinker with the settings.

Profile adjustments include:
* Setting full rsync mode to true
* Setting continue on errors to false (so any failures doesn't lure user into false sense of security about the consistency of the backup)
* Setting exclude by size to false (again, so that all files are backed up no matter what)
* Setting backup on restore to false (right now, this is only use for restoring an entire set of files from a working installation and that would be many, many files to backup and rename)
* Setting use checksum to false (triggers lots of I/O for checking files otherwise)
* Setting preserve ACL and xattrs to true (some system / user files might rely on these for a bootable / usable system)
* Setting copy links to false (some system / user files might rely on symbolic links being intact for a bootable / usable system)
* Set file include to root "/" (make sure to take all files in the backup snapshot)

Notes:
* I decided not to use a "create full system backup profile button" because for new beginning users of BIT who want to use this feature, the "Main Profile" would then be a secondary, extra profile which would never be used and could be a source of confusion ("why two profiles?")
* This is just the first step in the full system backup / restore work; it does not tackle backing up the bootloader or partitning the restore disk, as noted in the plan linked above. This will happen next.
* I've tested backing up with a profile generated by clicking the button, then restoring the full snapshot to a different disk (with the same partition map, disk uuids, and bootloader) and the restored system works perfectly.
* Modified the CHANGES file
* Added myself into copyright notices (hope you don't mind since I hope to contribute this and more code into master!)


Germar:
Looks great. Thanks :+1: 

I first thought the `Modify for Full System Backup` button would stretch the window, but it doesn't. Same size as before. I really like your idea of altering existing profile instead of making a new one and totally agree with all your adjustments. And of course you are welcome to add yourself to copyright!

There is one `_` missing in settingsdialog.py:891 for translation using gettext. But I'll add that after merge.

Thanks again for your first contribution on BIT!

dinoboy197:
Thanks! Glad that I can help out. Will be in touch soon about more work - holiday may slow me down. Happy holidays!

Germar:
Great. You can also contact me through germar DOT reitze HOSTED ON gmail DOT com.
Happy holidays to you, too!

PR #510, opened 2015-12-30
Add travis CI support

dinoboy197:
This is more of an experiment for fun than anything else! (So you won't make me feel bad if you don't want it for any reason).

I'm a big fan of continuous integration systems for running compilation, tests, etc on every check-in. I've used Cruisecontrol, Hudson, Jenkins, and others for many years and they've been a big help. I've recently been exposed to [TravisCI](https://travis-ci.org/), a free CI service for open-source projects. It's very similar in features to the other big CI systems.

I tried my hand at adding a [TravisCI configuration file](https://docs.travis-ci.com/user/languages/python), and I created a travis CI account and [connected it to my personal BIT fork](https://travis-ci.org/dinoboy197/backintime). So anytime I check in to any of my fork's branches, the CI system builds and runs tests, then emails me if there are any issues.

I've included the travisci configuration file that I'm using.

The one caveat is that the unit tests don't currently pass when executing in TravisCI. I'm not sure if this is because we need to tell travisCI to install some extra library dependencies, or the UnitTest framework isn't picking up the python paths of some of the files correctly, or something else entirely (like me setting something up incorrectly :). It's so close to working! I have the single line which runs unit tests commented out, but am still running the python compiler to verify that the syntax is ok.

Anyway, I thought you might be interested in taking a look. I'm not sure if you'll have access to my travisCI instance, but [here is one of the pages which shows the unit test failures](https://travis-ci.org/dinoboy197/backintime/jobs/99375128). I have the unit tests running in verbose mode to try to debug, but alas I'm not enough of a Python unit test expert to know exactly what might be going on (most of my unit test experience is in Java and Scala!)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,16 @@
+# TravisCI (https://travis-ci.org) configuration file
+# https://docs.travis-ci.com/user/languages/python
+
+language: python
+
+python:
+  - "3.4"
+  - "3.5"
+  - "3.5-dev" # 3.5 development branch
+  - "nightly" # currently points to 3.6-dev
+
+script:
+  # compile all files - ensure that syntax is correct
+  - python -m compileall common common/test common/plugins qt4 qt4/plugins
+  # run unit tests - ensure that functionality is correct
+  - cd common && ./configure && make unittest-v
--- a/common/test/__init__.py
+++ b/common/test/__init__.py
@@ -15,5 +15,11 @@
 # with this program; if not, write to the Free Software Foundation, Inc.,
 # 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 
+from test import test_applicationinstance
+from test import test_argparser
+from test import test_configfile
+from test import test_snapshots
+from test import test_tools
+
 if __name__ == '__main__':
     pass
--- a/common/tools.py
+++ b/common/tools.py
@@ -21,7 +21,6 @@ import sys
 import subprocess
 import signal
 import re
-import dbus
 import errno
 import gzip
 import tempfile
@@ -36,6 +35,18 @@ except:
     keyring = None
     os.putenv('BIT_USE_KEYRING', 'false')
     keyring_warn = True
+
+# getting dbus imports to work in Travis CI is a huge pain
+# use conditional dbus import
+ON_TRAVIS = 'TRAVIS' in os.environ and os.environ['TRAVIS'] == 'true'
+try:
+    import dbus
+except ImportError:
+    if ON_TRAVIS:
+        #python-dbus doesn't work on Travis yet.
+        dbus = None
+    else:
+        raise
 
 import configfile
 import logger
@@ -1173,6 +1184,9 @@ class SetupUdev(object):
     INTERFACE = 'net.launchpad.backintime.serviceHelper.UdevRules'
     MEMBERS = ('addRule', 'save', 'delete')
     def __init__(self):
+        if dbus is None:
+            self.isReady = False
+            return
         try:
             bus = dbus.SystemBus()
             conn = bus.get_object(SetupUdev.CONNECTION, SetupUdev.OBJECT)


dinoboy197:
Here's the kind of benefit that would be had if this were working and was merged to master:

Any future pull requests would automatically be built, and the pull request page [would show this stuff](http://stackoverflow.com/a/27373734/474043).

Germar:
This is really cool and I definitely want this! (Almost feels like late Christmas :blush:)

[This error](https://travis-ci.org/dinoboy197/backintime/jobs/99375147#L406) looks like Travis trying to `import test` and then failing to reference `test.test_applicationinstance`. This would work if you add this to `common/test/__init__.py`:

```
from test import test_applicationinstance
from test import test_argparser
from test import test_configfile
from test import test_snapshots
from test import test_tools
```

Germar:
Hey, I just saw you struggling with dbus (looks like you're on [this](https://abdealijk.wordpress.com/2015/06/09/python-dbus-and-continuous-integration/) path). Did you try calling `dbus-launch python...` for the tests?

`common/configure`:
```
printf "\tdbus-launch python${PYVERSION} -m unittest ${v} -b $i\n" >> ${MAKEFILE}
```

Germar:
There's one `$` to much. Also better use `make unittest-v` instead of `make test-v` as the integrationtests might never work (I don't like them either and will rewrite them soon as unittests)

Germar:
Sorry, need to go sleep. Highly appreciate your work! Good luck!

dinoboy197:
Yeah, it's kinda crazy how much extra work is required because python-dbus isn't pip compatible! :(

Unfortunately, even after following [those instructions](https://abdealijk.wordpress.com/2015/06/09/python-dbus-and-continuous-integration/), compiling dbus, dbus-glib, python-dbus, then copying over the python-dbus libs into the virtualenv that is set up in Travis, dbus only partially initializes.

[Latest build error is here](https://travis-ci.org/dinoboy197/backintime/jobs/99553789).

Any more python import magic up your sleeve for this one? :) Some others on the internet report similar issues, without any solution ([here](https://discuss.mopidy.com/t/python-errors-trying-to-run-mopidy/1021/2), [here](http://blender.stackexchange.com/questions/39904/how-can-i-import-or-use-dbus-in-my-python-script-in-blender?newreg=c7fb0e4586bd4d7b86c1361e75a920ed)) I'm a bit afraid that this one may not be solved by brute force, since we're really mucking around with python imports here.

Germar:
Happy New Year!

At the moment there is no unittest using `dbus` and even if we add tests using `dbus` later on we can skip them on Travis. So before we will end up with no Travis support at all I'd suggest we just let `import dbus` silently fail on Travis.

```
index e86b950..0b3d875 100644
--- a/common/tools.py
+++ b/common/tools.py
@@ -21,7 +21,6 @@ import sys
 import subprocess
 import signal
 import re
-import dbus
 import errno
 import gzip
 import tempfile
@@ -36,6 +35,16 @@ except:
     keyring = None
     os.putenv('BIT_USE_KEYRING', 'false')
     keyring_warn = True
+ON_TRAVIS = 'TRAVIS' in os.environ and os.environ['TRAVIS'] == 'true'
+try:
+    import dbus
+except ImportError:
+    if ON_TRAVIS:
+        #python-dbus doesn't work on Travis yet.
+        dbus = None
+        pass
+    else:
+        raise
 
 import configfile
 import logger
@@ -1173,6 +1182,9 @@ class SetupUdev(object):
     INTERFACE = 'net.launchpad.backintime.serviceHelper.UdevRules'
     MEMBERS = ('addRule', 'save', 'delete')
     def __init__(self):
+        if dbus is None:
+            self.isReady = False
+            return
         try:
             bus = dbus.SystemBus()
             conn = bus.get_object(SetupUdev.CONNECTION, SetupUdev.OBJECT)
```

dinoboy197:
Sounds good. I was going to ask you the same, but wasn't sure if you'd be comfortable with optional imports :)

dinoboy197:
And Happy 2016 to you too!

dinoboy197:
[Build is green](https://travis-ci.org/dinoboy197/backintime/builds/99768580) and I've rebased down to one commit, also rebased against upstream/master. Ready to ship if it looks ok to you.

Germar:
Thanks a lot! This is a real cool new feature!

PR #513, opened 2016-01-03
Store filesystem UUID mount points in info file during snapshot

dinoboy197:
Part of #480 

This will help future efforts to restore data to physical devices by storing the current filesystem mount structure for later reconstruction during.

Addition to the info file will look like this, for example:
```
filesystem_mounts={"/": {"original_uuid": "d683a7b1-2e46-4f63-ac73-7154c83c4a73"}, "/home": {"original_uuid": "1349a61f-690c-4137-89c7-e21162b135ff"}}
```

For the first part of this effort, BIT bare metal restore will only restore a snapshot to the same physical devices that it was taken from. However, this is just a stepping stone towards restoring to other devices.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,9 @@
 
 language: python
 
+# ensures that we have UUID filesystem mounts for proper testing
+sudo: true
+
 python:
   - "3.4"
   - "3.5"
--- a/common/snapshots.py
+++ b/common/snapshots.py
@@ -1,5 +1,5 @@
 #	Back In Time
-#	Copyright (C) 2008-2015 Oprea Dan, Bart de Koning, Richard Bailey, Germar Reitze
+#	Copyright (C) 2008-2015 Oprea Dan, Bart de Koning, Richard Bailey, Germar Reitze, Taylor Raack
 #
 #	This program is free software; you can redistribute it and/or modify
 #	it under the terms of the GNU General Public License as published by
@@ -16,6 +16,7 @@
 #	51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 
 
+import json
 import os
 import stat
 import datetime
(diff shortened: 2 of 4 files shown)


Germar:
Wouldn't it be easier to just have one dict `{MOUNTPOINT: UUID}`? I see no need for the encapsulated dict.

This would also make it easy to store it without the need of `json`
```
info_file.set_str_value('filesystem_mounts', ('str:mount_point', 'str:uuid'), list(tools.get_filesystem_mount_info().items()))
```

Germar:
BTW don't get shocked by `snapshots.py` ;)
I'm currently working on this in [branch sid-object](https://github.com/bit-team/backintime/tree/sid-object) and will merge that branch after releasing `1.1.10` next Saturday.

dinoboy197:
> Wouldn't it be easier to just have one dict {MOUNTPOINT: UUID}? I see no need for the encapsulated dict.

Ah, yes, I failed to mention why I did that! I'm future-proofing a little; per some of the plans for supporting new devices during restore, I was planning to store some more info in those nested dicts in the future (possibly device total block size, attribute flags, or other things) to verify that that upon restore, the target filesystems were both large enough and had at least the same attribute flags set to support restoring from a snapshot. Otherwise restoring might fail (not enough device space for one or more FS mounts) or attributes might not copy back and might cause program failure when attempting to access those files later.

It may turn out to be overkill, but if we use the nested dict now, then I don't need to change the key value of the encompassing dict from string to dict.

Does that make sense? Do you think I am "overthinking" this? (Just want to prevent too many incompatible format changes down the line).

dinoboy197:
> BTW don't get shocked by snapshots.py ;)
> I'm currently working on this in branch sid-object and will merge that branch after releasing 1.1.10 next Saturday.

Cool!

Germar:
Oh, sure. That makes sense! :+1:

I'll merge this after next release, okay?

dinoboy197:
> Oh, sure. That makes sense!

Great!

> I'll merge this after next release, okay?

Sounds good - no rush. I'll start some work on the extra restore pieces with this piece going to master in mind.

Germar:
This merge went somehow strange. My fault because I changed copyright date to 2016 while you added yourself to it. I manually merged it but it's now connected to my account. Hope you don't mind...

dinoboy197:
No problem.

In the future, if Github says that it can't merge cleanly, I'll be happy to rebase on my end up to master so that it applies properly without a merge conflict. Just let me know in a PR comment and I'll rebase things up so merge doesn't cause you to do more work!

Germar:
Okay, thanks. 

git and GitHub is still new for me. I merged from Launchpad and bzr couple month ago.

BTW I added source documentation on http://backintime.rtfd.org/
This is build automatically from docstrings. I choose [Google Style](http://sphinxcontrib-napoleon.readthedocs.org/en/latest/example_google.html) docstring format and will add those docstrings as I proceed adding further unittests.

PR #516, opened 2016-01-10
the -i switch will install the built package

graysky2:


Germar:
Would `makepkg` need to be called with `sudo` for this?

graysky2:
No need to invoke as with sudo... makepkg will try to install the build package via `sudo pacman -U ...` but if the user cannot he/she is presented with the standard:
```
We trust you have received the usual lecture from the local System Administrator. It usually boils down to these three things:
...
```

PR #528, opened 2016-01-24
Fix file info persistence errors

### LINKED ISSUE #275: [SOLVED] last fr.po translation in trunk (rev.523) is really wrong
It is solved, I have modified the wrong translations into launchpad. The bug can be closed.
-----------------------------------------------------------------------------------------------------------

I just checked last fr.po fix in trunk for version 0.9.22.1 : it is really wrong !

Please use the current translations in launchpad.

AFAK 0.9.22 is ok except two strings which are not translated.

------------------------------------
Imported from Launchpad using lp2gh.

 * date created: 2009-04-26T05:56:50Z
 * owner: mahikeulbody
 * assignee: danleweb
 * the launchpad url was https://bugs.launchpad.net/bugs/367154

### LINKED ISSUE #523: AssertionError: path is not str type on 'Save permissions'
Hi Germar,

on current git master, my ssh encrypted backup fails on "save permissions" with the following debug output:

```
IINFO: [common/snapshots.py:999 Snapshots._take_snapshot] Save permissions
DEBUG: [common/sshtools.py:258 SSH.unlock_ssh_agent] Private key /home/akki/.ssh/### is already unlocked in ssh agent
DEBUG: [common/encfstools.py:647 Decode.start_process] start 'encfsctl decode' process
DEBUG: [common/encfstools.py:651 Decode.start_process] Call command: encfsctl decode --extpass=backintime-askpass /home/akki/.local/share/backintime/mnt/6B0CA066/mountpoint
  File "/usr/share/backintime/common/backintime.py", line 987, in <module>
    start_app()
  File "/usr/share/backintime/common/backintime.py", line 465, in start_app
    args.func(args)
  File "/usr/share/backintime/common/backintime.py", line 668, in backup
    ret = take_snapshot(cfg, force)
  File "/usr/share/backintime/common/backintime.py", line 77, in take_snapshot
    ret = snapshots.Snapshots( cfg ).take_snapshot( force )
  File "/usr/share/backintime/common/snapshots.py", line 704, in take_snapshot
    ret_val, ret_error = self._take_snapshot( sid, now, include_folders )
  File "/usr/share/backintime/common/snapshots.py", line 1039, in _take_snapshot
    new_snapshot.fileInfo = fileInfoDict
  File "/usr/share/backintime/common/snapshots.py", line 2151, in fileInfo
    assert isinstance(path, str), 'path is not str type: {}'.format(path)
AssertionError: path is not str type: b'/home/akki/Dokumente/Archiv/Korrespondenz'
DEBUG: [common/encfstools.py:761 Decode.close] stop 'encfsctl decode' process
DEBUG: [common/encfstools.py:519 Encode.close] stop 'encfsctl encode' process

```

The path or file name in the last but two line varies with each attempt.
 
This error occurs since commit https://github.com/bit-team/backintime/commit/fa4508ea305978c16193579682653d1900f81000 ; earlier versions work fine...
 
Please let me know if you need any further information...

dinoboy197:
Fixes #523:
1) [```path``` is bytes (and is asserted to be so) when it is set as the key to the fileinfodict](https://github.com/bit-team/backintime/blob/master/common/snapshots.py#L811)
2) [```path``` is still bytes when it's getting ready to be saved, but is now asserted to be a string - which fails instantly](https://github.com/bit-team/backintime/blob/master/common/snapshots.py#L2167).

I fixed a unit test which made an incorrect assumption about the ```fileInfoDict``` keys - it assumed they would be ```String```s from upstream in code in ```_take_snapshot()```, when in fact they are ```Byte``` arrays. Also verified that this fix allows snapshots to complete successfully.

@Germar - Seems like having a big end-to-end test of the snapshot and restore functionality would be good to ensure that the full behavior of BIT (at least the main success cases) always pass. Should we work on building one?

### DIFF
--- a/common/snapshots.py
+++ b/common/snapshots.py
@@ -811,7 +811,7 @@ class Snapshots:
         assert isinstance(path, bytes), 'path is not bytes type: %s' % path
         if path and os.path.exists(path):
             info = os.stat(path)
-            mode = str(info.st_mode).encode('utf-8', 'replace')
+            mode = info.st_mode
             user = self.get_user_name(info.st_uid).encode('utf-8', 'replace')
             group = self.get_group_name(info.st_gid).encode('utf-8', 'replace')
             fileinfo[path] = (mode, user, group)
@@ -1000,7 +1000,7 @@ class Snapshots:
             self.set_take_snapshot_message( 0, _('Save permission ...') )
 
             permission_done = False
-            fileInfoDict = {}
+            fileInfoDict = FileInfoDict()
             if self.config.get_snapshots_mode() in ['ssh', 'ssh_encfs']:
                 path_to_explore_ssh = new_snapshot.pathBackup(use_mode = ['ssh', 'ssh_encfs']).rstrip( '/' )
                 cmd = self.cmd_ssh(['find', path_to_explore_ssh, '-print'])
@@ -1749,6 +1749,19 @@ class Snapshots:
 
         return (' '.join(items1), ' '.join(items2))
 
+class FileInfoDict(dict):
+    """
+    A `dict` that maps a path (as `bytes`) to (`int`, `bytes`, `bytes`).
+    """
+    def __setitem__(self, key, value):
+        assert isinstance(key, bytes), "key '{}' is not bytes instance".format(key)
+        assert isinstance(value, tuple), "value '{}' is not tuple instance".format(value)
+        assert len(value) == 3, "value '{}' does not have 3 items".format(value)
+        assert isinstance(value[0], int), "first value '{}' is not int instance".format(value[0])
+        assert isinstance(value[1], bytes), "second value '{}' is not bytes instance".format(value[1])
+        assert isinstance(value[2], bytes), "third value '{}' is not bytes instance".format(value[2])
+        super(FileInfoDict, self).__setitem__(key, value)
+
 class SID(object):
     """
     Snapshot ID object used to gather all information for a snapshot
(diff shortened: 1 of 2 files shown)


Germar:
Thanks for the pull request! I think it would be best to have fileInfoDict keys and values always contain just `bytes` for `path`, `user` and `group`. `permission` need to stay `int`. Both 'input sources' (`bz2.BZ2File` in `SID.fileInfo` getter and `snapshots._take_snapshot` save permission) return `bytes`. So it would be just logical to keep them `bytes`.

We could subclass `dict` and check all incomming keys and values agains a predefined list of types (add this to tools.py):

```
class InstanceRestrictedDict(dict):
    """
    A `dict` that only accept keys and values of predefined instances.
    Args:
        inst (type, tuple):     a single type or a tuple of types that all keys
                                and values must match
    """
    def __init__(self, inst = (bytes, int)):
        super(InstanceRestrictedDict, self).__init__({})
        self.inst = inst

    def __setitem__(self, key, value):
        assert isinstance(key, self.inst), "key '{}' is not '{}' instance".format(key, self.inst)
        assert isinstance(value, self.inst), "value '{}' is not '{}' instance".format(value, self.inst)
        super(InstanceRestrictedDict, self).__setitem__(key, value)
```

-----------

Yes, we really need a proper end-to-end test. The IGLA students who contributed the basic unittests added this in `common/test/test.sh`. But that bash script is not reliable and doesn't help much on finding the reason of the failure. So I'd like to rewrite this with unittests. At the moment I constantly add unittest for basic functions (and already discovered some bugs). But before next release I want to have some unittest which run the `take_snapshot` and `restore` methods at least on local system. The icing on the cake would be running the same on SSH (localhost). Would be great if you could help on this!

Germar:
Oh, forget about the `InstanceRestrictedDict` class. I just realized that fileInfoDict is not `key=value` but `key=(int, bytes, bytes)`.

```
class FileInfoDict(dict):
    """
    A `dict` that only accept `bytes` keys and values of (`int`, `bytes`, `bytes`).
    """
    def __setitem__(self, key, value):
        assert isinstance(key, bytes), "key '{}' is not bytes instance".format(key)
        assert isinstance(value, tuple), "value '{}' is not tuple instance".format(value)
        assert len(value) == 3, "value '{}' does not have 3 items".format(value)
        assert isinstance(value[0], int), "first value '{}' is not int instance".format(value[0])
        assert isinstance(value[1], bytes), "second value '{}' is not bytes instance".format(value[1])
        assert isinstance(value[2], bytes), "third value '{}' is not bytes instance".format(value[2])
        super(FileInfoDict, self).__setitem__(key, value)
```

dinoboy197:
Sounds good. I added an extra commit (didn't squash them yet) so you can see how things look with your the new ```FileInfoDict```.

If that looks ok to you, let me know before you merge and I'll squash these commits into one (makes for a cleaner repo without my second commit undoing some work from the first :)

common/snapshots.py
```diff
@@ -2161,17 +2174,13 @@ def fileInfo(self):
 
     @fileInfo.setter
     def fileInfo(self, d):
-        assert isinstance(d, dict), 'd is not dict type: {}'.format(d)
+        assert isinstance(d, FileInfoDict), 'd is not FileInfoDict type: {}'.format(d)
         with bz2.BZ2File(self.path(self.FILEINFO), 'wb') as f:
             for path, info in d.items():
-                assert isinstance(path, str), 'path is not str type: {}'.format(path)
```
dinoboy197: I removed these asserts because we're already asserting that ```d``` is a ```FileInfoDict```, and part of the contract of using its API is that its key is ```bytes``` and values are ```(int, bytes, bytes)```.

Germar:
Looks good! :+1: Thanks

dinoboy197:
OK. I've squashed these two commits down to one and rebased against the master branch. This is ready for you to merge into master!

dinoboy197:
Ah, I see that I can merge this in myself now. I'll do that!

PR #529, opened 2016-01-24
Fix file info reading

dinoboy197:
Issue #523 :

Reading fileinfo currently doesn't read the path back in as binary, which is expected by downstream restore code. As a result, when files are restored they don't get their uid / gid / perms set because the paths restored don't exist as keys in the fileinfo dict (the dict keys are strings, but the path checked is bytes).

I've verified that this fix actually works - restored files now get their uid / gid / perms set correctly.

### DIFF
--- a/common/snapshots.py
+++ b/common/snapshots.py
@@ -405,12 +405,11 @@ class Snapshots:
             key_path (bytes):       original path during backup.
                                     Same as in fileInfoDict.
             path (bytes):           current path of file that should be changed.
-            fileInfoDict (dict):    dict with key_path as key (bytes) and
-                                    value of tuple
-                                    (mode(int), username(bytes), groupname(bytes))
+            fileInfoDict (FileInfoDict):    FileInfoDict
         """
         assert isinstance(key_path, bytes), 'key_path is not bytes type: %s' % key_path
         assert isinstance(path, bytes), 'path is not bytes type: %s' % path
+        assert isinstance(fileInfoDict, FileInfoDict), 'fileInfoDict is not FileInfoDict type: %s' % fileInfoDict
         if key_path not in fileInfoDict or not os.path.exists(path):
             return
         info = fileInfoDict[key_path]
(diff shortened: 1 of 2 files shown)


dinoboy197:
Once #528 hits master, I'll rebase this PR and fix up to include the ```FileInfoDict``` just introduced.

dinoboy197:
@Germar - updated and ready for review.

common/snapshots.py
```diff
@@ -2165,7 +2164,7 @@ def fileInfo(self):
                         continue
                     info = line[:index].strip().split(' ')
                     if len(info) == 3:
-                        d[f] = (int(info[0]), info[1], info[2]) #perms, user, group
+                        d[f.encode('utf-8','replace')] = (int(info[0]), info[1].encode('utf-8','replace'), info[2].encode('utf-8','replace')) #perms, user, group
```
Germar: I'd rather remove `.decode('utf-8')` in line 2156, change line 2159 to `line.find(b'/')` and line 2166 to `.split(b' ')`. So we can skip all these `.encode(...)` in here

common/snapshots.py
```diff
@@ -2141,12 +2140,12 @@ def fileInfo(self):
         Load/save "fileinfo.bz2"
 
         Args:
-            d (dict):   dict of: {path: (permission, user, group)}
+            d (FileInfoDict):   FileInfoDict
 
         Returns:
-            dict:       dict of: {path: (permission, user, group)}
+            d:       FileInfoDict
```
Germar: Docstring format (Google style) for Returns should be `type: description`. So `FileInfoDict: dict of: {path: (permission, user, group)}` in this case

Germar:
Everything else looks great. Thanks :+1:

common/snapshots.py
dinoboy197: Ah, thanks! I'm very new to Python formatting so I appreciate the pointer.

common/snapshots.py
dinoboy197: Binary mode in Python is something I'm most certainly not too familiar with, but your explanation makes it easy to see how this works. Thanks!

dinoboy197:
Shipping to master...

common/snapshots.py
Germar: It's pretty much like `str`. You just need to stay with one of each. `b'foo bar'.split(' ')` for example wouldn't work but `b'foo bar'.split(b' ')`. To convert between both you have `str.encode()` and `bytes.decode()`. That's it. But I was also confused by `bytes` after I switched BIT from Python2 to Python3.

Issue #490, opened 2015-11-10
os.access(dirname, os.W_OK) does not recognize unwriteable folders in Snapshot.delete_path.handle_error


colinl:
I am a new user so apologies if this is operator error.
Using Ubuntu 15.10 with BIT 1.1.8 from stable ppa and gnome UI or command line I am unable to delete a snapshot.  From the UI I click Snapshots icon, select the one to be remove and click Delete and it just returns immediately to the snapshots dialog with no sign that anything is happening.  The display is not greyed and even though I left it 30 minutes the snapshot was not deleted.

From the command line I ran
```
backintime --debug remove
```
I expected this to prompt for which one to delete but I see
```
DEBUG: [common/backintime.py:450 arg_parse] Arguments: {'debug': True, 'func': <function remove at 0x7fc571a14d08>, 'command': 'remove'} | unknownArgs: []

Back In Time
Version: 1.1.8

Back In Time comes with ABSOLUTELY NO WARRANTY.
This is free software, and you are welcome to redistribute it
under certain conditions; type `backintime --license' for details.

DEBUG: [common/backintime.py:496 getConfig] config file: /home/colinl/.config/backintime
DEBUG: [common/backintime.py:497 getConfig] profiles: ['1']
DEBUG: [common/pluginmanager.py:88 PluginManager.load_plugins] Register plugin path /usr/share/backintime/plugins
DEBUG: [common/pluginmanager.py:104 PluginManager.load_plugins] Add plugin qt4plugin.py
DEBUG: [common/pluginmanager.py:104 PluginManager.load_plugins] Add plugin userscriptsplugin.py
DEBUG: [common/pluginmanager.py:104 PluginManager.load_plugins] Add plugin notifyplugin.py
DEBUG: [common/sshtools.py:192 SSH.unlock_ssh_agent] Add private key /home/colinl/.ssh/id_rsa to ssh agent
DEBUG: [common/sshtools.py:197 SSH.unlock_ssh_agent] Password available: True
DEBUG: [common/mount.py:298 SSH.mountprocess_lock_acquire] Acquire mountprocess lock /home/colinl/.local/share/backintime/mnt/11889.lock
INFO: [common/mount.py:203 SSH.mount] Mountpoint /home/colinl/.local/share/backintime/mnt/A5868FDC/mountpoint is already mounted
[...]

PR #532, opened 2016-01-28
Fuse mounts for SSH, Local Encrypted, and SSH encrypted are read-only by default

dinoboy197:
@Germar , per our email discussion this is part of the precursor work to moving towards full-rsync mode for all backup modes.

I thought it would be prudent to add read-only for mounts through fuse as a default constructor argument to the ```Mount``` object, in case there is a time in the future when mounting in non-read-only mode is desired. The UI or some other method calling into the ```Mount``` class could pass the read-only flag as they desire.

On testing - I'm not sure what the best way to unit-test or end-to-end test this is at the moment. I think I'd have to set up ssh for the test rig, which I think could be a bigger effort which might be better isolated in a different piece of work? What do you think?

### DIFF
--- a/common/encfstools.py
+++ b/common/encfstools.py
@@ -1,4 +1,4 @@
-#    Copyright (C) 2012-2016 Germar Reitze
+#    Copyright (C) 2012-2016 Germar Reitze, Taylor Raack
 #
 #    This program is free software; you can redistribute it and/or modify
 #    it under the terms of the GNU General Public License as published by
@@ -41,7 +41,7 @@ class EncFS_mount(mount.MountControl):
 
     CHECK_FUSE_GROUP = False
 
-    def __init__(self, cfg = None, profile_id = None, hash_id = None, tmp_mount = False, parent = None, symlink = True, **kwargs):
+    def __init__(self, cfg = None, profile_id = None, hash_id = None, tmp_mount = False, parent = None, symlink = True, read_only = True, **kwargs):
         self.config = cfg
         if self.config is None:
             self.config = config.Config()
@@ -54,6 +54,7 @@ class EncFS_mount(mount.MountControl):
         self.hash_id = hash_id
         self.parent = parent
         self.symlink = symlink
+        self.read_only = read_only
 
         #init MountControl
         super(EncFS_mount, self).__init__()
@@ -68,6 +69,7 @@ class EncFS_mount(mount.MountControl):
         self.setattr_kwargs('password', None, store = False, **kwargs)
         self.setattr_kwargs('hash_id_1', None, **kwargs)
         self.setattr_kwargs('hash_id_2', None, **kwargs)
+        self.setattr_kwargs('read_only', self.read_only, **kwargs)
 
         self.set_default_args()
 
@@ -91,6 +93,8 @@ class EncFS_mount(mount.MountControl):
             encfs += ['--reverse']
         if not self.is_configured():
             encfs += ['--standard']
+        if self.read_only:
+            encfs += ['-o', 'ro']
         encfs += [self.path, self.mountpoint]
         logger.debug('Call mount command: %s'
                      %' '.join(encfs),
@@ -267,7 +271,7 @@ class EncFS_SSH(EncFS_mount):
     Mount / with encfs --reverse.
     rsync will then sync the encrypted view on / to the remote path
     """
-    def __init__(self, cfg = None, profile_id = None, mode = None, parent = None,*args, **kwargs):
+    def __init__(self, cfg = None, profile_id = None, mode = None, parent = None, read_only = True, *args, **kwargs):
         self.config = cfg
         if self.config is None:
             self.config = config.Config()
@@ -279,12 +283,13 @@ class EncFS_SSH(EncFS_mount):
             self.mode = self.config.get_snapshots_mode(self.profile_id)
 
         self.parent = parent
+        self.read_only = read_only
         self.args = args
         self.kwargs = kwargs
 
         self.ssh = sshtools.SSH(*self.args, symlink = False, **self.split_kwargs('ssh'))
         self.rev_root = EncFS_mount(*self.args, symlink = False, **self.split_kwargs('encfs_reverse'))
-        super(EncFS_SSH, self).__init__(*self.args, **self.split_kwargs('encfs'))
+        super(EncFS_SSH, self).__init__(*self.args, read_only = self.read_only, **self.split_kwargs('encfs'))
 
     def mount(self, *args, **kwargs):
         """
--- a/common/mount.py
+++ b/common/mount.py
@@ -1,4 +1,4 @@
-#    Copyright (C) 2012-2016 Germar Reitze
+#    Copyright (C) 2012-2016 Germar Reitze, Taylor Raack
 #
 #    This program is free software; you can redistribute it and/or modify
 #    it under the terms of the GNU General Public License as published by
@@ -30,7 +30,7 @@ from exceptions import MountException, HashCollision
 _=gettext.gettext
 
 class Mount(object):
-    def __init__(self, cfg = None, profile_id = None, tmp_mount = False, parent = None):
+    def __init__(self, cfg = None, profile_id = None, tmp_mount = False, parent = None, read_only = True):
         self.config = cfg
         if self.config is None:
             self.config = config.Config()
@@ -41,6 +41,7 @@ class Mount(object):
 
         self.tmp_mount = tmp_mount
         self.parent = parent
+        self.read_only = read_only
 
         if self.config.get_password_use_cache(self.profile_id):
             pw_cache = password.Password_Cache(self.config)
@@ -82,7 +83,7 @@ class Mount(object):
                     mounttools = self.config.SNAPSHOT_MODES[mode][0]
                     tools = mounttools(cfg = self.config, profile_id = self.profile_id,
                                        tmp_mount = self.tmp_mount, mode = mode,
-                                       parent = self.parent, **kwargs)
+                                       parent = self.parent, read_only = self.read_only, **kwargs)
                     return tools.mount(check = check)
                 except HashCollision as ex:
                     logger.warning(str(ex), self)
@@ -128,7 +129,7 @@ class Mount(object):
             mounttools = self.config.SNAPSHOT_MODES[mode][0]
             tools = mounttools(cfg = self.config, profile_id = self.profile_id,
                                tmp_mount = self.tmp_mount, mode = mode,
-                               parent = self.parent, **kwargs)
+                               parent = self.parent, read_only = self.read_only, **kwargs)
             return tools.pre_mount_check(first_run)
 
     def remount(self, new_profile_id, mode = None, hash_id = None, **kwargs):
@@ -156,7 +157,7 @@ class Mount(object):
         mounttools = self.config.SNAPSHOT_MODES[mode][0]
         tools = mounttools(cfg = self.config, profile_id = new_profile_id,
                            tmp_mount = self.tmp_mount, mode = mode,
-                           parent = self.parent, **kwargs)
+                           parent = self.parent, read_only = self.read_only, **kwargs)
         if tools.compare_remount(hash_id):
             #profiles uses the same settings. just swap the symlinks
             tools.remove_symlink(profile_id = self.profile_id)
--- a/common/sshtools.py
+++ b/common/sshtools.py
@@ -1,4 +1,4 @@
-#    Copyright (C) 2012-2016 Germar Reitze
+#    Copyright (C) 2012-2016 Germar Reitze, Taylor Raack
 #
 #    This program is free software; you can redistribute it and/or modify
 #    it under the terms of the GNU General Public License as published by
@@ -42,7 +42,7 @@ class SSH(mount.MountControl):
 
     CHECK_FUSE_GROUP = False
 
-    def __init__(self, cfg = None, profile_id = None, hash_id = None, tmp_mount = False, parent = None, symlink = True, **kwargs):
+    def __init__(self, cfg = None, profile_id = None, hash_id = None, tmp_mount = False, parent = None, symlink = True, read_only = True, **kwargs):
         self.config = cfg
         if self.config is None:
             self.config = config.Config()
@@ -55,6 +55,7 @@ class SSH(mount.MountControl):
         self.hash_id = hash_id
         self.parent = parent
         self.symlink = symlink
+        self.read_only = read_only
 
         #init MountControl
         super(SSH, self).__init__()
@@ -75,6 +76,7 @@ class SSH(mount.MountControl):
         self.setattr_kwargs('ionice', self.config.is_run_ionice_on_remote_enabled(self.profile_id), store = False, **kwargs)
         self.setattr_kwargs('nocache', self.config.is_run_nocache_on_remote_enabled(self.profile_id), store = False, **kwargs)
         self.setattr_kwargs('password', None, store = False, **kwargs)
+        self.setattr_kwargs('read_only', self.read_only, **kwargs)
 
         if not self.path:
             self.path = './'
@@ -115,6 +117,11 @@ class SSH(mount.MountControl):
         if not self.cipher == 'default':
             sshfs.extend(['-o', 'Ciphers=%s' % self.cipher])
         sshfs.extend(['-o', 'idmap=user'])
+
+        # use read only mount if requested
+        if self.read_only:
+            sshfs.extend(['-o', 'ro'])
+
         sshfs.extend([self.user_host_path, self.mountpoint])
         #bugfix: sshfs doesn't mount if locale in LC_ALL is not available on remote host
         #LANG or other envirnoment variable are no problem.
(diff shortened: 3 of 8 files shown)


dinoboy197:
OK, turns out this isn't quite done yet :) Hit the Create PR button too fast!

common/sshtools.py
```diff
@@ -42,7 +42,7 @@ class SSH(mount.MountControl):
 
     CHECK_FUSE_GROUP = True
 
-    def __init__(self, cfg = None, profile_id = None, hash_id = None, tmp_mount = False, parent = None, symlink = True, **kwargs):
+    def __init__(self, cfg = None, profile_id = None, hash_id = None, tmp_mount = False, parent = None, symlink = True, read_only = True, **kwargs):
```
Germar: `read_only` need to be written into `unmount` info. This is would be done by `self.setattr_kwargs('read_only', False, **kwargs)` (in line 67). This is some 'magic' (actually no magic at all just crappy, undocumented code from my lazy self :wink:) that will set `self.read_only` and also put it into `unmount` info. 

`unmount` info is responsible for sharing the mount-point. 'Magic' behind this is that BIT uses a `crc32` hash of that file as mount-point name. If there is a second process trying to mount with the exact same args it will find this mount-point already mounted and just add its locks and symlink.

So don't put `read_only` into `__init__` args. `**kwargs` will fetch this. And also no `self.read_only = read_only`. Just add the `self.setattr_kwargs(...)` line in 67.

dinoboy197: Ha, yeah, I figured out the kwargs piece after a bit of debugging :) Thanks for confirming my understanding (after a bit of catting /etc/mtab, watching logs, and playing with code).

I've got a bit more to code here, which is why I closed the PR for now - certainly not done yet, as you see :)

Germar: Sorry for the mess without documentation. I promise to get better. Already working on it :v:

dinoboy197: No problem at all. It's good practice for me to get acquainted with the code anyway! Don't feel bad.

dinoboy197:
OK, reopening as I think I have the code done and working now! Though you might to take a look; though I'm still cleaning up this code a little bit.

Note that I am writing some unit tests for saving the configs for local, local_enc, ssh, and ssh_enc modes. I'll be adding those to this PR shortly.

common/test/test_sshtools.py
```diff
@@ -0,0 +1,74 @@
...
+        with tempdir() as dirpath:
+            self.config.set_snapshots_path(dirpath)
+
+            mnt = mount.Mount(cfg = self.config, tmp_mount = True, read_only = read_only)
+            mnt.pre_mount_check(mode = 'ssh', first_run = True, **self.mount_kwargs)
+            
+            hash_id = mnt.mount(mode = 'ssh', check = False, **self.mount_kwargs)
+            
+            # test mount r/o vs r/w
+            if test_rw == True:
```
dinoboy197: Not done yet - still have to implement the r/w and r/o checks

common/test/test_sshtools.py
```diff
@@ -0,0 +1,74 @@
...
+        shutil.rmtree(name)
+
+class TestSSH(unittest.TestCase):
+
+    def setUp(self):
+        logger.DEBUG = '-v' in sys.argv
+        self.config = config.Config()
+        self.config.set_snapshots_mode('ssh')
+        self.config.set_ssh_host('localhost')
+        self.config.set_ssh_user('traack')
```
dinoboy197: Will have to parameterize the ssh user and password. Will look for a in-process Python ssh server library to use for unit tests or will have to set up a test account at unit test time to verify that ssh'ing works ok.

common/encfstools.py
```diff
@@ -279,12 +283,13 @@ def __init__(self, cfg = None, profile_id = None, mode = None, parent = None,*ar
...
         self.parent = parent
+        self.read_only = read_only
         self.args = args
         self.kwargs = kwargs
 
-        self.ssh = sshtools.SSH(*self.args, symlink = False, **self.split_kwargs('ssh'))
-        self.rev_root = EncFS_mount(*self.args, symlink = False, **self.split_kwargs('encfs_reverse'))
-        super(EncFS_SSH, self).__init__(*self.args, **self.split_kwargs('encfs'))
+        self.ssh = sshtools.SSH(*self.args, symlink = False, read_only = self.read_only, **self.split_kwargs('ssh'))
+        self.rev_root = EncFS_mount(*self.args, symlink = False, read_only = self.read_only, **self.split_kwargs('encfs_reverse'))
```
Germar: We don't need to mount ssh and encfs reverse with ro in here. These can be shared with other processes using rw. The top-layer encfs mount (line 292) mounting read-only is enough

common/encfstools.py
```diff
@@ -306,7 +311,7 @@ def mount(self, *args, **kwargs):
                 tmp_kwargs = self.split_kwargs('encfs_reverse')
                 tmp_kwargs['path'] = src
                 tmp_kwargs['config_path'] = src
-                tmp_mount = EncFS_mount(*self.args, symlink = False, **tmp_kwargs)
+                tmp_mount = EncFS_mount(*self.args, symlink = False, read_only = self.read_only, **tmp_kwargs)
```
Germar: Not necessary here neither. This is just to create a new encfs-config which will be unmounted immediately.

qt4/settingsdialog.py
```diff
@@ -1320,9 +1320,18 @@ def save_profile( self ):
...
 
+        # TODO - consider a single API method to bridge the UI layer (settings dialog) and backend layer (config)
+        # when setting snapshots path rather than having to call the mount module from the UI layer
+        # 
+        # currently, setting snapshots path requires the path to be mounted. it seems that it might be nice,
+        # since the config object is more than a data structure, but has side-effect logic as well, to have the
+        # config.set_snapshots_path() method take care of everything it needs to perform its job
+        # (mounting and unmounting the fuse filesystem if necessary).
+        # https://en.wikipedia.org/wiki/Single_responsibility_principle
+
```
Germar: :+1:

common/encfstools.py
dinoboy197: Got it, thanks. Removing.

common/encfstools.py
dinoboy197: Sounds good! I wasn't sure how paranoid you wanted to be about ro on these bits :) Adjusting.

common/test/test_sshtools.py
dinoboy197: r/w and r/o checks implemented for ssh mode.

common/test/test_sshtools.py
dinoboy197: Decided to just use ssh keys for this. Travis builds seem to be a bit "stuck" - going to close the computer now, but maybe they'll be working for me in a few hours.

common/mount.py
```diff
@@ -57,9 +58,8 @@ def __init__(self, cfg = None, profile_id = None, tmp_mount = False, parent = No
                 cmd = [bit, 'pw-cache', action]
                 logger.debug('Call command: %s'
                              %' '.join(cmd), self)
-                proc = subprocess.Popen(cmd,
-                                        stdout = open(os.devnull, 'w'),
-                                        stderr = open(os.devnull, 'w'))
+                with open(os.devnull, 'w') as out:
```
dinoboy197: Side fix: Make sure that the stream to /dev/null gets closed properly

common/tools.py
```diff
@@ -1114,11 +1114,12 @@ def getSshKeyFingerprint(path):
     if not os.path.exists(path):
         return
     cmd = ['ssh-keygen', '-l', '-f', path]
-    proc = subprocess.Popen(cmd, stdout = subprocess.PIPE, stderr = open(os.devnull, 'w'))
-    output = proc.communicate()[0]
-    m = re.match(b'\d+\s+([a-zA-Z0-9:]+).*', output)
-    if m:
-        return m.group(1).decode('UTF-8')
+    with open(os.devnull, 'w') as out:
```
dinoboy197: Side fix: Make sure that the stream to /dev/null gets closed properly

common/mount.py
Germar: Rather use `subprocess.DEVNULL` which will do this automatically. BIT requires Python >= v3.3 anyways.

common/test/test_sshtools.py
```diff
@@ -0,0 +1,70 @@
...
+    def setUp(self):
+        logger.DEBUG = '-v' in sys.argv
+        self.config = config.Config()
+        self.config.set_snapshots_mode('ssh')
+        self.config.set_ssh_host('localhost')
+        self.config.set_ssh_private_key_file(os.path.expanduser(os.path.join("~",".ssh","id_rsa")))
+        self.mount_kwargs = {}
+
+    def test_can_mount_ssh_rw(self):
+        doOrWarn(lambda: self.internal_test(read_only = False, implicit_read_only = False))
```
Germar: You could do the same with
```
ON_TRAVIS = os.environ.get('TRAVIS', 'None').lower() == 'true'
@unittest.skipIf(not ON_TRAVIS, 'skip because foo')
def test_can_mount_ssh_rw(self):
    self.internal_test(read_only = False, implicit_read_only = False)
```

This would also create correct test result with `xx successful, 1 skipped, 0 error`

common/mount.py
dinoboy197: Will do

common/test/test_sshtools.py
dinoboy197: Thanks for the tip :) Looks much cleaner.

common/test/test_snapshots.py
```diff
@@ -342,6 +344,9 @@ def test_change_owner_without_root(self):
         self.assertEqual(s.st_gid, CURRENTGID)
 
     @unittest.skipIf(NO_GROUPS, "Current user is in no other group. So this test won't work.")
+    @unittest.skipIf(ON_TRAVIS, "File change group commands don't work in travis")
+    # TODO - Find way to run travis-specified Python version when using settings to allow changing groups
+    # http://stackoverflow.com/questions/35117270/how-can-i-change-the-group-of-a-file-when-executing-in-a-travis-ci-build
```
dinoboy197: Turns out that getting travis to give the process running unit tests change group ability (required for this test) while maintaining the virtualenv for the specific python versions we want during the test is a bit tricky. I've got a TODO in there for fixing that along with a stackoverflow question. Maybe you have some bright idea? Until then, we can still run that specific test locally outside of travis.

dinoboy197:
@Germar - final commits are in. Let me know if this looks ok, and if so, I'll squash the commits down to one and merge into master. Thanks!

Germar:
Found a solution for the group issue on [Travis docs](https://docs.travis-ci.com/user/trusty-ci-environment#Group-membership). I'm a bit short on time. Will have a look on the rest in couple hours.

dinoboy197:
> Found a solution for the group issue on Travis docs. I'm a bit short on time. Will have a look on the rest in couple hours.

Yeah, I found that too - unfortunately, using that has other negative side effects (mainly that it destroys the python virtualenv that travis uses to select the correct python during various stages of the build). I posted [a stackoverflow ticket on my findings so far](http://stackoverflow.com/questions/35117270/how-can-i-change-the-group-of-a-file-when-executing-in-a-travis-ci-build). Maybe you have some other idea!

Germar:
Virtualenv looks okay for me...

[Python 3.4](https://travis-ci.org/bit-team/backintime/jobs/106815251#L407)
```
/home/travis/virtualenv/python3.4.2/bin/coverage run -p -m unittest -v -b test/test_snapshots.py
```

[Python 3.5](https://travis-ci.org/bit-team/backintime/jobs/106815253#L407)
```
/home/travis/virtualenv/python3.5.0/bin/coverage run -p -m unittest -v -b test/test_snapshots.py
```

[Python 3.5-dev](https://travis-ci.org/bit-team/backintime/jobs/106815255#L401)
```
/home/travis/virtualenv/python3.5-dev/bin/coverage run -p -m unittest -v -b test/test_snapshots.py
```

[Python 3.6 Nightly](https://travis-ci.org/bit-team/backintime/jobs/106815256#L401)
```
/home/travis/virtualenv/python3.6-dev/bin/coverage run -p -m unittest -v -b test/test_snapshots.py
```

dinoboy197:
> Virtualenv looks okay for me...

Totally agreed - we can pick out the virtualenv directly.

But what is bothering me is that if we used the `sudo -E su $USER -c` method to ensure that the unit test can actually change the group of the file, then the travis configuration needs some kind of switch statement in the figure out which particular python version travis is testing against at that particular moment (3.4, 3.5, or otherwise), and explicitly use that particular virtualenv for that version.

It seems like an unfortunate hack to have to make to be able to test multiple python versions for testing forward compatibility, but I haven't found any workaround in travis for this. Maybe we'll just have to settle for that - or just drop down to unit testing one fixed python version (which wouldn't be the end of the world, but it does seem like a thing that travis provides).

What do you think about those options?

Germar:
Not sure if I understood correctly. But these abs-path calls for coverage only came from changing `CMD="coverage run -p"` to `CMD="${COVERAGE} run -p"` in `configure`. So there are no switches for different versions.

I just [run one test on Travis](https://travis-ci.org/bit-team/backintime/builds/106848481) which printed out the python version and it is the correct python version for each test.

But I'm also fine with just skipping that test on Travis. It's not so important and the method itself get tested in other tests, too.

On the other hand I already have plans on adding a testcase which will run completely with `sudo` on Travis to test methods that need root permissions. We could move this into that testcase when done.

dinoboy197:
> I just run one test on Travis which printed out the python version and it is the correct python version for each test.

Sorry for all of the confusion - I know this probably seems all very strange. You're probably curious about why this all coming up in this PR, anyway :) It's been driving me crazy for the past several days :)

The reason is that in order to have functional unit tests for the new ssh fuse read-only mount code, I need to add the travis user to the fuse group, and in doing so, that activated the snapshot test which tests changing group ownership (test_change_group) (a test previously skipped in travis since the travis user is part of only one group by default).

So now, running that snapshot unit test for changing file group fails because running them with just plain ```python``` won't give python the ability to change the group - the operating system in travis blocks it. It needs the python unittest command to be run like ```sudo -E su $USER -c "make unittest-v"``` to do that (as you saw in the travis documentation) - and the moment you use that wrapper around ```make unittest-v```, the python which gets used to run the unit tests is the system python 3.2 instead of the virtualenv python 3.4 / 3.5 with coveralls (not to mention that then other runtime code breaks which needs features in python 3.3 and above...

Suffice it to say, it's taken me very far down the rabbit hole ... :)

Germar:
Thanks for explanation.

> So now, running that snapshot unit test for changing file group fails because running them with just plain python won't give python the ability to change the group - the operating system in travis blocks it.

Actually it's not the operating system blocking this but after adding a user to a new group the user need to log out and log in again to apply that new group. So this made me curious on why your new `TestSSH` test could work without log out and log in!? And I just found out that at least on Ubuntu 14.04 `fuse` membership is not necessary anymore to mount a FuseFS. Even more in Ubuntu 15.10 `fuse` group is gone completely.

I will set up a new 12.04 VM later this evening to test how `fuse` membership was handled in there (as Travis is running 12.04). So if `fuse` membership isn't necessary anymore for mounting a FuseFS we could skip that test in BIT.

Germar:
Oh, I just noticed you're using `os.access` in that test. I [recently run into troubles](https://github.com/bit-team/backintime/issues/490#issuecomment-156265196) with `os.access` on mounted share. So it might be better to actually just try to write into that folder

```
with self.assertRaise(PermissionError):
    with open(path, 'wt') as f:
        f.write('foo')
```

dinoboy197:
> So this made me curious on why your new TestSSH test could work without log out and log in!? And I just found out that at least on Ubuntu 14.04 fuse membership is not necessary anymore to mount a FuseFS. Even more in Ubuntu 15.10 fuse group is gone completely.

The rabbit hole goes even deeper! :)

> in that test. I recently run into troubles with os.access on mounted share. So it might be better to actually just try to write into that folder

Thanks much. I found ```os.access``` on some stackoverflow post, but having your extra context on possible failure modes is great. Let me adjust that code.

dinoboy197:
> I will set up a new 12.04 VM later this evening to test how fuse membership was handled in there (as Travis is running 12.04). So if fuse membership isn't necessary anymore for mounting a FuseFS we could skip that test in BIT.

Depending on how widely you currently distribute BIT among Ubuntu releases, we might consider [changing the Travis build to start with Trusty](https://docs.travis-ci.com/user/trusty-ci-environment/), based on the results of your tests.

Germar:
Done. 12.04 doesn't need fuse group either. I double checked with an unprivileged user who was only in his own group and was able to mount FuseFSs. So I changed default `configure` options to `--no-fuse-group` as Ubuntu/Debian were the only Distributions who needed `fuse` group before. Didn't check with Debian yet but I'm pretty sure it's the same because Ubuntus encfs and sshfs are direct imports from Debian.

So if you merge master you can remove the `sudo gpasswd -a travis fuse` from `.travis.yml` again.

Indeed an interesting rabbit hole you stumbled in! :)

For [our PPA](https://launchpad.net/~bit-team/+archive/ubuntu/stable) I only build packages for Trusty, Vivid, Wily and Xenial. But as Trusty on Travis is still Beta I think we should stay on Precise as long as there is no real need to switch.

dinoboy197:
Sounds good. This is going to master now - and will move on. Learned quite a bit on this PR :)

Germar:
Good job, thanks!

PR #533, opened 2016-01-29
Starting unit tests for config module

dinoboy197:
I was working a bit with testing functionality in the ```config``` module, and I thought I'd start adding some unit tests!

And I found a bug by accident :) I'm not sure if I fixed in the way that is correct, as I'm not 100% sure about the usage of the ```parent``` kwarg in the ```logger.error()``` method.

### DIFF
--- a/common/test/test_config.py
+++ b/common/test/test_config.py
@@ -1,0 +1,52 @@
+# Back In Time
+# Copyright (C) 2016 Taylor Raack
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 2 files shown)


Germar:
Oh, right. Good catch. Probably a copy/paste error. I'll add a unittest which will trigger the error in `test/test_tools.py`, too.

`parent` is just a workaround to get the class where the log method was called 
```
DEBUG: [common/config.py:244 Config.check_config] Check profile Hauptprofil
                             ^^^^^^
                             this part
```
I didn't find a way to extract that class from `sys._getframe`

BTW in Python3 there is `tempfile.TemporaryDirectory` which can be used as a context manager. You don't need to create your own context manager with `mkdtemp` anymore

dinoboy197:
> BTW in Python3 there is tempfile.TemporaryDirectory which can be used as a context manager. You don't need to create your own context manager with mkdtemp anymore

Nice! Thanks for the tip.

PR #537, opened 2016-02-07
Added basic local snapshot end to end test

dinoboy197:
I took the time to write the first end-to-end test of BIT :)

It's certainly not a full test suite, but we can build on this going forward. It gives us a place to start. I've written many such end-to-end tests for other software that I maintain professionally and I think approaching end-to-end tests as coded here are incredibly valuable for hard-to-spot regressions.

Ideas:
* test one of highest level interfaces - in this case, the command line - to ensure that everything from argument parsing, functionality, and output works as expected.
 * Maybe in the future we can think about testing the GUI in an automated way (though that is out of my area of expertise)
* don't call python functions directly; instead, test what a command line user would use - the command line itself in a shell, by exec'ing the python interpreter
* can signal subtle bugs not covered in method unit tests, which can go a long way towards preventing regressions during major refactors. this can happen when individual unit tests might pass due to API compatibility but mismatched expectations between semantics of module inputs and outputs

Example of immediate effectiveness: This extremely basic test, which doesn't even cover advanced configuration, nor ssh or encrypted mode, already uncovered some CLI bugs. Seems like a win!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,9 +12,6 @@ addons:
 
 python:
   - "3.4"
-  - "3.5"
-  - "3.5-dev" # 3.5 development branch
-  - "nightly" # currently points to 3.6-dev
 
 before_install:
   - sudo apt-get -qq update
--- a/common/cli.py
+++ b/common/cli.py
@@ -35,7 +35,7 @@ def restore(cfg, snapshot_id = None, what = None, where = None, **kwargs):
 
     snapshots_list = snapshots.listSnapshots(cfg)
 
-    sid = selectSnapshot(snapshots_list, snapshot_id, 'SnapshotID to restore')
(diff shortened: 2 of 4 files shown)


Germar:
This is great! Thanks. Only think I would like to change is to call `./backintime --config test/config check-config` instead of `python3 backintime.py ...`. Because that would be the command which a user would call (except the slightly different folder structure after installing)

> Example of immediate effectiveness

And also an example of my sloppy work...  :flushed:

dinoboy197:
> Only think I would like to change is to call ./backintime --config test/config check-config instead of python3 backintime.py .... Because that would be the command which a user would call (except the slightly different folder structure after installing)

You're absolutely right - thanks for the good catch. I'm merging this in to master now.

dinoboy197:
> And also an example of my sloppy work...

Oh and believe me, I've written my fair share of bugs in my years of coding. :bomb:

PR #539, opened 2016-02-08
Local mode mounts over bindfs

dinoboy197:
* Added bindfs as an optional dependency
* Changed all local mode usage to bindfs, mounting r/w for snapshotting and r/o for all other operations (same as ssh, local_encfs, ssh_encfs)
* Fixed a couple of bugs which integration tests found
* Added a ```get_local_path``` and ```set_local_path``` method pair in the ```Config``` class for shuttling the snapshot path that the user selects (without it being conflated with the ```get_snapshot_path```, which is overloaded for use with the bindfs mount point)

### DIFF
--- a/common/bindfstools.py
+++ b/common/bindfstools.py
@@ -1,0 +1,72 @@
+#    Copyright (C) 2016 Taylor Raack
+#
+#    This program is free software; you can redistribute it and/or modify
+#    it under the terms of the GNU General Public License as published by
+#    the Free Software Foundation; either version 2 of the License, or
+#    (at your option) any later version.
+#
+#    This program is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU General Public License for more details.
+#
+#    You should have received a copy of the GNU General Public License along
+#    with this program; if not, write to the Free Software Foundation, Inc.,
+#    51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+
+import subprocess
+import gettext
+
+from mount import MountControl
+import logger
+from exceptions import MountException
+
+_=gettext.gettext
+
+class Local(MountControl):
+    """
+    Mount local path with bindfs. This allows for read-only viewing of snapshots stored on a locally mounted filesystem.
+    """
+
+    def __init__(self, *args, **kwargs):
+        #init MountControl
+        super(Local, self).__init__(*args, **kwargs)
+
+        #First we need to map the settings.
+        self.setattr_kwargs('path', self.config.get_local_path(self.profile_id), **kwargs)
+
+        self.set_default_args()
+        
+        self.mountproc = 'bindfs'
+        self.symlink_subfolder = None
+
+        self.log_command = '%s: %s' % (self.mode, self.path)
+
+    def _mount(self):
+        """
+        mount the local filesystem over bind
+        """
+        bindfs = [self.mountproc, '-n']
+
+        # use read only mount if requested
+        if self.read_only:
+            bindfs.extend(['-o', 'ro'])
+
+        bindfs.extend([self.path, self.mountpoint])
+        logger.debug('Call mount command: %s'
+                     %' '.join(bindfs),
+                     self)
+        try:
+            subprocess.check_call(bindfs)
+        except subprocess.CalledProcessError:
+            raise MountException( _('Can\'t mount %s') % ' '.join(bindfs))
+
+    def pre_mount_check(self, first_run = False):
+        """
+        check what ever conditions must be given for the mount to be done successful
+        raise MountException( _('Error description') ) if service can not mount
+        return True if everything is okay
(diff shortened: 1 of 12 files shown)


common/bindfstools.py
```diff
@@ -0,0 +1,157 @@
...
+import config
+import mount
+import logger
+import tools
+from exceptions import MountException
+
+_=gettext.gettext
+
+# TODO - quite a few shared methods between the various mount.MountControl subclasses. Perhaps move some common
+# methods into the base class?
```
Germar: Sure. I did this already yesterday for the [pre|post]_u?mount_checks in commit 9e484938224a68a3e30fbef18270af960d454497. But this would also work for `_umount()` and `_check_fuse()`

common/test/test_bindfstools.py
```diff
@@ -0,0 +1,79 @@
...
+class TestLocal(generic.TestCase):
+
+    def setUp(self):
+        super(TestLocal, self).setUp()
+        logger.DEBUG = '-v' in sys.argv
+        self.config = config.Config()
+        self.config.set_snapshots_mode('local')
+        self.mount_kwargs = {}
+
+    @unittest.skipIf(not ON_TRAVIS, 'Skip as this test requires a local ssh server, public and private keys installed')
```
Germar: copy/paste? Should rather be `skipIf(not BINDFS_EXIST)`

common/test/test_snapshots.py
```diff
@@ -226,7 +226,7 @@ def test_rsyncSuffix(self):
                                                        '*blub',
                                                        '/bar/2'])
         self.assertRegex(suffix, r'^ --chmod=Du\+wx  ' +
-                                 r'--exclude="/tmp/.*?" ' +
+                                 r'--exclude=".*?\.local/share/backintime/mnt.*?" ' +
```
Germar: The mountpoint is already excluded with next line. But the destination `/tmp/...` must stay excluded here. Otherwise we would backup snapshots again

common/test/test_bindfstools.py
dinoboy197: Ah, right - good catch, thanks.

common/test/test_snapshots.py
dinoboy197: OK, sounds good. I'll make the change.

Germar:
I moved `check_fuse` and `_umount` into `mount.MountControl` base class in commit 7f54d16dbcca4eeadfe4985897b2468d053d0f95. There is a new arg which need to be set for this in `__init__` from subclass: `self.mountproc = 'bindfs'`

Please take a look on `encfstools.py` on how the new `__init__` now looks like.

common/bindfstools.py
```diff
@@ -0,0 +1,157 @@
...
+        # use read only mount if requested
+        if self.read_only:
+            bindfs.extend(['-o', 'ro'])
+
+        bindfs.extend([self.path, self.mountpoint])
+        #bugfix: bindfs doesn't mount if locale in LC_ALL is not available on remote host
+        #LANG or other envirnoment variable are no problem.
+        env = os.environ.copy()
+        if 'LC_ALL' in list(env.keys()):
+            env['LC_ALL'] = 'C'
```
Germar: This was just a workaround for `sshfs`. I don't think `bindfs` has problems with locales, does it?

dinoboy197: Ah, you're right, it does not appear to have that problem. I had misread the comment and thought it was at the fuse layer. I'll remove this here.

dinoboy197:
OK @Germar , I addressed your comments. Ready for you to take a second look.

common/test/test_backintime.py
```diff
@@ -93,7 +111,20 @@ def test_local_snapshot_is_successful(self):
...
-''', re.MULTILINE))
+INFO: Lock(
+INFO: Inhibit Suspend started. Reason: take snapshot)?
+INFO: mount local: /tmp/snapshots on .*
+INFO: Take a new snapshot. Profile: 1 Main profile
+INFO: Call rsync to take the snapshot
+INFO: Save config file
+INFO: Save permissions
+INFO: Create info file
+INFO: Remove backups older than: 20060201-000000
```
Germar: Date must be a regex, too

common/bindfstools.py
```diff
@@ -0,0 +1,112 @@
...
+
+_=gettext.gettext
+
+# TODO - Move some common methods (_umount() and check_fuse()) into the base class
+class Local(MountControl):
+    """
+    Mount local path with bindfs. This allows for read-only viewing of snapshots stored on a locally mounted filesystem.
+    """
+
+    CHECK_FUSE_GROUP = False
```
Germar: I moved this into `MountControl` base-class, too. `common/configure` will change it only in `mount.py` based on the givven arguments `--fuse-group` or `--no-fuse-group`. This would break it.

Germar:
Looks good. Thanks!

common/test/test_backintime.py
dinoboy197: Another good catch!

common/bindfstools.py
dinoboy197: Excellent. Will remove this here.

PR #541, opened 2016-02-11
imperative → progressive

to-ba:
Should be better in progressive form

### DIFF
--- a/common/po/messages.pot
+++ b/common/po/messages.pot
@@ -537,7 +537,7 @@ msgstr ""
 
 #: ../../common/snapshots.py:1340
 #, python-format
-msgid "Compare with snapshot %s"
+msgid "Comparing with snapshot %s"
 msgstr ""
 
 #: ../../common/snapshots.py:1360
@@ -549,7 +549,7 @@ msgid "Save config file ..."
 msgstr ""
 
 #: ../../common/snapshots.py:1445
-msgid "Save permission ..."
+msgid "Saving permissions ..."
 msgstr ""
 
 #: ../../common/snapshots.py:1470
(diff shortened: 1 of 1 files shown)


dinoboy197:
Thanks for the suggestions!

There seem to be some other English language verbs to change ("Save config file...", "Restore to...") - should those change as well?

Germar:
Thanks for the pull request! `messages.pot` is just an auto generated file from source (done with `common/po/genmessages.sh`). Those changes need to be done directly in `common/snapshots.py`.

Changing the original phrase will reset all translations. To avoid this I used to also change the `.po` files and import them back into Launchpads Translations. But this part is a bit tricky. So I'll do this my self.

@dinoboy197 `Restore to...` is a command in GUI. IMHO this should stay imperative. But I'm no native English speaker. So I might be wrong...

dinoboy197:
> @dinoboy197 Restore to... is a command in GUI. IMHO this should stay imperative. But I'm no native English speaker. So I might be wrong...

Perfect - your description about this being a auto generated file helps me understand where these come from :)

Germar:
@dinoboy197 after adding new phrases in source you need to run `common/po/getmessages.sh` and commit `messages.pot`. Launchpad Translations will automatically import that template couple hours later.

Before releasing a new version I manually [download](https://translations.launchpad.net/backintime/trunk/+export) current translation in `.po` format, put that file into `common/po/` and run `common/po/updatelaunchpad.sh`. This will unpack and rename all language files. Finally commit the new `*.po` files.

to-ba:
@Germar I hope you will correct the right file :-)
@dinoboy197 I don't know where it is used ("Save config file...", "Restore to...") but if it is a command don't change it.

@all »Take snapshot« should be splited in two strings (or more). One for the activity, used in the taskbar »Taking snapshot« or »Creating snapshot«; second for the command in the menu, where it is correct. I don't know where it is also used, looks like four times.
#: ../../common/encfstools.py:512 ../../common/snapshots.py:1131
#: ../../common/snapshots.py:1388 ../../qt4/app.py:78
msgid "Take snapshot"

Germar:
Done in commit 64d56ee65cb4e7e3a61d5dd3c1f5ceb1be83a2ad

to-ba:
Danke, Thanks :-)

PR #550, opened 2016-02-27
Fix #549 + german commen for .desktop file

### LINKED ISSUE #549: BIT-root wont start if started over the Gnome Shell louncher(.desktop file)
As far as I understand it the issue is that the pkexec executed from the .desktop file is not able to draw the "ask for password dialog".

* I looked how Gparted dos the trick - and figured that they use a wrapper script.
* I tweaked their script to fit BIT. -Pullrequest coming soon

Wabuo:
* Fix #549 by adding an wrapper script to execute pkexec corect.

   * Changed the BIT-root.desktop file to use the wrapper script.

* Added a german comment to both .desktop files.


qt4/backintime-qt4_polkit
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+if [ $(which pkexec) ]; then
+        pkexec --disable-internal-agent "/usr/bin/backintime-qt4" "$@"
```
dinoboy197: As I understand pkexec from man mages and other issues on the web, ```--disable-internal-agent``` helps when you're trying to execute a script with PolicyKit but you don't have a GUI (and thus no authentication agent).

However, it seems to me that one not execute the GUI for BIT (```backintime-qt4```) without a GUI already running.

Are you having difficulty getting prompted from the GUI because you don't have an authentication agent registered, or are having issues having BIT launch in root mode from a terminal (no GUI) because it won't prompt for a password (or some other situation?)

Wabuo: * No I have a working authentication agent (the built in one from Gnome3).
* Other Applications work just fine.
* Manually starting it in the terminal which `pkexec backintime-qt4` or with ALT F2 run works fine.
* Only starting it over the .desktop file won't work!
* With the mentioned changes it now works like a charm!

About the `--disable-internal-agent` I didn't looked in to it I just changed the calls of GParted in to BIT.

dinoboy197: Got it, thanks.

It's worth noting that while ```gparted``` is called with ```--disable-internal-agent``` on your system, on my system (and presumably other Ubuntu systems), ```gparted``` is NOT called with ```--disable-internal-agent```. I am certainly not a ```pkexec``` expert, but I'd be interested in a quick survey of how ```pkexec``` is used to activate authentication agents among different distributions before changing the way that we use it for everyone.

What distro are you using?

Wabuo:
Makes sense ... =)

Wabuo:
Archlinux ore to be correct Antergos.

Wabuo:
From the [Ubuntu manuals](http://manpages.ubuntu.com/manpages/precise/en/man1/pkexec.1.html)

> AUTHENTICATION AGENT

       pkexec, like any other PolicyKit application, will use the
       authentication agent registered for the calling process. However, if no
       authentication agent is available, then pkexec will register its own
       textual authentication agent. This behavior can be turned off by
       passing the --disable-internal-agent option.

In my opinion it makes sense to disable the textual authentication agent because if the wrapper script is called from a .desktop file or another script, and no authentication agent is available, the textual authentication agent would be run in the background and wait for user input, no one will ever provide.
But with `--disable-internal-agent` in the same case it will exit with a return value of 127.

> If the calling process is not authorized or an authorization
> not be obtained through authentication or an error occured,
> exits with a return value of 127

Germar:
For me this sounds like there are some environ vars missing during launching `pkexec` from `.desktop` file. I'm wondering if we really need a wrapper script to fix it or if we could rather just change `backintime-qt4-root.desktop`:
```
Exec=pkexec --disable-internal-agent backintime-qt4
```

Wabuo:
I tried to directly use `Exec=pkexec --disable-internal-agent backintime-qt4` in the .desktop file but it doesn't work!
But after reading the [Desktop Entry Specification](https://developer.gnome.org/desktop-entry-spec/) I still don't understand why ...

Germar:
Okay, I checked this. `pkexec` will fail with `Refusing to render service to dead parents.` like described in [this bug report](https://bugs.launchpad.net/ubuntu/+source/update-manager/+bug/1020115). Seems like a wrapper script is really the only (ugly) workaround.

So I'm fine with merging this but first please remove the check for `pkexec` (and the alternate call of plain `backintime-qt4`). `pkexec` is a hard dependency of BIT and instead of running BIT as non-priveleged user if it is (for what ever reason) not available I'd rather let it fail with a bang. @Wabuo would you please change this?

Wabuo:
@Germar done.

Germar:
Good job! Thanks

PR #607, opened 2016-07-28
Retrieve path device and filesystem via df command

karlicoss:
So, I finally decided to automate my backups and use backintime. I tested scheduled udev-triggered backups on a flash drive few days ago, everything worked fine. So I decided to configure backups on my external HDD today, but it didn't work, plugging in the drive didn't trigger backintime.

After a short investigation, I found out my udev file was set up to trigger on some unknown UUID, even `blkid` wouldn't list it. So, I dug in source code and figured out that the `tools.device` function returned `None` for my drive.

The reason was: its label ('karlicoss hdd') contained a space, so it was mounted to `/media/karlicos/karlicoss hdd/`, however, `/etc/mtab` treats spaces as separators, so he escapes paths like

    `/dev/sdb1 /media/karlicos/karlicoss\040hdd fuseblk rw,nosuid,nodev,....`

As you can see, spaces map to `\040`. So, probably a good idea is, instead of manually parsing `/etc/mtab`, use the [`df` command from coreutils](https://www.gnu.org/software/coreutils/manual/html_node/df-invocation.html), which does all the job for us.

A followup question, what is the rationale behind [this](https://github.com/bit-team/backintime/blob/master/common/config.py#L1537) UUID fallback logic? In my case, it seemed to have cached the flash drives' UUID, and silently used it instead my HDD's UUID. Maybe, we should at list show some warning/notification to user so he knows something went not as expected?

P.S. I messed with [previous](https://github.com/bit-team/backintime/pull/606) PR and apparently github doesn't let to reopen branch which you force pushed so I made a new PR.

### DIFF
--- a/common/test/test_tools.py
+++ b/common/test/test_tools.py
@@ -408,17 +408,15 @@ class TestTools(generic.TestCase):
         proc = os.path.join('/proc', str(os.getpid()), 'fd')
         self.assertEqual(tools.mountpoint(proc), '/proc')
 
-    def test_mountArgs(self):
-        rootArgs = tools.mountArgs('/')
-        self.assertIsInstance(rootArgs, list)
-        self.assertGreaterEqual(len(rootArgs), 3)
-        self.assertEqual(rootArgs[1], '/')
+    def test_get_df_output(self):
+        df_root = tools.get_df_output('/')
+        self.assertIsInstance(df_root, list)
+        self.assertGreaterEqual(len(df_root), 2)
 
-        procArgs = tools.mountArgs('/proc')
-        self.assertGreaterEqual(len(procArgs), 3)
-        self.assertEqual(procArgs[0], 'proc')
-        self.assertEqual(procArgs[1], '/proc')
-        self.assertEqual(procArgs[2], 'proc')
+        df_proc = tools.get_df_output('/proc')
+        self.assertGreaterEqual(len(df_root), 2)
+        self.assertEqual(df_proc[0], 'proc')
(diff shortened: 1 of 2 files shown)


Germar:
Thanks karlicoss for the Pull Request. I used `df` for this before but changed it to reduce dependencies. I prefer to stay with parsing `/etc/mtab` and just decode the string correctly.

The UUID fallback is used if your drive is not connected during you make changes inside a different profile. To prevent using a wrong UUID like it happen in your case I will now remove the cached UUID if a user changes the snapshot path (which will most likely be if the user changes the drive).

karlicoss:
@Germar hmm, df is in coreutils, so not sure how one could not have it in his system, but okay, thanks for the fix.

Germar:
Under normal circumstances it should be on the system, sure. But if you think about chroot or containers it must be a dependency. The ArchLinux folks for example run unittests for packages in a clean chroot environ and already failed without coreutils dependency before. Also Ubuntus new Snap containers for Applications would need the dependency.

karlicoss:
Oh I see, I probably didn't have the big picture. Thanks for making this clear :)

jackyohh:
Is this bug officially fixed? ... because i still get this error with Version 1.1.20. drive names with space(s) still make problems when i want to autobackup on external drives (with udev).

"Error: Couldn't find UUID for "/media/DRIVE WITH SPACE/BACKUP-FOLDER/backintime/PC-NAME/root/1"

Germar:
It's fixed in current dev version 1.2.0 but not in 1.1.20

PR #618, opened 2016-08-14
improved debug output to identify profiles

buhtz:
This is my first PR here. So it is kind of a practice for me. ;)
Sorry for promoting two modifications in one PR. This is uncommon an I will do it better next time. Hope this is ok today.

1.
In `configfile.py`: `profilesSortedByName()`
The dictionary there now is ordered a more pythonic way. It depends on the offical documentation.

2.
In `backintime.py`:
This is the main reason for this PR.
In the debug output not the ProfileID is "connected" with the ProfileName.
I am not sure if this is the best solution. Maybe I should add a new methode to `class ConfigFileWithProfiles` like `def profilesIdNames()` returning a dictionary?

### DIFF
--- a/common/backintime.py
+++ b/common/backintime.py
@@ -649,7 +649,11 @@ def getConfig(args, check = True):
     cfg = config.Config(config_path = args.config, data_path = args.share_path)
     logger.debug('config file: %s' % cfg._LOCAL_CONFIG_PATH)
     logger.debug('share path: %s' % cfg._LOCAL_DATA_FOLDER)
-    logger.debug('profiles: %s' % cfg.profiles())
+    dbgProfileIdNames = ''
+    for profileId in cfg.profiles():
+        dbgProfileIdNames += '{}={} '.format(profileId,
+                                          cfg.profileName(profileId))
+    logger.debug('profiles: %s' % dbgProfileIdNames.strip())
     if 'profile_id' in args and args.profile_id:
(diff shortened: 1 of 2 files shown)


Germar:
Thanks Christian,

1. like Travis already complained, the output of `profilesSortedByName()` must be a list of ID's instead of profile-names. `return list(profiles_sorted.values())` should fix this.

2. a more pythonic way: :wink: 
```
logger.debug('profiles: %s' % ', '.join(['%s=%s' %(x, cfg.profileName(x)) for x in cfg.profiles()]))
```

buhtz:
A general question: What would be the right ithub-way now? How can I modify an existing PullRequest? Should I close it, do my modifications, commit again,and create a new PR?

1.
Ah yes. Sorry. Here on the PR-page I see a travis-error but no details. Clicking on "Details" gives me a white page (Firefox with deactivated addblocker).

2.
Great! I was thinking about something like that but couldn't find out.

Germar:
You can commit to your branch and push it to github. This will automatically be picked up by the PR.

1. Here you are
travis

You can (and should) also run those unittest local:
```
cd common
./configure
make test
```

buhtz:
Btw I would suggest `make test-v` instead of `make test` on vey old machines (like mine). It looks like that the machine freezes - but of course it doesn't.

Germar:
Thanks Christian. And congrats for your first PR. Good job!

PR #620, opened 2016-08-18
Update .gitignore

buhtz:
This file appears on my Sidcution GNU/LInux (Debian unstable) environment. Don't know why beause I am not so familar with deb-package building.


Germar:
You're right. Same here. Didn't notice before. Thanks

PR #628, opened 2016-08-31
using long versions of rsync options

buhtz:
Use long versions of the rsync options. E.g. `--perms` instead of `-p`.

This make it easier to understand the code, the debug messages and doing diagnosis.

btw: unitest on my own machine is not possible currently because of problems in the test-suite. Germar knows about it and work on it.


Germar:
Thanks Christian

PR #656, opened 2016-10-30
fix spelling and typos throughout

kortschak:
I noticed some spelling issues while I was spelunking yesterday and thought I thank you for fixing my issue by going over the code for English spelling.

I have left some spelling in the form it is (cases where English would not normally use compounded word as German does, e.g. "commandline", though see PR code comment).

@Germar Please take a look.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -675,7 +675,7 @@ Version 0.8
 * add English help (docbook)
 * add help button in main toolbar
 * the application can be started with a 'path' to a folder or file as command line parameter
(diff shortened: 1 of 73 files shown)


kortschak:
It seems the file in question is not in the PR. I did consider changing "user-callback returncode: " [here](https://github.com/bit-team/backintime/blob/b3b4b36b7613347a84ed4f5b0cc2792cd14582f1/common/plugins/usercallbackplugin.py#L62) to "user-callback return code: ", but left it as it may break grep on logs for some users.

Germar:
Thanks a lot! I definitely need to enable spellcheck in Atom :see_no_evil:

PR #727, opened 2017-03-29
Various hardening measures

mgerstner:
I'm currently doing a security review of backintime for openSUSE and SUSE Linux distributions.

While I found no exploitable security issues I'd like to include the following hardening measures for improved defense in depth.

### DIFF
--- a/common/backintime
+++ b/common/backintime
@@ -24,4 +24,4 @@ else
 	APP_PATH=$(readlink -m "${CUR_PATH}/../share/backintime/common")
 fi
 
-python3 $APP_PATH/backintime.py "$@"
+python3 -Es $APP_PATH/backintime.py "$@"
--- a/common/backintime-askpass
+++ b/common/backintime-askpass
@@ -28,4 +28,4 @@ else
 	APP_PATH=$(readlink -m "${CUR_PATH}/../share/backintime/common")
 fi
 
-python3 $APP_PATH/askpass.py "$@"
+python3 -Es $APP_PATH/askpass.py "$@"
--- a/common/config.py
+++ b/common/config.py
@@ -37,7 +37,7 @@ import sshtools
 import encfstools
 import password
 import pluginmanager
-from exceptions import PermissionDeniedByPolicy, InvalidChar
+from exceptions import PermissionDeniedByPolicy, InvalidChar, InvalidCmd
 
 _=gettext.gettext
 
@@ -637,7 +637,7 @@ class Config(configfile.ConfigFileWithProfiles):
             ssh += ['ionice', '-c2', '-n7']
         # run 'nice' on remote host
         if nice and self.niceOnRemote(profile_id) and cmd:
-            ssh += ['nice', '-n 19']
+            ssh += ['nice', '-n19']
         # run prefix on remote host
         if prefix and cmd and self.sshPrefixEnabled(profile_id):
             ssh += self.sshPrefixCmd(profile_id, cmd_type = list)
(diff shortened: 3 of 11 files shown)


Germar:
Thanks a lot for your effort! I'll check and merge it soon.

mgerstner:
We've found an issue with the polkit authorization, see commit 7f208dc. This should be adressed. Thank you.

Germar:
Thanks @mgerstner. I highly appreciate your work!

Your PR is merged and I backported 7f208dc547f569b689c888103e3b593a48cd1869 into Series1.1

mgerstner:
We're happy to help. Thanks for merging!

buhtz:
Dear Matthias,
sorry for warming this up. My question is about this part of your commit:

https://github.com/bit-team/backintime/blob/7d24d1d07bf5ecb47990348dc9f291add52a6a49/qt/serviceHelper.py#L108

I would like to find a better solution (because of #1027). So I have some basic questions.
I understand that there need to be a limit somehow. 

But I don't see why BIT have to do the check instead of DBus itself. Isn't there such a mechanism?

And how high should the limit be? Even with 100 chars you can do a lot of bad things. How to decide about the limit?

Maybe it is a better approach to somehow parse the giving "cmd" if it has valid values in it? But IMHO even this doesn't made much sense. That limit do exists only inside of BIT and not on DBus site. So why should BIT check this when BIT itself do create that cmd?

Germar:
@buhtz sorry, but you totally misunderstood the concept of DBus  in BiT. The class `UdevRules` in `serviceHelper.py` is executed with root permission so it is able to write into `/etc/udev/rules.d/`.
The unprivileged BiT process will send two strings with `addRule(cmd, uuid)` through DBus to that privileged process. To prevent someone (not only an other BiT process) from writing unauthorized code with root permission there are some rules in `serviceHelper.py`. For example we only accept the commands `nice`, `ionice` and `backintime`. And it only accept a max lenght of 100 chars. It wouldn't make sense to check this in the unprivileged BiT process.

buhtz:
Yes, I'm not clear about DBus and especially DBus in BIT. But thanks for explaining. The picture becomes a bit clearer.

Am I right to assume that this is only relevant for snapshot profiles using Udev scheduling (start when USB-Drive is plugged in)?

I didn't separate between "unpriviliged BIT process" and "serviceHelper.py". I meant checking in the serviceHelper. Would it make sense to check the "cmd" variable if there are only that allowed commands in it and the rest need to be valid paths? In that case there wouldn't be no need to check for a a max length. But I might make it to easy here.

mgerstner:
> I would like to find a better solution (because of #1027). So I have some basic questions. I understand that there need to be a limit somehow.
[...] 
> And how high should the limit be? Even with 100 chars you can do a lot of bad things. How to decide about the limit?
> 
> Maybe it is a better approach to somehow parse the giving "cmd" if it has valid values in it? But IMHO even this doesn't made much sense. That limit do exists only inside of BIT and not on DBus site. So why should BIT check this when BIT itself do create that cmd?

On D-Bus level it is a bit like the TCP protocol with some basic data types defined on top. So there is a maximum message length (128 MiB) but that doesn't really help here.

I'm not sure about this old contribution of mine. It seems you already had some other 'arbitrary limits' enforced there and I just chimed in.

Actually, since this check happens on the privileged `serverHelper` side, the data has already been transfered, so D-Bus is not even involved. It is a policy decision in the privileged helper. I agree that arbitrarily enforcing less than 100 characters doesn't really make sense.

There already *is* a validation of the command in `_validateCmd`, but it only checks for the base command (nice, ionice) and that `backintime` is the target command for the base command.

I don't recommend validating the involved paths, this is even messier to do in userspace.

Since the calls to the `serviceHelper` are already authorized via Polkit I guess dropping the maximum command line length check is also okay (but then you should maybe also think again about the other limit enforcements).

PR #729, opened 2017-03-29
Fix subprocess call (fixes #728)

### LINKED ISSUE #728: After upgrade from 1.1.14 to 1.1.16 manual snapshot button doesn't work
I upgraded BIT via apt upgrade today to 1.1.16 from version 1.1.14. I'm suffering the issue/bug #722. 
Now, the manual snapshot button doesn't work, seems to do nothing.

I tried uninstall BIT and make a fresh install, but same problem.

My system: Ubuntu Desktop 16.04.2 with 4.8.0 kernel.

Thanks.

corro:
Popen() requires the optional argument `shell=True` in order to be able to run a shell command with arguments. This fixes #728.

### DIFF
--- a/common/backintime.py
+++ b/common/backintime.py
@@ -66,7 +66,7 @@ def take_snapshot_now_async( cfg ):
             del env[i]
         except:
             pass
-    subprocess.Popen(cmd, env = env)
+    subprocess.Popen(cmd, env = env, shell = True)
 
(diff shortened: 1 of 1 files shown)


Germar:
Thanks Robin, this would work. But I prefer to not use `shell=True` and rather provide the command as a List like it is supposed to be...

PR #913, opened 2018-07-20
FIX: Workaround for #912: Backup operation broke older permissions

### LINKED ISSUE #912: backup operation modifies previous backup dir's permissions
With current master (012b0783) and the following setting (all are necessary conditions for this behaviour):

- I have "always checksum" enabled.
- "Always take snapshot" disabled (it should only take a snapshot when there are changes).

 I see the following problem:

1. Create snapshot of a changed source, as usual.
2. Try to create another snapshot -> nothing seems to happen.
3. Try to create another snapshot again -> a full snapshot is taken, this also does not create any/all hard links (-> disk space is used up).

Further examination of duplicated files (although they should just be hard-linked via `--link-dest`) shows that they changed their permission during step 2. above.  Files that were not writable by anyone (`r--------`) become user-writeable (`rw-------`) in that step.  This probably also causes rsync to not hard-link the files (and see them as changed in the first place).

quazgar:
Attempting a backup when there are no changes to be submitted added the u+w
permission bit for all files.  I am not certain if this workaround works for
read-only directories (maybe it only works for normal files).

Fixes #912

### DIFF
--- a/common/tools.py
+++ b/common/tools.py
@@ -660,7 +660,10 @@ def rsyncRemove(config, run_local = True):
     Returns:
         list:                   rsync command with all args
     """
-    cmd = ['rsync', '-a', '--delete']
+    # '--super' is needed as a workaround for rsync's bug #12806:
+    # https://bugzilla.samba.org/show_bug.cgi?id=12806
+    # The option can probably be dropped after that bug has been fixed.
+    cmd = ['rsync', '-a', '--delete', '--super']
     if run_local:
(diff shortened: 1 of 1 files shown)


Codeberg-AsGithubAlternative-buhtz:
No answer after 1,5 year. Can be closed.

Codeberg-AsGithubAlternative-buhtz:
@emtiu : I vote for close because the PR is described as a workaround.

But the related #912 should kept open and needs further investigation. Maybe we can reproduce it.

PR #937, opened 2018-10-04
Fixed: test cases were failing when run as root

mellertson:
Bug fixed: test cases were failing when run as root on Debian 9 (stretch).  The test case's comments said it was saving the file "foo" to the /tmp folder, but it was actually saving the file to the / directory.  I confirmed by running test cases on my system and saw the file "foo" was in fact saved in '/foo', and left there after the test cases failed.  

Additionally, the test case was asserting False instead of asserting True.


Germar:
This was about testing the fail routine in `ConfigFile.save` but I didn't thought about running the tests as root.

Issue #950, opened 2018-11-12
AppName fix from python3 to Back In Time


kupiqu:
It looks bad when applets displaying the application/window titles (typically used as a partner of the global menu applet) show python3 as the application name of back in time. This needs to be fixed to instead display Back In Time.

Please see screenshot:

![image](https://user-images.githubusercontent.com/3875196/48320601-c16bac80-e5e9-11e8-98e7-0350131e35a7.png)

PR #951, opened 2018-11-12
Fix issue #950: AppName fix from python3 to Back In Time

kupiqu:
Something similar may be needed for other GUI's

### DIFF
--- a/qt/app.py
+++ b/qt/app.py
@@ -1601,6 +1601,7 @@ if __name__ == '__main__':
 
     logger.openlog()
     qapp = qttools.createQApplication(cfg.APP_NAME)
+    qapp.setApplicationName(cfg.APP_NAME)
     translator = qttools.translator()
     qapp.installTranslator(translator)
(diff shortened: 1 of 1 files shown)


Germar:
In Qt 4.8 this was done with `QApplication(sys.argv + ['-title', app_name])`
But it looks like this was drooped in Qt5. I'd rather put your solution in `qttools.createQApplication` and remove ` + ['-title', app_name]` so it would work for every GUI used by BiT

kupiqu:
> In Qt 4.8 this was done with `QApplication(sys.argv + ['-title', app_name])`
> But it looks like this was drooped in Qt5. I'd rather put your solution in `qttools.createQApplication` and remove ` + ['-title', app_name]` so it would work for every GUI used by BiT

Sure, please go ahead

Germar:
was fixed in ba37b1e7d452200585a21224eb240b517b8c0fd4

PR #959, opened 2018-12-01
CLI fixes for docker

belsander:
When trying to run backintime just as a cli in Docker, some fixes have to be applied. Check commit messages for more details.


belsander:
Same thing with python 3.5.1-3 however...

belsander:
result ended up in the following Docker image: https://hub.docker.com/r/intelliops/backintime/

PR #1020, opened 2019-07-16
Update it.po

albanobattistella:


Germar:
Thanks. I uploaded this in https://translations.launchpad.net/backintime which is the translation service for BiT. The 'po' files in this repository will get overwritten by files from Launchpad

PR #1021, opened 2019-07-17
Fix translation of "Saving permissions..." in en_CA.po & en_GB.po

MPvHarmelen:


Germar:
Thanks. I fixed this in https://translations.launchpad.net/backintime which is the translation service for BiT. The 'po' files in this repository will get overwritten by files from Launchpad

Issue #994, opened 2019-05-04
Files in snapshot with the wrong permissions


colinl:
Using version 1.2 from stable ppa, on Ubuntu 19.04, backing up to local drive, sometimes files are being written to the snapshot with the wrong permissions.  This makes itself known as after adding files and doing a snapshot if another snapshot is performed then sometimes files are modified showing permission differences.  I have analysed one of these and see that for the orginal snapshot the permissions in fileinfo.bz2 do not match those in the written file, so I see
```bash
$ ls -l 20190504-162212-796/backup/home/colinl/tobackup/gitrepos/bin.git/objects/08/e1c829f014423d7eaa09b2e48643f32eff54b5 
-rw-r--r-- 1 colinl colinl 185 May  4 16:20 20190504-162212-796/backup/home/colinl/tobackup/gitrepos/bin.git/objects/08/e1c829f014423d7eaa09b2e48643f32eff54b5
```
but
```bash
$ cat 20190504-162212-796/fileinfo.bz2 | bzip2 -d | grep e1c829
33060 colinl colinl /home/colinl/tobackup/gitrepos/bin.git/objects/08/e1c829f014423d7eaa09b2e48643f32eff54b5
```
and I think 33060 maps to 100444 which is r--r--r-- which is what the source file is.
After running the second backup fileinfo still has 33060 but the permissions in that backup are r--r--r-- as they should be.
I can make the log for the first backup available if necessary but I would rather not do it publicly.  There is nothing obviously odd for the entry for that file.

colinl:
Further information, it appears that it is all the files with permissions r--r--r-- that it is getting wrong, at least having looked at a selection of failing ones they all have those permissions, and I can't see any with those permissions that have not failed.  I will add just one file with those permissions and see if it fails reliably...
Hmm, no not that, having added just one file it went in ok.. It must be something more subtle.

colinl:
Yet further information, it is to do with r--r--r-- files.  It is necessary to close BiT and re-open it to trigger the problem, and in fact with those files often, but not always, if I open BiT and do a backup it will flip the permissions on those files between rw-r--r-- and r--r--r--. It is not only the first time after adding the file that it gets it wrong.

rcjhawk:
This is still an issue.  Running Linux Mint 20, backintime version 1.2.1.  The 5TB /ext4 formatted backup disk is completely filling up, apparently because of this bug.  The main disk is 2TB, and only about 35% filled, with most files unchanged for months or years.

[...]

Issue #988, opened 2019-04-28
Version 1.2.0 works very slow


kosal75:
I have just updated from 1.1.24 to 1.2.0 (common, gnome, qt4), and making a snapshot takes hours. When downgrading to 1.1.24, all is normal (5-6 minutes, as always).

Ubuntu 18.10 with kernel 4.18.0-18.

BelladonnavGFH:
Hi Germar, hope all is good!
Like @kosal75 I too got updated today, and ran into many issues. These were:

- My previous snapshots (over 2 years worth) were no longer recognised/ found. Although they were visible in the BiT-GUI, a new update triggered a new "virgin" rsync **of all files on all disks**.
- Having said that, $HOME/.config/backintime/config was found and recognised though and executed without errors.
- As @kosal75 writes, the rsync is extremely slow. With my backup disk on a type C connection, write speeds are blow < 1MB/s
- After the BiT/ rsync has been running for a while, my backup disk is unmounted. As a result BiT throws endless notification errors.

I too downgraded to one version lower available on my ppa:
`sudo apt install backintime-qt4=1.1.12-2 backintime-common=1.1.12-2 backintime-gnome=1.1.12-2`
`sudo apt-mark hold backintime-common backintime-qt4 backintime-gnome`

After this, all is good, and BiT performs with its previous accuracy, speed, and without flaws.

FYI: currently I'm on Xubuntu 18.04 LTS HWE (kernel 4.18.0-18).

Hope this helps...

kosal75:
I have found one more issue after downgrading: the cron settings were 
completely reset by 1.2.0.
--

pehlm:
Thanks **@BelladonnavGFH** for your tip! I haven't tried to do a backup with 1.2.0 and cannot say anything about the slowness, but the old backups wasn't, as you said, recognized any longer and I panicked. I'm also on Xubuntu 18.04 and it wasn't even possible to rezise the window, it was super big! Not possible to get it smaller. 
Thanks to you I have it working again. Even if I was forced to get an older version 1.1.12, the version before 1.2.0 was 1.1.24.
To the developers: I'm hoping you can get the new version to work, Back In Time is an extremely good backup-program!

BelladonnavGFH:
No thanks @pehlm. If it now works for you, perfect!
[...]

PR #1086, opened 2020-05-07
Make v1.1.x style permission handling configurable via expert settings.

b3nmore:


Codeberg-AsGithubAlternative-buhtz:
Why do we need this PR?

I think there was a reason why the current (1.2) version does handle the permission different then the old version (1.1.x). I do not know the reason but maybe @Germar does?

b3nmore:
> Why do we need this PR? Dear @b3nmore please provide more details.
> 
When upgrading from 1.1.x to 1.2 incremental backup does not work for the first backup, basically due to a change how rsync is called in 1.2. That means, that even if there is only one single file change in respect to the last (1.1.x) snapshot, 1.2 will make a full backup of all files. And that means it takes very long for a new snapshot and it will need nearly twice the disc space. With this PR you can continue to use your old snapshots for incremental backups even with 1.2.

> I think there was a reason why the current (1.2) version does handle the permission different then the old version (1.1.x). I do not know the reason but maybe @Germar does?
If I remember correctly there something about the reasoning in the referenced issue.

SirHackselot:
I uploaded a dpkg-repack'ed version of backintime with @b3nmore patch [here](https://github.com/SirHackselot/BackInTime-1.2.1-with-Permission-Option), in case it is of help to someone.

SirHackselot:
@b3nmore Thanks again for the patch. There seems so be at least two more lines that need to be changed in common/snapshots.py.

Line 447 `cmd_prefix = tools.rsyncPrefix(self.config, no_perms = False, use_mode = ['ssh'])`
Line  848 `cmd = tools.rsyncPrefix(self.config, no_perms = False)`

I removed the "no_perms = False" part of both and am doing some testing.

ti-nok:
I created the PR #1128 which removes the new permissions handling without adding any option.

aryoda:
The charm of this PR is that the user can configure the `--perms`/`--no-perms` behavior in the GUI settings while in PR #1128 the new permissions handling will be removed again.

aryoda:
I have just pulled the PR into a branch of my forked BiT.

The changed settings dialog looks like this:

PR1086 - changed settings dialog

The merge breaks at some unit tests that do not expect "no_perms" and needed additional work to fix/adjust the expected results of the affected unit tests:

```
$ python3 -m unittest  -b test/test_backintime.py
======================================================================
FAIL: test_local_snapshot_is_successful (test.test_backintime.TestBackInTime)
end to end test - from BIT initialization through snapshot
  File "/home/xxx/dev/backintime/common/test/test_backintime.py", line 153, in test_local_snapshot_is_successful
...
    cmd_prefix = tools.rsyncPrefix(self.config, no_perms = False, use_mode = ['ssh'])
TypeError: rsyncPrefix() got an unexpected keyword argument 'no_perms'
stdout: 
Back In Time
Version: 1.3.2 git branch '1086_perms_settings_GUI' hash '2186964'
```

As discussed in #1128 I do put this on hold until we can decide about the new permission system in general...

buhtz:
Dear @b3nmore ,

my apologize. I set this to "Draft" because it is outdated which is not your fault.

Your contribution won't disappear. We will take this into account while investigating and fixing #988.

Kind
Christian

buhtz:
I appreciate the work. But I vote to close the PR because this seems not to be the long term solution for the underlying problem.

The two major issues relevant here are #994 and #988 . The problem behind it is much bigger and need further investigation. The root of the problem is not even clear. I'm not convinced that that "v1.1.x style permission handling" is the right way compared to the current "1.2 style". And I won't try to support two different ways of style permission handling.

Myself I need much deeper understanding of how rsync handle things like this and how BIT do use rsync here. This will be ground work.

emtiu:
I agree with @buhtz here. I hope to see the day when we finally kill #994 and #988!

aryoda:
If we should ever revive this closed PR:

The code changes would also require to implement the fix for #994 which is (as known so far) triggered by `rsync --delete` causing to change permissions of read-only files under some circumstances...

PR #1137, opened 2021-03-09
Shortcuts for menu bar and tabs, toolbar items and remove exit from toolbar

### LINKED ISSUE #172: Do not show a Quit button on the toolbar
It is redundant. The user can close the app using the window manager's standard close button (or keyboard shortcuts), and it is against the GNOME HIG: http://library.gnome.org/devel/hig-book/stable/toolbars-appearance.html.en

---

Imported from Launchpad using lp2gh.
- date created: 2010-12-07T18:35:27Z
- owner: kiddo
- the launchpad url was https://bugs.launchpad.net/bugs/686695

### LINKED ISSUE #1104: Menus do not support accelerator keys
Accelerator keys do not appear to be available for selecting menus or menu items. (e.g., I'd expect, say, "Alt+H" to access the "Help" menu, and say "Alt+U" to invoke the "Up" action in the "View" menu.)

Version: 1.2.1
Source: Kubuntu 20.04 repositories

TheTimeWalker:
Adds accelerator keys to the menu bars and tabs in the settings screen. Most of the toolbar buttons now also have shortcut keys for easier keyboard access. This also removes the exit button from the toolbar and changes the shortcut to the more standard CTRL + Q instead of CTRL + W which is often used for closing tabs

Closes #1104, closes #172

### DIFF
--- a/qt/app.py
+++ b/qt/app.py
@@ -151,8 +151,9 @@ class MainWindow(QMainWindow):
         self.btnShutdown.setEnabled(self.shutdown.canShutdown())
         self.btnShutdown.toggled.connect(self.btnShutdownToggled)
 
-        self.btnQuit = self.mainToolbar.addAction(icon.EXIT, _('Exit'))
-        self.btnQuit.setShortcut(QKeySequence(Qt.CTRL + Qt.Key_W))
+        self.menuSnapshot = self.menuBar().addMenu(_('&Snapshot'))
+        self.btnQuit = self.menuSnapshot.addAction(icon.EXIT, _('Exit'))
+        self.btnQuit.setShortcut(QKeySequence(Qt.CTRL + Qt.Key_Q))
         self.btnQuit.triggered.connect(self.close)
 
         empty = QWidget(self)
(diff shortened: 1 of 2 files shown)


Codeberg-AsGithubAlternative-buhtz:
Sorry for the late answer and thanks for your PR.

@emtiu I vote for merge (incl. #1104 and #172). Small code modifications regarding the UI.

Codeberg-AsGithubAlternative-buhtz:
I agree to you with the close Button. It doesn't hurt anyone when the button stays where it is.

We can take care of that when we do a bigger redesign of the UI (someday 😄 ).

Technically I don't know what would be the next steps if the opener @TheTimeWalker won't do a new commit. You are "Mr. Rights" here.

TheTimeWalker:
This is how it looks like for KDE users:

image

The "Exit" button looks like a normal cancel/close button. I personally have to specifically remind myself to not push that button and I can think that people new to using the tool probably get confused by that button as well since it's against any guidelines and no other tool does that.

emtiu:
> This is how it looks like for KDE users:
> 
> image
> 
> The "Exit" button looks like a normal cancel/close button. I personally have to specifically remind myself to not push that button and I can think that people new to using the tool probably get confused by that button as well since it's against any guidelines and no other tool does that.

You're right, it doesn't even look like an "Exit this window" button. Combined with the fact that ⏸ (pause) and ⏹ (stop) buttons appear at the left end of the toolbar when a snapshot is being taken, this is rather confusing.

I'll just have to re-train my muscle memory ;)

Also: If somebody _did_ mistakenly push the "Shutdown after snapshot" button when they actually meant to just close the window, it wouldn't immediately do anything, but just visibly toggle the setting. Chances are rather slim that this would happen right in the instant before a snapshot is finished, leading to an unwanted shutdown. (And this is the comment everyone can angrily link to when that inevitably _does_ happen and is reported as a bug ;))

PR #1139, opened 2021-03-11
Match old and new rsync version format

### LINKED ISSUE #1138: Progress indicator in GUI not working in Linux Mint since BIT version 1.2 and Mint version 20
When I upgraded form Linux Mint 19.3 to Linux Mint 20, I upgraded BackInTime from 1.1.12 to 1.2.  Ever since then, the progress indicator in the GUI hasn't worked.  I'm now on Linux Mint 20.1/BackInTime 1.2.1 and the progress indicator still doesn't work.  The backups work, but you have no idea when they'll finish.  The BIT GUI just says "0%" and "working" for as long as the backup is running.  When the backup completes, the GUI switches abruptly to "done."  Any ideas on what to check?

TheTimeWalker:
Newer rsync versions have a different version style now, which breaks the regex match to add the progress2 capability. This adds a new regex match so that this gets correctly detected to show progress in for example the QT GUI

Fixes #1138

### DIFF
--- a/common/tools.py
+++ b/common/tools.py
@@ -529,9 +529,12 @@ def rsyncCaps(data = None):
         data = proc.communicate()[0]
     caps = []
     #rsync >= 3.1 does provide --info=progress2
-    m = re.match(r'rsync\s*version\s*(\d\.\d)', data)
-    if m and StrictVersion(m.group(1)) >= StrictVersion('3.1'):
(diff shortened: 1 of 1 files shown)


hannes101:
I can confirm that this patch works as intended. Incorporated it into the fedora package. Thanks!

Issue #1142, opened 2021-03-23
Scheduled backup ask for user input and non-error states are logged to syslog as ERROR


kortschak:
I have a Back in Time profile that stores to a NAS that is powered down for most of the time. The NAS is set to wake with a WOL packet that is sent by a user-callback which waits after the WOL packet has been sent then returns a success status if the NAS comes up within the timeout. This all works very nicely since #654 was fixed.

However, the backup to the NAS is scheduled for the middle of the night and often in the morning I see a notification like so

![notify](https://user-images.githubusercontent.com/275221/112221328-3732b900-8c77-11eb-9fe5-5c0a8476a56e.png)

despite the backup having succeeded. This is due to this https://github.com/bit-team/backintime/blob/9310acc30362907c7d14f3c8c31a0e4a366d44f5/common/snapshots.py#L669-L674 which makes sense in the context of attended backups, but not in the context of unattended scheduled backups where there is no user to follow the instructions. In the later context it leaves doubt in the mind of the user as to whether the backup succeeded. The issue appears to arise due to a race between `backintime` and the remote being mounted by the system since the user-callback waits until the NAS is up before it returns. `backintime` iterates a test for the presence of the remote mount each second, but first notifying the user that the remote mount is not present.

In looking into this issue, I found that `backintime` also logs as an ERROR to syslog for checks where the remount mount is not present (actually it checks whether it is not a directory): https://github.com/bit-team/backintime/blob/9310acc30362907c7d14f3c8c31a0e4a366d44f5/common/config.py#L1356-L1360 This results in misleading ERROR log lines in syslog.

```
Mar 22 01:00:02 client backintime (user/2): INFO: Lock
Mar 22 01:01:52 client backintime (user/2): INFO: user-callback returned 'user-callback: 2021/03/22 01:00:20 sending wake packet#012user-callback: 2021/03/22 01:01:52 server ready'
Mar 22 01:01:53 client backintime (user/2): INFO: mount ssh: user@server.local:/data/backups on /home/user/.local/share/backintime/mnt/XXXXXXXX/mountpoint
Mar 22 01:01:53 client backintime (user/2): INFO: Take a new snapshot. Profile: 2 remote
Mar 22 01:01:53 client backintime (user/2): ERROR: /home/user/.local/share/backintime/mnt/2_28727/backintime/client/user/remote does not exist
[...]

PR #1143, opened 2021-03-23
Better error logging in canBackup()

kortschak:
Please take a look.

Updates #1142.

### DIFF
--- a/common/config.py
+++ b/common/config.py
@@ -1353,10 +1353,13 @@ class Config(configfile.ConfigFileWithProfiles):
         if not self.isConfigured(profile_id):
             return False
 
-        if not os.path.isdir(self.snapshotsFullPath(profile_id)):
-            logger.error("%s does not exist"
-                         %self.snapshotsFullPath(profile_id),
-                         self)
+        path = self.snapshotsFullPath(profile_id)
+
+        if not os.path.exists(path):
+            return False
+
+        if not os.path.isdir(path):
+            logger.error("%s is not a directory"%path, self)
             return False
 
         return True


Codeberg-AsGithubAlternative-buhtz:
I vote for merge.

But this still reminds me to one of my private (but far away) TODOs for replacing (nearly) all of that `config.py` code with Pythons standard packages like `configparser` and `pathlib`.

emtiu:
I think this needs more discussion, over in the issue #1142. The main question is: Under which circumstances should BiT warn about a non-existing backup destination? Right now, I don't have enough of an overview to judge how much this is a bugfix, and how much a true behavior change.

kortschak:
This is a partial fix and replaces clearly incorrect logic for determining whether a directory is present.

buhtzz:
@kortschak Can you summarize the background of that PR for us please. It is not totally clear what happens here and we try to avoid breaking changes before the next release.

This PR also misses a unittest. I would like to have a unittest that proof the bug you try to fix. It means a unittest that would fail without your PR.

buhtzz:
> IMO the code is self-documentingly simple and not worth a unit test.

But the context is missing. That is what comments are for.

Each code need a to be tested. If the code is simple then you can test it. If it is complex then you made something wrong and the code becomes untestable. Simple and easy code only can have good tests. And a unittest would also be able to explain your goal because a unittest is also kind of a documentation.

I still ask for a summarize of the whole Issue the problem, the goals, etc.
Help us to understand.

In the current state I have to invest a lot of time to investigate that myself. I don't have that time because of other priorities related to BIT. I appreciate the time your invested in this but you also have to help us.

kortschak:
The code in full is
```
        def canBackup(self, profile_id = None):
        """
        Checks if snapshots_path exists
        """
        if not self.isConfigured(profile_id):
            return False

        path = self.snapshotsFullPath(profile_id)

        if not os.path.exists(path):
            return False

        if not os.path.isdir(path):
            logger.error("%s is not a directory"%path, self)
            return False

        return True
```
The comment at the top of the function is context. The linked issue explains the broader context. Quoted here

In looking into this issue, I found that `backintime` also logs as an ERROR to syslog for checks where the remote mount is not present (actually it checks whether it is not a directory): https://github.com/bit-team/backintime/blob/9310acc30362907c7d14f3c8c31a0e4a366d44f5/common/config.py#L1356-L1360 This results in misleading ERROR log lines in syslog.

```
Mar 22 01:00:02 client backintime (user/2): INFO: Lock
Mar 22 01:01:52 client backintime (user/2): INFO: user-callback returned 'user-callback: 2021/03/22 01:00:20 sending wake packet#012user-callback: 2021/03/22 01:01:52 server ready'
Mar 22 01:01:53 client backintime (user/2): INFO: mount ssh: user@server.local:/data/backups on /home/user/.local/share/backintime/mnt/XXXXXXXX/mountpoint
Mar 22 01:01:53 client backintime (user/2): INFO: Take a new snapshot. Profile: 2 remote
Mar 22 01:01:53 client backintime (user/2): ERROR: /home/user/.local/share/backintime/mnt/2_28727/backintime/client/user/remote does not exist
Mar 22 01:01:54 client backintime (user/2): ERROR: /home/user/.local/share/backintime/mnt/2_28727/backintime/client/user/remote does not exist
Mar 22 01:01:57 client backintime (user/2): INFO: Call rsync to take the snapshot
Mar 22 02:23:40 client backintime (user/2): WARNING: Command "rsync --recursive ...
Mar 22 02:23:40 client backintime (user/2): WARNING: CONTINUE: --exclude...
Mar 22 02:23:40 client backintime (user/2): INFO: Save config file
Mar 22 02:23:40 client backintime (user/2): INFO: Save permissions
Mar 22 02:26:27 client backintime (user/2): INFO: Create info file
Mar 22 02:26:31 client backintime (user/2): INFO: [smart remove] remove snapshots: [20210315-010002-225]
Mar 22 02:46:50 client backintime (user/2): INFO: unmount ssh: user@server.local:/data/backups from /home/user/.local/share/backintime/mnt/XXXXXXXX/mountpoint
Mar 22 02:46:50 client backintime (user/2): INFO: Unlock
```
The code is simple, constructing a test to demonstrate that this works as it should is not.

If you don't want the change, that's fine, I'll close the PR.

emtiu:
Dan, thank you for your patience and clarity in dealing with this issue, and taking the time to explain your troubles and your proposed solutions. I've reviewed your code, and I'm certain that it should work as intended and not introduce new problems.

I am however, deeply puzzled as to **_why_** this code change actually helps with your problem. I think it's worth discussing, if only to better understand the code (which we're now maintaining, but didn't write) and the circumstances (which occur in your setup).

First off, let's agree that (as you explained in https://github.com/bit-team/backintime/issues/1142#issuecomment-1259106730), you're dealing with two separate annoyances:
## 1. Spurious desktop notification
You're receiving a desktop notification that says `Can't find snapshots folder. If it is on a removable drive […]` every 30 seconds, when in fact your `user-callback` script is still _in the middle of waiting for the NAS_ to come up. This is a race condition, and should be fixed.

However, that's **_not_** what this PR is about, and so I suggest that the discussion about this race condition continue in #1142.
## 2. Spurious ERRORs in syslog
This **_is_** what this PR is about.
### Cause of the problem
You're seeing syslog messages like this:
```
Mar 22 01:01:53 client backintime (user/2): ERROR: /home/user/.local/share/backintime/mnt/2_28727/backintime/client/user/remote does not exist
Mar 22 01:01:54 client backintime (user/2): ERROR: /home/user/.local/share/backintime/mnt/2_28727/backintime/client/user/remote does not exist
```
at a time _after_ the NAS has booted, as indicated by your log. The message is logged every second, but only twice. Whatever causes the message apparently resolves itself after three seconds.

The code that spawns the message is:
https://github.com/bit-team/backintime/blob/88d19d42e9128f60b8952a64ed1fff23bd28108f/common/config.py#L1356-L1359

Therfore, over a time period of <3 seconds, this check fails and returns `False`:
```
os.path.isdir('/home/user/.local/share/backintime/mnt/2_28727/backintime/client/user/remote')
```
### Solution in your PR
You split the original check `os.path.isdir(path)` up into two separate checks:
```python
if not os.path.exists(path):
    return False

if not os.path.isdir(path):
    logger.error("%s is not a directory"%path, self)
    return False
```
This has the benefit of correctly identifying the error condition where `path` points to a file (instead of a directory). However, I'm sure you'll agree that's **_not_** what happens in your setup.

I think the change that really eliminates the ERRORs in your syslog is the fact that **_the first check fails silently_**, and only the second check (which, in your setup, is never reached), logs an ERROR.

**_If_** your code logged both errors equally:
```python
if not os.path.exists(path):
    logger.error("%s does not exist"%path, self)
    return False

if not os.path.isdir(path):
    logger.error("%s is not a directory"%path, self)
    return False
```
I'm certain that you'd still be seeing `ERROR: [path] does not exist` messages (for a few seconds).
### Conclusion
I think it's fine to split up the `os.path.isdir(path)` check into two checks, but **_not_** to have just the first one fail silently, while the second logs an error.

I think what's really happening in your case is that:
1. **There really is an error**, causing both `os.path.exists(path)` and `os.path.isdir(path)` to fail, meaning that it's an error worth being logged to syslog, **_but_**
2. it resolves itself after ~3 seconds.

Now, you could argue that even if `canBackup` didn't log anything at all, the failure wouldn't really be silent, because the other functions calling `canBackup` still react to receiving a `False` in return.

This is true, but I think it's not a good reason to eliminate logging from `canBackup` altogether. Because as your very case shows, the messages in the syslog (and their timing!) really help to diagnose problems.
### Recommendation
If we were to merge your PR, I would add an equal error logging to the first check:
```python
if not os.path.exists(path):
    logger.error("%s does not exist"%path, self)
    return False

if not os.path.isdir(path):
    logger.error("%s is not a directory"%path, self)
    return False
```
This would slightly improve the quality of the error logging – but of course, **_you would be back to square one_**, because you'd be seeing the errors in the logs again (for 3 seconds).

Therefore, I suggest that you continue running a modified version of the code that eliminates error logging from `canBackup`, keeping your logs free of the error message during the seconds it takes to resolve itself.

## finally:
Whatever happens to `canBackup`, it's worth keeping in mind that **_all this should go away_** once the issue in #1142 is fixed, and backintime properly waits for the user-callback script preparing the mount (by booting the NAS).

So let's continue the discussion there :)

kortschak:
If we were to merge your PR, I would add an equal error logging to the first check:

```
if not os.path.exists(path):
    logger.error("%s does not exist"%path, self)
    return False

if not os.path.isdir(path):
    logger.error("%s is not a directory"%path, self)
    return False
```

This would slightly improve the quality of the error logging – but of course, you would be back to square one, because you'd be seeing the errors in the logs again (for 3 seconds).

This would be incorrect. Please take a look at the callers of `canBackup`. They error if after some number of iterations have been successful or on the first instance if it was a Qt-invoked call.

emtiu:
> Please take a look at the callers of `canBackup`. They error if after some number of iterations have been successful or on the first instance if it was a Qt-invoked call.

I see that you're saying: `canBackup` shouldn't log errors in cases where the snapshot will _eventually_ succeed (as happens for you). Instead, `canBackup` should leave erroring to the calling functions.

I'm saying: `canBackup` should _never fail silently_, because it's a noteworthy condition when `os.path.isdir(path)` (or `os.path.exists(path)`) fails, regardless of the timing and the condition of the calling functions. As your case shows, it helps debugging immensely.

In any case, it would be inconsistent to have `canBackup` log an error if `os.path.isdir(path)` fails, but not if `os.path.exists(path)` fails.

I see two ways to resolve this:
1. We might demote the logging of `canBackup` from an error to a non-error warning.
2. You might add another ~5 second wait to your user-callback script. This should eliminate the condition where `canBackup` runs into your mountpoint while the check still fails.

kortschak:
It's not failing silently, it's correctly answering the question that it as asked, "Can I back-up?" The decision about whether that needs to be an errorable case is down to the caller, unless `canBackup` can find a situation that will not change, in this case when there is a non-dir file where a dir-file is expected. This is what I wrote. In the absence of a language that does sane error handling, this is the best that can be achieved.

> In any case, it would be inconsistent to have `canBackup` log an error if `os.path.isdir(path)` fails, but not if `os.path.exists(path)` fails.

I disagree. First, they are not failures, they are answers. One is an expected case (a file path may not point to a file) and one is exceptional (the file is there but is it not the kind that is expected and we can't do anything about it). This makes them fundamentally different. If I were writing this in Go it would return a bool and an error, and the error would correspond to the loggin case here.

I have added a final wait to my wake code.

emtiu:
Yes, I see, your argument is sound.

I'm leaning towards merging your code, and adding a call to `logger` to the outside function. Currently, this code only spawns the gui warning, and no info to `logger`:
https://github.com/bit-team/backintime/blob/88d19d42e9128f60b8952a64ed1fff23bd28108f/common/snapshots.py#L668-L678

It would stop `canBackup` from logging errors, but still serve debugging equally well to have a purely informational message like:
```python
logger.info("Cannot start snapshot yet: %s is not accessible. Waiting one more second."%path)
```
 spawned from here. I'll think about it some more.

emtiu:
@kortschak Would you open this PR to edits from me (so your account enters the commit history), or would you prefer that I copy your code into a new commit of my own?

kortschak:
What change is needed?

emtiu:
> What change is needed?

The procedure is described here: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork

kortschak:
No, I'm asking what changes you want to make to the contribution. I'd prefer to make changes or discuss merits.

emtiu:
> No, I'm asking what changes you want to make to the contribution. I'd prefer to make changes or discuss merits.

I see, sure.

In the loop calling `canBackup`:
https://github.com/bit-team/backintime/blob/88d19d42e9128f60b8952a64ed1fff23bd28108f/common/snapshots.py#L675-L678

I'd like to add the following call _before_ the `sleep()` call:
```python
logger.info("Cannot start snapshot yet: %s is not accessible. Waiting 1 more second."%path)
```

In your modified `canBackup` function, I'd like to add a code comment to document the rationale behind the remaining error logging, like so:
```python
        if not os.path.isdir(path):
            # path exists, but is no dir: something's very wrong.
            logger.error("%s is not a directory"%path, self)
            return False
```

kortschak:
If you want those in the change I'll let you make the changes. I don't think those are necessary; log spam is what I'm trying to avoid and the commit message explains the rationale.

PR #1153, opened 2021-04-08
Use Link Color instead of lightGray as not to break theming

newhinton:
This removes the Qt.lightGray Color for the excluded list in the settings, which does not work well with KDE theming, especially dark themes. It uses the link-color instead, which gets themed and looks better (on KDE)

I am not sure if this is a good color choice, but i am not familiar enough with the Qt -framework.


newhinton:
It may be a good idea to edit the Header-color for the snapshot-timeline and folderselection in the main-ui aswell.

Issue #991, opened 2019-05-02
Snapshot Size


thwaller:
Enhancement Idea:
I was looking for an option to show the size of the snapshots, but could not. Is this something you may consider adding?

What I am looking to see is the scale of each snapshot. For example, Seeing a large snapshot shows me it is largely inclusive whereas a small size snapshot is more supplemental. Basically, a visual to full backups vs incremental or differential backups. I would find this useful in managing backup history kept.

colinl:
I too would find it very convenient to see the additional disc space used after each backup if that were possible.  Approximate would be good enough.

Germar:
I totally agree this would be a nice to have feature. But technically this is not possible. BiT creates snapshots with hard-links. So if you backup a file that didn't change BiT will just create an other link to the already stored data. If you remove the link, the data is still available. Only when you remove the last link to the data the space is considered to be available by the filesystem again.
So you can't count which snapshot uses the data and also you can't tell which snapshot was the first to use the data. And when you delete the first snapshot, you would have to know which snapshot the data is still attached to.
The only way to know the exact size of each snapshot would be to run something like `ncdu` which scans the snapshots in chronology order. I wasn't able to force `ncdu` to do so last time I checked. Also tools like `ncdu` take ages on huge backups with lots of snapshots...

thwaller:
It appears my understanding of the process in somewhat incomplete. Rather then getting into a discussion here, is there a document you have that outlines more specifically the methodology of the backup process? I would rather start fresh as my mind is used to the norm of a full backup followed by change based backups tied to the last full backup.

Germar:
This [FAQ](https://github.com/bit-team/backintime/wiki/FAQ#how-does-snapshots-with-hard-links-work) does shed a light on how snapshots with hard-links work

thwaller:
Is it possible to mark which snapshots are full backups?

Germar:
There is no 'full backup'. BiT only sync those files which are not already in the previous snapshot. All others will be hard-linked. There is no way of knowing if no files at all where hard-linked as this is done by `rsync`

DonEdwards:
"Is it possible to mark which snapshots are full backups?"

All of them are.

thwaller:
> "Is it possible to mark which snapshots are full backups?"
> 
> All of them are.

[...]

Issue #406, opened 2015-10-11
Analyze/show disk space used by snapshots


Germar:
It would be very useful to be able to browse backups by actual size (with dereferenced hardlinks) and by files that were explicitly modified. I make full disk backup using BiT and would like to see what takes a lot of room (to possibly exclude some directories that only contain dynamic data). It would also be really helpful to know if a file is part of a specific snapshot or a hardlink from an older snapshot.

---

Imported from Launchpad using lp2gh.
- date created: 2012-04-13T13:49:26Z
- owner: xsmurfster+launchpad
- the launchpad url was https://bugs.launchpad.net/bugs/980814

Germar:
(by danleweb)
From command line you can use "du" (disk usage) to get the size "on disk" used by a file or directory.
This tool handle correctly hardlinks.

Regards,
Dan

Germar:
(by tmetro+ubuntu)
"Disk Usage Analyzer" is one of several GUI tools that will show you graphs of how storage space is allocated among directories and files on a selected path. You can browse to your BiT target directory and run it.

I just tried this, and it apparently does intelligently report space used by hard linked files only once. If you look at the attached screen shot, you'll see my initial snapshot from November is taking up 50% (69 GB) of the relative space (138 GB total), then one from February is taking up 36% (49 GB), and then most of the subsequent snapshots are taking up 2% or less.

Although "Disk Usage Analyzer" thinks the February backup has a significant quantity of unique files, I don't expect this is reality, and I think it is more a side effect of the order in which "Disk Usage Analyzer" happened to process the files. When it runs across a hard linked file, I don't know whether it assigns the space it takes up to the first or last directory it is found in. And I don't know if it processes the directories in sorted order.

In order for the report to make most sense for BiT you'd want it to process the directories in a consistent order - lets say ascending alpha-numeric - and then assign space for linked files to the first directory in which they are encountered.

It would seem that with a few tweaks to the "Disk Usage Analyzer" source this could easily be accomplished, and then BiT could include a button that launches "Disk Usage Analyzer" pointed at the directory containing the snapshots for the current profile. It would be a fairly small amount of integration effort for a nice functionality gain.

Codeberg-AsGithubAlternative-buhtz:
@emtiu 
I was referenced to this Issue by https://github.com/bit-team/backintime/pull/1162

[...]

PR #1162, opened 2021-05-19
show filesizes in log

newhinton:
This modifies the logoutput of rsync to include filesizes in byte. This allows to quickly see which files may bloat your backup.

Drawbacks: 
 - sizes are not encrypted and are always readable.

Missing:
 - <del>currently the log cannot be sorted by filesizes</del>
 - <del>calculate and show size of all files in log-viewer</del>
 - <del>The log-header with the snapshotdate should stay on top</del>
 - <del>maybe show full size in log aswell?</del>
 - behaviour of old logs is strange when the sorting is used, since all the items are zero bytes in length
 - behaviour of continued snapshot-log is strange.

this may help with #991, #406

### DIFF
--- a/common/snapshotlog.py
+++ b/common/snapshotlog.py
@@ -22,6 +22,7 @@ import gettext
 import logger
 import snapshots
 import tools
+import humanfriendly
 
 _=gettext.gettext
 
@@ -80,7 +81,6 @@ class LogFilter(object):
             str:        decoded ``line`` or ``None`` if the line was filtered
             int:        size of the file in bytes, ``None`` if no size was found.
         """
-
         has_size = False
         original_line = str(line)
         size = 0
@@ -88,7 +88,6 @@ class LogFilter(object):
             has_size = True
             start = line.find("[", 3)+1
             end = line.find(" bytes]")
-            import humanfriendly
             size = int(line[start:end])
             size_hr = " ["+humanfriendly.format_size(size)+"]"
             line = line[0:start-2]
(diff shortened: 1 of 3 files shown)


newhinton:
apt install python3-humanfriendly is required for now

Codeberg-AsGithubAlternative-buhtz:
Thanks for your contribution.
Can you please show an example output of the log file.

Does it only show "Bytes" or also "MB" and "GB" etc if needed?

Codeberg-AsGithubAlternative-buhtz:
It would help me the best if you could show me an before and after log output to visualize the difference between the "original" logoutput and your logoutput.

newhinton:
Sadly i have not used bit for quite a while, so i dont have a working dev setup or screenshots, sorry. However, i just added the conversion, just running this code and reviewing the log should get you a comparison.

Also some of the issues with old logs are not fixed, and i dont intend to update this PR.

Codeberg-AsGithubAlternative-buhtz:
Thank you very much for your feedback.

@emtiu In that case I vote for Close without merge.

Other reasons are that IMHO this feature isn't in the scope of BIT and the concrete implementation (the PR) itself may not be the best solution to implement such a feature. Without diving deeper into it I don't like the dependency to `humanfriendly` for such a "simple" feature.

newhinton:
Since i dont use BIT anymore my opinion doesnt count for much, i have to disagree with this beeing out of scope. 

Improving ease of use and accessibility for everyone is always a worthwhile goal that should be pursued, and filesizes in bytes is not really easily understandable for everyone ;)

On the rest you are probably right

Codeberg-AsGithubAlternative-buhtz:
> Since i dont use BIT anymore my opinion doesnt count for much

Your opinion count as much as all the others, as mine and as Germar!

> but i have to disagree with this beeing out of scope.

You might be right. To be honest it isn't clear for me what your PR adds to BIT. That is why I asked for a screenshot and before/after-log-example.

> and filesizes in bytes is not really easily understandable for everyone ;)

I totally agree.

emtiu:
First of all, thanks for the work you put in, @newhinton! Since there's no easy way for you or us to test this new feature at the moment, I'd also like to close this PR.

**But** I think space usage is an important aspect, so out of the very similar issues #991 and #406, I'd like to keep at least #406 open, and I'll add your idea of filesize logging so we don't forget it :)

PR #1196, opened 2021-11-21
GUI: Improve scheduling section (was: Adding execution time restriction)

makoehr:
This allows to execute cron jobs only within specified hours during a day.


Codeberg-AsGithubAlternative-buhtz:
This is a big and nice PR. Thank's a lot for that.

Please be aware of the discussion in https://github.com/bit-team/backintime/issues/1232 especially https://github.com/bit-team/backintime/issues/1232#issuecomment-1146882621 .

buhtz:
Dear @makoehr ,
can you please add a screenshot of the modified SettingsDialog please.

Can you also please describe your use case. Why do you need such a restriction? What scheduling options do you use?

Note to me:
 - Remove the po-file modifications. Need to be done in a separate PR.
 - Remove the config-manpage modifications because this manpage will be autogenerated based on the docstrings in config.py.
 - Update the docstrings in config.py based on the config-manpage modifications.

I converted that into a _Draft_ because it is IMHO a workaround and not a solution. The  PR just adding a text field to the GUI.

But this PR can be a good start to re-design the Scheduling section of the GUI. There are some cron-related GUI applications around that could stand as a good example about how to represent the cron-concept in a GUI. But to my quick research I don't like most of them because they just move the crontab syntax in a GUI.

makoehr:
> Dear @makoehr , can you please add a screenshot of the modified SettingsDialog please.

Please find the screenshot attached.
> 
> Can you also please describe your use case. Why do you need such a restriction? What scheduling options do you use?

I use e.g. a time period between 6am and 9pm.

Screenshot

aryoda:
@makoehr Thanks a lot for your really valuable contribution!

I suggest
-  to rename `Execution time restrictions` to something like `cron jobs time restrictions (HOURS)` to make it more clear what it is
- add logic to check if the cron HOURS expression is valid (see https://crontab.guru/ for details).
  This logic should not sit in the GUI but in the profile processing class recognize invalid expressions in hand-crafted profile files too

buhtz:
Thanks for your feedback.

> > Can you also please describe your use case. Why do you need such a restriction? What scheduling options do you use?
> 
> I use e.g. a time period between 6am and 9pm.

But what is your schedule setting? What have you selected in the schedule dropdown menu on the first tab?

Aryoda: I'm interested in a re-design of the scheduling section. So I take this PR as an inspiration. I have some things in mind.

buhtz:
Let me add a further question about my approach. In short: I vote to remove the cron part from the GUI (but not from BIT itself).

# Cron & Anacron
I assume most of the readers here do know the difference. But I want to break it down again.

The BIT user can say _Every 4 Hours_ resulting in the `crontab` line `0 */4 * * *`.
image

The BIT user can also say _Repeatedly Every 4 Hours_ resulting in an `anacron` job. I wasn't able to find the file location where this job goes.
image

The first using `cron` do run only at specific time points. In our case it is `0:00`, `4:00`, `8:00`, ... every day. If the machine isn't running the jobs are missing and won't made up.
The second using `anacron` results in timespans of minimum 4 hours. The job do run after 4 hours have passed. If the machine isn't running after 4 hours, it is no problem because `anacron` will notice at next time the machine boots up that more than 4 hours have passed and the job is run.

In short: `cron` focus on specific time points and `anacron` on timespans.

# Represent that in a GUI to the user
My point is that not all users may know or understand the difference between cron and anacron. Even if they understand they shouldn't need to think about it. It is the job of a GUI to abstract that and make it easy to them.

## Cron & Anacron in the GUI
If we would represent both concepts in a GUI I would modify the terms like his:
 - Cron: `Run each 4th Hour` or more generally `Run each n-th Minute/Hour/Day/Week/Month`
 - Anacron: `Run after 4 Hours` or `Run after n Hour/Day/Week/Month`

I looked up at the difference between "every" and "each". To my understanding "each" do fit better here for the cron part.
We also could add some info labels or tooltips to make it more clear to the user what the results of the settings are.

## Anacron only in the GUI
It is IMHO not easy to understand the GUI if cron and anacron are in there. For some users it may result in headaches to figuring out the difference between "each n hour" and "after n hours".

To make it easier for the users I would vote to remove (or kind of hide) the cron part of that GUI.

# Additional settings
Because of backwards compatibility the cron functionality shouldn't be removed in BIT itself; only in the GUI.
I would like to add the ability for the users to write their own cron line. This would be like an expert option.

I will try to draw a Mockup with QtCreator soon.

# EDIT
I learn on [bit-dev](https://mail.python.org/archives/list/bit-dev@python.org/message/TWZDODKE4Y5KVMQQSBHRB7N62WRFSHSI/) mailing list that BIT do not use anacron but simulates the anacron behavior. BIT do start itself every 15 minutes and then check the last snapshot execution like anacron would do. This is an important info but do not interfer with my question if we could remove the cron-like scheduling options from the GUI.

buhtz:
Dear @makoehr ,
we would be glad if you could join our discussion at #1449 , where we have mockup for a new scheduling dialog. What do you think about it? How could your restricted-hours use-case could be realized in such a dialog.

buhtz:
Dear Mathias,
again, thank you very much for your contribution.

I will close this PR because your idea and your input go lives on in #1449 and won't be lost. We plan to completely redesign the schedule section in the GUI.

Issue #898, opened 2018-06-14
AttributeError: 'str' object has no attribute 'toUtf8'


dmsierra:
Backintime version 1.1.12

Select a file on "Files View", then click on the Snapshot button on "Files Toolbar", click on "Diff Options", just click "Ok" and got this error on console:

```
Traceback (most recent call last):
  File "/usr/share/backintime/qt4/snapshotsdialog.py", line 72, in accept
    diff_cmd = str( self.edit_command.text().toUtf8() )
AttributeError: 'str' object has no attribute 'toUtf8'
```

Workaround for me: remove the "toUtf8()" from that line and the next one.

Codeberg-AsGithubAlternative-buhtz:
Diving into the Qt docu I can't get it why `text()` return a `str` instead of a `QString` instance. The latter does have a `toUtf8()` method.

https://doc.qt.io/qt-6/qlineedit.html#text-prop

The Issue is a bit old. Can you still reproduce that problem?

If so please give us detailed information's about your environment and versions: Operating System, Plattform, Python version, Qt version.

Codeberg-AsGithubAlternative-buhtz:
Just for documentation
![image](https://user-images.githubusercontent.com/11861496/186417004-ce91f168-45a2-4c8f-9fec-c33c5e983d0c.png)

I tested with BiT 1.2.1 on Debian 11 (ARM64) and can reproduce that error.

The question why no one else had this problem before really could be that no one else before used that dialog. It is hidden behind a lot of clicks and behind toolbar buttons (in the "file view toolbar") that I never noticed before. ;) It may sound strange but I think this is the best explanation.

Issue #1232, opened 2022-03-05
Is this project dead? Taking stock of pressing issues, Discussion welcome


emtiu:
I love backintime, it's been the backbone of my multi-level, multi-device backup strategy for many years now. The backups are simple, resilient and portable. In this regard, I think it's much superior to borg, restic or bup (which absurdly seem to think that everything in the world should be stored in the form of a git repo).

But after upgrading my system and trying to set up backintime >=1.2, I'm shocked at the sorry state of it:

- The new default of `rsync --perms` is causing mayhem, most notably with #988 (which points to a design flaw) and #994 (which points to an implementation error). It's likely that many users are running into this issue, but cannot identify the cause and just give up, as probably happened for #1132.

  - #1086 and #1128 are hacks to restore the old behaviour, but neither has been picked up.

- backintime does not work with Python 3.10, as evidenced by #1210 and #1223. Python 3.10 is the default in the upcoming *buntu 22.04 LTS, so this is very significant.

  - There seems to be a trivial fix given in #1174, but it hasn't been picked up.

- There seems to be no maintenance of this github repository's issues and pull requests, as evidenced by:

  - trivial fixes like #1077 not getting picked up

  - silly issues like #1160 not getting closed

  - completely resolved issues like #1123 being left open

I love and rely on backintime enough to consider creating a fork to maintain basic functionality, but that won't help many others who install it from this official source. With backintime not even running on *buntu 22.04 LTS and the fix not being implemented, there's probably not much time left before distributions drop backintime completely for being unmaintained. That would be a great loss.

I'm willing to help fix the code, and also with basic housekeeping of this repo's issues and pull requests. But I'm not sure is there's anyone left here doing any actual maintaining, and so I'd like to know what the official team, or other loyal users are thinking before committing.

gsker:
I too rely on it. 
I was also wondering why I didn't see 1.3.1 in debian testing or impish --  
(1.3.1 contains the trivial fix for the MutableSet breaking change) and I found this likely cause for that:
    https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1003776
That doesn't address the github repo maintenance though.  

I wonder if @Germar or @dinoboy197 or @LanceGundersen  have the ability to delegate this maintenance...

emtiu:
> I wonder if @Germar or @dinoboy197 or @LanceGundersen have the ability to delegate this maintenance...

[...]

PR #1246, opened 2022-04-10
Fix: `str` don't have `toUtf8` method.

jhilmer:
`QLineEdit.text()` returns `str` which don't have `toUtf8` method.
Later an command is build up based on this input. I have tested
that it still works with unicode characters.

### DIFF
--- a/qt/snapshotsdialog.py
+++ b/qt/snapshotsdialog.py
@@ -73,8 +73,8 @@ class DiffOptionsDialog(QDialog):
         self.mainLayout.addWidget(buttonBox, 3, 0, 3, 2)
 
     def accept(self):
-        diffCmd = str(self.editCmd.text().toUtf8())
-        diffParams = str(self.editParams.text().toUtf8())
+        diffCmd = self.editCmd.text()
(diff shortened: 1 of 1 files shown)


jhilmer:
Fix for https://github.com/bit-team/backintime/issues/898

Codeberg-AsGithubAlternative-buhtz:
Thanks for your contribution. Please be aware of the discussion in https://github.com/bit-team/backintime/issues/1232 especially https://github.com/bit-team/backintime/issues/1232#issuecomment-1146882621 .

I wonder why this old code (`.toUtf8()`) could live so long. Do you have any explanation for that?

Codeberg-AsGithubAlternative-buhtz:
Please excuse my timidness about this two tiny lines. 😆 

After doing some research and testing I agree this is an reproducible issue and that PR does fix it.
I also checked the repo for more uses of `toUtf8()` but couldn't find them.

The reason why no one else before found it is just that this dialog is rarely or nearly never used by someone before.

So dear @emtiu please just merge it and close #898.

PR #1261, opened 2022-06-25
Update readme: known problems

AlexSchr:
As discussed on the mailing list.


Codeberg-AsGithubAlternative-buhtz:
Fine for me.

speedy-10:
It would be nice, if people would develop further BIT and the corresponding README :-).

Issue #648, opened 2016-10-18
[enhancement] [callback] add level 9 - all done


kdupke:
I learned to use the callback script to wake-on-lan the backup system, but miss a hook to power off the backup system after backup.

The last cal to the callback script is level 8, but this is before umount so I can't use it for poweroff.

That said I appreciate a callback level 9, when all is done just before or after releasing all stuff.

Germar:
What about starting your shutdown command detached with a delay of 30sec?

kdupke:
That is the workaround I'm using at the moment. But it does not handle errors very well.
It isn't a clean solution.
Looking at the code it looks like there is an 'unlock' at the end of the backup, which could be the point of the callback.release entry. Unfortunately I never did python yet, else I could propose code.

kdupke:
Just in case, today I tried to change the time of daily backup. I started backintime-qt4, opened the config dialog and tried to change the time.

However, when I open the config dialog it calls the callback on level 7 and when clicking OK it calls callback with level 8.

So both are not really suitable for a WOL and Powerdown. Level 7 does not harm as it wakes up a server already up.  But level 8 is bad as it calls the delayed shutdown while backintime is still up.

Conclusion: backup power management needs separate treatment.

Germar:
Hmm. So we would also need to check if either GUI is showing that profile or a snapshot for the profile is still running before launching that new callback event...

kdupke:
Not really. Might be the idea of using user-callback for this feature isn't the right thing.

Or the handling within the GUI could be enhanced.

Let me start more basic, think about a remote server not available before powering on.

If I start the GUI, then I might not need the remote system because I want to change the config, or just check something.

So the process could be:
GUI:
Start GUI, load local data, ask if the remote data (aka backups) should be loaded, if so start the remote system. If not, then only display what's local. Of course starting remote could be automated, showing a message 'starting remote' (of course, start up needs some time, a 'just user'  could assume something is wrong).

If I change something which has to be stored on the remote (AFAIK config change, right?), and the remote isn't available, ask for starting remote and wait for it to be done.

After all, when I leave the GUI and the remote was powered on, then power it off.

As for the command line this is much easier, right? Some of the commands need the remote to be available, in which case the remote needs to be powered on. And if powered on then power it off at the end (using callback does not really works because of the way subshells are handled and waited for an SSH to finish).

That said, might be the idea of using callback to fulfill remote power management is wrong?

Germar:
[...]

PR #1269, opened 2022-08-09
Feature: Add postUnmount plugin callback

tatokis:
This is useful for unmounting the remote drive over ssh after sshfs is unmounted.

If the remote drive is unmounted before sshfs with the existing callback, then sshfs can not be unmounted and the mountpoint shows up as
```
d????????? ? ?         ?            ?            ? mountpoint
```
until the remote drive is remounted.

The existing `unmount()` function was not renamed to `preUnmount()` in order to not break any existing plugins.


aryoda (commented):
@tatokis I am trying to understand how the implementation supports your described use case:

> useful for unmounting the remote drive over ssh after sshfs is unmounted.

Your proposed implementation introduces a new `postUnmount()` method but it is always called together with the `unmount()` user-callback method:

https://github.com/bit-team/backintime/blob/76a4b1bc1e0df8344069c15f5ec9b13cad688243/common/mount.py#L157,L179

So I am trying to understand why the extra `postUnmount()` is required at all.
Could it suffice for your use case to implement the unmount of sshfs and the remote drive together (sequentially) directly in the `unmount()` user-callback script only?

tatokis:
@aryoda The key is that it's called after `backend.umount()`. That's when backintime unmounts the sshfs. The end user doesn't have control of the sshfs mount as it's handled by backintime itself.

The issue is that the remote drive that the sshfs points to would be unmounted before the sshfs itself if done in the unmount callback, rendering the sshfs non-unmountable afterwards by BIT.

The only way other way would be to try to detect which sshfs mountpoint BIT is using in the unmount callback, and manually unmount it before BIT tries to, but this is difficult and failure prone at best.

aryoda:
THX! So the problem I think is that

https://github.com/bit-team/backintime/blob/76a4b1bc1e0df8344069c15f5ec9b13cad688243/common/mount.py#L178

is called by BiT **after** `self.config.PLUGIN_MANAGER.unmount(self.profile_id)` (=> user-callback)
which is strange I think because it should be called after above unmount so that we don't need a `postUnmount` at all.

I am not sure if there is already a good unit test to cover these lines of code with sshfs so I don't dare to change this at the moment.

I will come back to look deeper into this PR once I have got more time left (and after stabilizing BiT by fixing the most urgent bugs).

buhtz:
I did some research about the basic functionality and behavior of user-callback. As a side-product I created some unit-tests (#1639) but they won't help solving this actual PR.

To my knowledge user-callback behavior is not covered by any of the current tests. Of course the code is executed in some tests (and covered in a technical way) but not tested.

IMHO the **behavior to test** would be if the callback reasons/steps are called on the right position on the timeline after/before some other specific steps of the taking a snapshot process.

My assumption is that this is nearly impossible in the current state of the productive code. The code is not isolated enough. And especially in testing user-callback behavior to many elements of code are involved. I see no way doing this with "unit tests" or "integration tests" in the near time.

A "dirty" approach could be to do a full snapshot run in a test and then parsing the log output for the expected messages (or their patterns) and if they appear in the correct order. We still have tests doing full backups. They do use some helper code from `test/generic.py`.

buhtz:
> is called by BiT **after** `self.config.PLUGIN_MANAGER.unmount(self.profile_id)` (=> user-callback) which is strange I think because it should be called after above unmount so that we don't need a `postUnmount` at all.
> 
> **Edit:** Or deprecate "unmount" and introduce "beforeUnmount" and "afterUnmount" (like "processBegin" and "processEnd")...

There is an old Issue (#648) from 2016 requesting also a "level 9" in the callback scripts to signaling that everything is done (after unmount).

buhtz:
If no one else will work on it I vote to transform this into an Issue, linking this (closed) PR as a potential approach to solve the Issue.

buhtz:
Closing this PR based on the comment above. Feel free to reopen and work further on it.
There is a follow-up issue #1869 to keep track of that contribution and not losing it.

Best regards,

PR #1280, opened 2022-09-02
Closes #1279: Add code coverage of unit tests as MAKE target

### LINKED ISSUE #1279: Development: Add code coverage of unit tests as MAKE target
Currently there are two `make` targets generated by `common/configure` that executes all unit tests with different verbosity levels:

```
make unittest    # same as make test
make unittest-v # same as make test-v
```

It would be great to know how good the existing python code is covered by the unit test ("code coverage").

I propose to extend `common/configure` to generate a new make target for this:

```
make coverage
```

I am working on this... (hope it works in the end ;-)

aryoda:
I have implemented my DEV feature request now.

To test the new `make` target:

```
cd common
./configure
make coverage
```

@emtiu Could you please check and merge my PR? THX!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,3 +37,10 @@ qt/Makefile
 #sphinx doc build
 common/doc-dev/_build/doctrees
 common/doc-dev/_build/html
+
+# PyCharm IDE project settings
+.idea
+
+# coverage result files (raw data and reports)
+# .coverage  # already ignored, see above
+**/htmlcov/**
--- a/common/configure
+++ b/common/configure
@@ -311,6 +311,16 @@ for v in "" "-v"; do
     printf "\n" >> ${MAKEFILE}
 done
 
+# For details about the "coverage" package args see:
+# https://coverage.readthedocs.io/en/stable/cmd.html
+printf "coverage:\n" >> ${MAKEFILE}
+printf "\tpython3 -m coverage run --source=.,../qt --omit */test/* -m unittest -b\n" >> ${MAKEFILE}
+printf "\tpython3 -m coverage report\n" >> ${MAKEFILE}
+printf "\t# To show an interactive HTML coverage report use:\n" >> ${MAKEFILE}
+printf "\t# python3 -m coverage html\n" >> ${MAKEFILE}
+printf "\t# firefox htmlcov/index.html\n" >> ${MAKEFILE}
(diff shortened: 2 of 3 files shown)


aryoda:
Please put this PR on hold, I
- forgot to put the common/plugins and qt/plugins folders under observation
- forgot to extend the `make clean` target (remove the coverage data and report files again).

Furthermore I am wondering what coveralls.io is measuring (click on the [coverage badge](https://coveralls.io/github/bit-team/backintime?branch=master) on the README file).
The reported value seems to high (may be caused by the unit test files which are considered too in the overall percentage which is false IMHO - the unit test coverage should be separated from the app code coverage because I only want to discover unit test fragments that never get executed due to an failure or logical coding error).

aryoda:
@emtiu OK, I have fixed the missing make clean target and included the forgotten code files from the "plugins" folders. You can merge now I everything is fine.
BTW: I have not rebased into one single "commit", is this a policy here?

Codeberg-AsGithubAlternative-buhtz:
> BTW: I have not rebased into one single "commit", is this a policy here?

Where good point! 😄 There is no policy yet. Just another TODO bullet on my list.

As an exception the current PR would be fine for me.
But because of that situation usually a PR should not created against `master` (or `main`). I would say we can discuss a "policy" on the mailing list when Germar granted us access as maintainers.

Until then and as a temporary "workaround" I would suggest to create PR's only against a new branch: So `aryoda:1279_make_coverage` should be created against `bit-team:1279_make_coverage`.

aryoda:
> I would suggest to create PR's only against a new branch: So `aryoda:1279_make_coverage` should be created against `bit-team:1279_make_coverage`.

Is there a reason to send pull requests against a new upstream branch (how could I create it BTW)?
IMHO this makes things more complicated until there is a good reason to do so. Just my 5 cents...

Codeberg-AsGithubAlternative-buhtz:
> Is there a reason to send pull requests against a new upstream branch

Yes. But let's discuss that later on the dev-bit list. There are several options ("git workflows") to handle such cases.

> (how could I create it BTW)?

Ah... I see it is not possible on GitHub or not possible because of our rights or BIT project settings. So not possible yet. Sorry my fault.

Technically it is (sometimes) possible to create a new branch when creating a PR.
image

PR #1293, opened 2022-09-13
Fix docstring with backslashes (DeprecationWarning: invalid escape sequence)

aryoda:
After a `./configure/make/make test` I get this warning:

```
backintime/common/mount.py:618: DeprecationWarning: invalid escape sequence \ 
```

The reason are several single backslashes in the docstring which may be interpreted as invalid escape sequence.

To avoid this follow the PEP instruction:

https://peps.python.org/pep-0257/#what-is-a-docstring

> Use r"""raw triple double quotes""" if you use any backslashes in your docstrings.

PS: The above warning is shown only after the first `make` since later `make`s do not rebuild unchanged artifacts like documentation.

### DIFF
--- a/common/mount.py
+++ b/common/mount.py
@@ -615,7 +615,7 @@ class MountControl(object):
             return False
 
     def createMountStructure(self):
-        """
+        r"""
         Create folders that are necessary for mounting.
 
         Folder structure in ~/.local/share/backintime/mnt/ (self.mount_root)::


common/mount.py
```diff
@@ -615,7 +615,7 @@ def mounted(self):
             return False
 
     def createMountStructure(self):
-        """
+        r"""
```
buhtzz, line 618: This could work but I wouldn't treat docstrings as r-strings.

Just replace backslashes.
This ` |\ <pid>.lock    ` should be  |- <pid>.lock    `.

Sorry for commenting. Currently I see not technical way to commit something into a "foreign" PR.

buhtzz (changes requested):
Please don't tread docstrings as r-strings. It is IMHO unusual and could irritate doc-tools.

Just replace the backslashes in the docstring: `\` to  `-`.

emtiu:
It seems we have to make a decision. On the one hand, the PEP257 that @aryoda linked explicitly recommends an r-string as docstring. On the other hand, @buhtzz comments that it might lead to other trouble.

I don't have enough experience to have an informed opinion, but my instinct says: If this is the _only_ offending docsctring in the project because of an unfortunate character, we might not gain much by making it the _only_ r-docstring in the project, potentially causing more upset later on ;)

So I'm in favor of replacing the `\` – unless @aryoda has a strong opinion here :)

aryoda:
Thx for your comments and painting skills ;-)

Removing the backslashes is the most defensive solution here, so I have changed my PR (see second commit).

PR #1296, opened 2022-09-16
fix: PEP8 in sshtools

buhtzz:
While I was working with something else, I was filed with this. 😄 
So I did some PEP8 improvement.

btw: `pytest` captured two warnings about escape sequences.
image

```
'echo \"rmdir \$(mktemp -d)\"; tmp3=$(mktemp -d); test -z "$tmp3" && cleanup 1; rmdir $tmp3 >/dev/null; err_rmdir=$?; '
              ^
'echo \"rmdir $(mktemp -d)\"; tmp3=$(mktemp -d); test -z "$tmp3" && cleanup 1; rmdir $tmp3 >/dev/null; err_rmdir=$?; '
```

In my understanding there is no need for escaping an `$` and all tests **on my machine** are green. But Travis has a problem with it and I don't understand why.

The second warning needs also further knowledge and investigation. Later...

### DIFF
--- a/common/sshtools.py
+++ b/common/sshtools.py
@@ -765,30 +765,37 @@ class SSH(MountControl):
                 f.write('foo')
 
             # check rsync
-            rsync1 =  tools.rsyncPrefix(self.config, no_perms = False, progress = False)
+            rsync1 =  tools.rsyncPrefix(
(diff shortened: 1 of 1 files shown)


buhtzz:
I tests didn't fail on my own system. I investigate further...

PR #1307, opened 2022-09-22
Refactoring and improvement: Config class is now a singleton

buhtzz:
The most important changes were made in `common/config.py` and `common/test/generic.py`.

As discussed the class `config.Config` now acts like a singleton. No matter that there are a lot of possible variants to implement the singleton design patter in Python I choose a pragmatic one, that is IMHO easy to understand for newbies, using less lines and is easy to maintain.

 - Create an instance as usual via `config.Config()`
 - If an instance is still there this will raise an Exception.
 - Get an existing instance via `config.Config.instance()`
 - This also will raise an Exception if no instance is there.

From an academic point of view this may not fulfill all aspects of the singleton pattern. But it prevents new contributors for misusing it.

I touched a lot of other files to use `config.Config.instance()` instead of `Config()`. I also improved the unittests because most of them did create their own `Config` instance (most of them via `generic.TestCaseCfg` base class). Now all unittest do destroy there instance if they have created that instance them selfs.

PS: You have to squash before merge Sorry for the 13 commits. This PR was intended to have only one commit but I misused `git`.


common/askpass.py
```diff
@@ -14,6 +14,9 @@
 #    with this program; if not, write to the Free Software Foundation, Inc.,
 #    51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 
+"""What is this modul about?
+"""
```
emtiu, line 18:
```suggestion
"""
Helper tool for piping passwords into ssh/sshfs and encfs.
Options are read from environ variables.
"""
```

common/backintime.py
```diff
@@ -656,14 +657,14 @@ def aliasParser(args):
...
+def getConfig(args, check=True):
     """
     Load config and change to profile selected on commandline.
 
     Args:
-        args (argparse.Namespace):
-                        previously parsed arguments
-        check (bool):   if ``True`` check if config is valid
+        args (argparse.Namespace): Previously parsed arguments
+        check (bool): Ff ``True`` check if config is valid
```
emtiu, line 667:
```suggestion
        check (bool): If ``True`` check if config is valid
```

common/configfile.py
```diff
@@ -15,19 +16,28 @@
 #    with this program; if not, write to the Free Software Foundation, Inc.,
 #    51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 
+"""This module implements a generic configuration file and low level
+operations on keys and their values.
+
+Development info: That module will be replaced in the future by Pythons own
+``configparser`` module or an alternative.
```
emtiu, line 23:
```suggestion
Development info: This module will be replaced in the future by Python's
 own ``configparser`` module, or some alternative.
```

common/mount.py
```diff
@@ -14,6 +14,9 @@
 #    with this program; if not, write to the Free Software Foundation, Inc.,
 #    51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 
+"""Modul docstring
+"""
```
emtiu, line 18: """High-level mount API for mounting, umounting, checks etc.
"""

common/test/generic.py
```diff
@@ -156,7 +156,16 @@ class TestCaseCfg(TestCase):
     """
 
     def setUp(self):
+        """
+        """
         super(TestCaseCfg, self).setUp()
+
+        # Testing this explicite instead of try-except because
+        # we doing unittesting here.
```
emtiu, line 164:
```suggestion
        # Testing explicitly instead of try-except
        # because we are doing unittesting here.
```

common/snapshots.py
```diff
@@ -57,9 +57,16 @@ class Snapshots:
     GLOBAL_FLOCK = '/tmp/backintime.lock'
 
     def __init__(self, cfg = None):
-        self.config = cfg
-        if self.config is None:
-            self.config = config.Config()
+        """
+        """
+        # This raise an exception if not config is loaded
```
emtiu, line 62:
```suggestion
        # This raises an exception if no config is loaded
```

common/snapshots.py
```diff
@@ -2601,7 +2608,7 @@ def lastSnapshot(cfg):
     if sids:
         return sids[0]
```
emtiu, line 2610:
```suggestion

#commented out when implementing config.Config as a singleton
```

common/sshMaxArg.py
```diff
@@ -22,56 +22,92 @@
...
+#                         help='Start checking with MID arg length')
+
+#     args = parser.parse_args()
+
+#     import config
+
+#     cfg = config.Config()
+#     mid = maxArgLength(cfg, args.MID)
+
+#     reportResult(cfg.sshHost(), mid)
```
emtiu, line 113: Can we be reasonably sure that this commenting-out doesn't break anything?

common/test/test_backintime.py
```diff
@@ -19,18 +19,42 @@
...
+    def tearDown(self):
+        config.Config._instance = None
+
+    def test_argument_diagnostics(self):
+        """Command-line argument '--diagnostics'"""
+
+        output = subprocess.check_output(['./backintime', '--diagnostics'])
+
+        self.assertIsNotNone(output)
+        self.assertTrue(output)  # len greather then 0
```
emtiu, line 47:
```suggestion
        self.assertTrue(output)  # len greather than 0
```

common/config.py
```diff
@@ -32,23 +33,33 @@
...
 
 gettext.bindtextdomain('backintime', os.path.join(tools.sharePath(), 'locale'))
 gettext.textdomain('backintime')
 
 
 class Config(configfile.ConfigFileWithProfiles):
+    """Represent the configuration of Back In Time.
+
+    That class provides high level operations on the configuration and some
+    additional information's. It also does handle the cronfile.
```
emtiu, line 52:
```suggestion
    This class provides high level operations on the configuration and some
    additional information. It also handles the cronfile.
```

common/config.py
```diff
@@ -76,87 +87,157 @@ class Config(configfile.ConfigFileWithProfiles):
...
+        #       f'{config_path}, data_path={data_path})')
+        # import inspect
+        # Config._buhtz.append(inspect.stack()[1])
+
+        # Exception when an instance exists
+        if __class__._instance:
+            # # DEBUG
+            # for e in Config._buhtz:
+            #     print(e)
+            #     print('°'*70)
```
emtiu, line 226:
```suggestion
```

common/test/test_config.py
```diff
@@ -18,151 +18,305 @@
...
+            config.Config
+        )
+
+
+class TestConfigSnapshots(generic.TestCaseCfg):
+    """More high level tests for the `Config` class using pre-setups via
+    deriving from classes in `generic` module.
+
+    Development note: It is not clear for me (buht) why this is related to the
+    `Config` class.
```
emtiu, line 76:
```suggestion
    Development note: It is not clear for me (buhtz) why this is related
     to the `Config` class.
```

emtiu (changes requested):
Hauptsächlich Typos, eine inhaltliche Frage (zur Änderung an `sshMaxArg.py`)

common/mount.py
buhtzz: It is a comment not a file modification. Not sure how the system reacts when I commit on top of your suggested changes. ;)

I will keep that in mind and modify it later, OK?

buhtzz: I edited that file directly in GitHub and added the commit to my branch. Seems like it worked.

common/sshMaxArg.py
buhtzz: I check the whole code via "grep" and searched for `subprocess` based calls of `sssMaxArg.py`.
I also checked the start scripts in the bin folder.
I couldn't find anything.

I assume that Germar or someone else had the idea first to implement that as an external tool and call it via `subprocess` when needed. But then it turned out to be usable directly via `import sshMaxArg`. But the `__main__` was left here.

We keep the code as a comment and not deleting it.

There are a lot of other files having there own `__main__` which makes not much sense in the current days. For example the unittest files. I think someone just wanted to run only one test file directly via `python test_foo.py`. With packages like `unittest` and `pytest` there is no need for this anymore.

aryoda:
I raise concerns that a singleton config does work correctly since I see
- multi-threaded code in the qt GUI (we have no unit tests for the GUI)
- (parallel) async backups where each running backup possibly requires a different config instance.
  At the first glance it looks OK since the config is only used to spawn a new sub process
  but not to reconcile the sub process results into the same config instance but I have not further checked this

See eg. the code when the "Take snapshot" button is clicked:

https://github.com/bit-team/backintime/blob/b14a98fc8aa24c6e51be0c7cd167b24a298da28a/qt/app.py#L873-L875

and

https://github.com/bit-team/backintime/blob/b14a98fc8aa24c6e51be0c7cd167b24a298da28a/common/backintime.py#L48-L80



> Get an existing instance via config.Config.instance()
> This also will raise an Exception if no instance is there.

I personally would prefer if  the `(get)Instance()` function does **not** raise an exception
but create a new (single) instance internally (stored in a private attribute) if none exists so far.
Exceptions must be tested or handled, this is just extra work.

buhtzz:
I still vote for merge but understand and agree with some of your concerns.

Most of the time the config file is just read. In that case it doesn't matter how often it is opened. In fact in the current situation before that PR **dozens** of `Config` instances where created while unittesting in one BIT instances and other BIT instances created via `subprocess.Popen()` calls.

And that is the intention of that singleton to bring up other problems that might resist in the code.

>     * multi-threaded code in the qt GUI (we have no unit tests for the GUI)

See above. This isn't a problem. GUI is mutli-threaded by definition. And I assume you mean mutli-processing where `backintime` is called via `subprocess.Popen()`.

>     * (parallel) async backups where each running backup possibly requires a different config instance.

I don't see a problem. The term "async" is missleading in the code. It is multi-processing because a complete new backintime instances (and its own python interpreter) is created via `Popen()`.

I see no problem if a sub-process need another config that it should use it. This doesn't effect the singleton. Because the Config instances only exists in the context of one specific backintime instances. When you call `subprocess.Popen(['backintime'])` that new process does create its own config instance.

> See eg. the code when the "Take snapshot" button is clicked:
> https://github.com/bit-team/backintime/blob/b14a98fc8aa24c6e51be0c7cd167b24a298da28a/qt/app.py#L873-L875

That is not a real problem. But it points to refactoring.
Handing a whole config object to such a function is bad design. The function should get only that information's that it needs. The point is that other layers of the application and other threads/process should never touch the config instance directly. But currently it is that way. Yes it is easy in the first place just to hand over the whole config. But it is not a good design and cause a lot of trouble. That is one intention of that singleton approach: Reducing the trouble.

That is a big problem of the whole code. To many parts of the code do handling direct the config object. That cause a lot of problems, bad code, etc. Also that BIT does call itself via `Popen()` isn't a good idea. There are better ways today. But this problem isn't solvable at once. This is a step-by-step thing for the next years.


> > Get an existing instance via config.Config.instance()
> > This also will raise an Exception if no instance is there.
> 
> I personally would prefer if the `(get)Instance()` function does **not** raise an exception but create a new (single) instance internally (stored in a private attribute) if none exists so far.

I strongly disagree because this would undermine the very objective of that PR and the singleton. The `config.Config.instance()` is not a service to the user/developer. It is a guard who will hit on your fingers if you (or the "old" code you are using) forgot to instantiate a Config object. And `Config.__init__()` hit on your fingers when you try to instantiate your own Config object.

> Exceptions must be tested or handled, this is just extra work.

No. Exceptions like this never need to be tested. That is what exceptions are for. This is not a returning error code. When someone is doing this the whole BIT thing need to stop. That is why there is an exception. A situation like this should never happen.
Having an exceptions points you to the fact that you do something wrong or that some of the code that is involved did something wrong. Somewhere there the loading of a config file doesn't happen. Such an exception is not a usual "error" but a disaster. ;)

common/sshMaxArg.py
emtiu: Okay, I understand. Let's go ahead with that.

Before any next release, we should absolutely test this in a real-world sshfs setup to make sure (not just rely on unit tests).

buhtzz: Agree. I still collecting real-world-scenarios on my private todo-list. Because I use BIT only over SSH without something special.

emtiu (approved):

common/sshMaxArg.py
aryoda: The main function servers a (debugging and issue testing) purpose documented here:

https://github.com/bit-team/backintime/blob/b14a98fc8aa24c6e51be0c7cd167b24a298da28a/common/config.py#L552-L555

buhtzz: But I couldn't find a hint that `sshMaxArg.py` is every used like this.

The arg-length checks are done but via `import sshMaxArg`. Of course that ssh-marx-arg-thing also is an Issue for further investigation and refactoring. I was surprised to here that there is an argument limit. Maybe on older systems only?

aryoda: > But I couldn't find a hint that `sshMaxArg.py` is every used like this.

Neither did I and I think this is only done manually at the console by a human to diagnose SSH backup issues.
So I would keep this main function and put `python3 sshMaxArg.py <user@ssh-host>` on a checklist to diagnose SSH problems in issues (I didn't know until now that this functionality exists).

emtiu:
We've agreed to stash this refactoring in its own branch first, while the main branch receives mainly bugfixes.

PR #1313, opened 2022-09-26
pep8 and docstrings

buhtzz:
Nothing changed here except some codeformating and doscstrings.

I was diving deeper into the code and trying to understand "things". Now I have a headache and go to bed. 🌙

### DIFF
--- a/common/applicationinstance.py
+++ b/common/applicationinstance.py
@@ -15,7 +15,9 @@
 #    with this program; if not, write to the Free Software Foundation, Inc.,
 #    51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 
-"""modul docstring
+"""
+This module holds the ApplicationInstance class, used to handle
+the one application instance mechanism 
 """
 
 import os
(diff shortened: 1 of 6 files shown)


aryoda:
I really like unifying the code formatting to a defined standard and doing this in a "re-formatting only" PR (mixing format and functional changes into one PR makes
it difficult to review the code changes - you have to find them within the many changes - see eg. this example: https://github.com/bit-team/backintime/pull/1307/files ;-)

I saw a lot of code formatting issues too and I intentionally tried to ignore them because
fixing them may cause a lot of merge conflicts if someone is working on a branch that was created **before** the code reformatting.

**I suggest to do "global" code formatting only after a feature freeze (eg. when preparing a release candidate) to minimize the merge conflicts...**

PS: Local reformatting of lines of code that are changed during the implementation of a feature (or bug fix) are OK of course IMHO...

buhtzz:
I see your point and agree. I tried to implement something but ended up trying to understand the code and do reformatting.
Different from my earlier PR's this one is "isolated" and reformatting only.

The reformatting happens "automatic" because without I'm not able to understand code. I don't want to waste my time reading "not so beautiful" code. I can not ignore them. But I try to keep it isolated/atomic and don't mix up code formatting with implementation/fixing something in one PR. But in the current (quality) state of the code basis it isn't easy to fit that goals every time.

Merge conflicts are not a problem if they are only depending on formatting. They can be solved easielly.
And I try to avoid this by not opening to much PRs at once.

PR #1319, opened 2022-10-04
Add unit tests for backintime.takeSnapshotAsync()

aryoda:
Just a small step (increases the code coverage by far less than 1 % ;-)

The OuT ("object under test") is the said function `backintime.takeSnapshotAsync()` which is used by the GUI
and code coverage said it was not yet executed by any unit test...


aryoda:
ON HOLD - DO NOT MERGE.

Travis CI is making problems to produce the same test results as on local computer for my new unit tests:

```
FAIL: test_backup_async (test.test_backup.TestBackup)
:py:func:`backintime.takeSnapshotAsync`:
  File "/opt/python/3.6.12/lib/python3.6/unittest/mock.py", line 1183, in patched
    return func(*args, **keywargs)
  File "/home/travis/build/bit-team/backintime/common/test/test_backup.py", line 64, in test_backup_async
    self.assertEqual(Popen_mock.call_args.args[0][0], "backintime")
AssertionError: 'a' != 'backintime'
- a
+ backintime
```

I will investigate this in the next days (est. not before the weekend)...

buhtzz:
When you testing a function in the `backintime.py` module why not using `test_backintime.py` file for testing?

I'm not sure but it seems to me your tests are real pure unittests and not integration tests. I love that! ❤️ If this is really the case then you could just derive your test class from `unittest.TestCase` instead of `generic.SnapshotsTestCase`. The latter has a lot of side effects and is IMHO more then we need here.

aryoda:
@buhtzz THX for your hints, the `generic` inheritance drives my crazy (so bloated ;-) and I decided to take and modify the next-similar unit test which was for `takeSnapshot()`. I will look into it when I have a free time slot again...

buhtz:
The errors from TravisCI only happens with Python 3.6.

Because of [that SO post](https://stackoverflow.com/q/60105443/4865723) and the official docu about ['unittest.mock.Mock.call_args`](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.call_args) (see the _Changed in version 3.8_ at the end there) I had an idea.

I tried to run Python3.6 in a docker container but give up. I'm not a docker pro. :D

I would say this problem is Specific to Python 3.7 and earlier. In #1358 I still gave some arguments why we can drop support for Python 3.7 and earlier now.

Because of supporting Python 3.8 and younger I would consider that PR as approved.

buhtz:
The errors from TravisCI only happens with Python 3.6.

Because of [that SO post](https://stackoverflow.com/q/60105443/4865723) and the official docu about ['unittest.mock.Mock.call_args`](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.call_args) (see the _Changed in version 3.8_ at the end there) I had an idea.

I tried to run Python3.6 in a docker container but give up. I'm not a docker pro. :D

I would say this problem is Specific to Python 3.7 and earlier. In #1358 I still gave some arguments why we can drop support for Python 3.7 and earlier now.

Because of supporting Python 3.8 and younger I would consider that PR as approved.

buhtz:
The errors from TravisCI only happens with Python 3.6.

Because of [that SO post](https://stackoverflow.com/q/60105443/4865723) and the official docu about ['unittest.mock.Mock.call_args`](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.call_args) (see the _Changed in version 3.8_ at the end there) I had an idea.

I tried to run Python3.6 in a docker container but give up. I'm not a docker pro. :D

I would say this problem is Specific to Python 3.7 and earlier. In #1358 I still gave some arguments why we can drop support for Python 3.7 and earlier now.

Because of supporting Python 3.8 and younger I would consider that PR as approved.

buhtz:
Sorry, for the chaos and commits. Should be reset again to your initial version now.

Because I wasn't able to run BIT with Python 3.6 and Python 3.8+ on my local system (and docker) I tried to do this with TravisCI.
But here my problem is that `python unittest` don't show the `stdout` (`print()`) statements when a test passes with OK. But that is what I need to compare the output between Python 3.8 and 3.6. Because of that I tried to run the tests on Travis with `pytest` because it's `-s` switch does catch `stdout` and print it in every case. But I wasn't able to modify `configure` that way that it use `pytest`.

buhtz:
OK, it seems I got it now. Test is fixed and green lights on all Travis builds.

The cause of that problem was that `call_args` was modified since Python 3.8 and because of that behave different in our Travis builds with Python 3.6 and 3.10.

The solution is to use [`assert_called_once_with()`](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.assert_called_once_with) to test how a mock was called. This works in all Python versions the same way.

But `Popen()` is called by BIT with a second parameter `env=os.environ.copy()`. That dict is quit big and ordering of keys is not guaranteed. I tested this on Travis. Because of that I also mocked the return value here.

Thats it with the solution.

A secondary problem was the reason why that error was quit hard to break down. The "problem" with mock objects in Python is that they are really magic. You can write everything you want behind the dot. It will never raise an AttributeError. To make sure that a mock object offers the same public methods and nothing more than that the `autospecs=True` need to be used in the `patch()` function.

Good night folks. 🛌

buhtz:
Ah.... 🌞  I found `unittest.mock.ANY` and replaced the `os.environ.copy()` mock with it. Make it simple.

💤

buhtz:
@aryoda What do you think? Can we merge this? I ask because it is your own PR.
I see no risk. Just test code is involved.

I added a lot of commits. Maybe it would help and save your time if I summarize what I have changed in relation to your initial PR.

 - I modified the check for called arguments. Now `assert_called_once_with()` is used. That method of checking works in all Python versions.
 - I removed `@patch('time.sleep')` from each test method and put it on top of the test class. This way one patch-line affects all test methods.
 - Your own intention of the tests are not modified.

PR #1323, opened 2022-10-06
Fix DeprecationWarning about StrictVersion

buhtzz:
The class `distutils.version.StrictVersion` doing version comparison and its package `distutils` is deprecated. They are replaced by `packaging.Version` (see [PEP440](https://peps.python.org/pep-0440/)).

This is related to #1288 where I did the same but obviously missed some parts of the code. Not sure why I missed them. It seems to me that this DeprecationWarning doesn't occur with Python 3.9 but 3.10. Now I search via `grep` for the use of `distutils` package. I suspect that the deprecated `distutils` package isn't used anymore.

### DIFF
--- a/common/encfstools.py
+++ b/common/encfstools.py
@@ -22,7 +22,7 @@ import re
 import shutil
 import tempfile
 from datetime import datetime
-from distutils.version import StrictVersion
+from packaging.version import Version
 
 import config
 import password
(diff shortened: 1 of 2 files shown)


aryoda:
I also found only these two files using `distutils`. Should be fixed with this PR.

PR #1328, opened 2022-10-07
Improve logging for canBackup check

emtiu:
Fixes one half of #1142, closes #1143.

### DIFF
--- a/common/snapshots.py
+++ b/common/snapshots.py
@@ -677,7 +677,7 @@ restore is done. The pid of the already running restore is in %s.  Maybe delete 
                                     gettext.ngettext('Waiting %s second.', 'Waiting %s seconds.', 30) % 30,
                                     30)
                         for counter in range(30, 0, -1):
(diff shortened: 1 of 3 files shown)


common/snapshots.py
```diff
@@ -677,6 +677,7 @@ def backup(self, force = False):
                                     gettext.ngettext('Waiting %s second.', 'Waiting %s seconds.', 30) % 30,
                                     30)
                         for counter in range(30, 0, -1):
+                            logger.info("Cannot start snapshot yet: target directory not accessible. Waiting 1s.")
```
kortschak, line 680: I'd just like to point out that since backintime by default logs at INFO this is potentially 30 non-informative log lines in syslog. A more sensible approach would be to either 1. not log this or 2. if logging this is crucial (I don't believe it is since the failure that can happen is logged below) then count how many times the iteration has happened and only log if it is greater than 1.

Issue #1333, opened 2022-10-13
[Arch] Backintime fails to build


afettouhi:
Hi, I have an up to date Arch install and I am trying install the git version of backintime via the AUR using yay. But the install fails during unittesting for reason. Here is the output when it fails

OK
python3 -m unittest  -b test/test_backintime.py
E..
======================================================================
ERROR: test_diagnostics_arg (test.test_backintime.TestBackInTime)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/af/.cache/yay/backintime-git/src/backintime/common/test/test_backintime.py", line 181, in test_diagnostics_arg
    diagnostics = json.loads(output)
  File "/usr/lib/python3.10/json/__init__.py", line 346, in loads
    return _default_decoder.decode(s)
  File "/usr/lib/python3.10/json/decoder.py", line 337, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/usr/lib/python3.10/json/decoder.py", line 355, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)

----------------------------------------------------------------------
Ran 3 tests in 3.040s

aryoda:
Thanks for reporting this, I am already aware of this bug in another issue (no real duplicate):
See https://github.com/bit-team/backintime/issues/1233#issuecomment-1279248296

The current work-around to install from AUR is described here:
https://github.com/bit-team/backintime/issues/921#issuecomment-1276888138
It would be great if you report back here if it works or something else is missing...

aryoda:
Fix submitted via PR #1335 (also updates the AUR installation section in README incl. a link for a work-around to this issue)

afettouhi:
Still fails to build for me on Arch. Though the error is different now

`python3 -m unittest  -b test/test_backintime.py
Traceback (most recent call last):
  File "/home/af/.cache/yay/backintime-git/src/backintime/common/backintime.py", line 1190, in <module>
    startApp()
  File "/home/af/.cache/yay/backintime-git/src/backintime/common/backintime.py", line 507, in startApp
    args = argParse(None)
  File "/home/af/.cache/yay/backintime-git/src/backintime/common/backintime.py", line 568, in argParse
    args, unknownArgs = mainParser.parse_known_args(args)
  File "/usr/lib/python3.10/argparse.py", line 1859, in parse_known_args
[...]

Issue #1233, opened 2022-03-09
[Arch] Build test fails & installation is aborted


Kleysley:
I've been trying to install backintime on my Arch installation (kernel 5.16.13).
No matter if I build it from source or whether I try to install the gui or the cli version, the build always fails the same way.
Here is the console output:

```
Ran 19 tests in 0.029s
OK
python -m unittest  -b test/test_argparser.py
.....................
----------------------------------------------------------------------
Ran 21 tests in 1.990s

OK
python -m unittest  -b test/test_backintime.py
FF
Stderr:
/home/kleysley/.cache/yay/backintime/src/backintime-1.3.1/common/test/test_backintime.py:33: DeprecationWarning: Please use assertEqual instead.
  self.assertEquals("", subprocess.getoutput("python3 backintime.py --quiet"))

======================================================================
FAIL: test_local_snapshot_is_successful (test.test_backintime.TestBackInTime)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/kleysley/.cache/yay/backintime/src/backintime-1.3.1/common/test/test_backintime.py", line 68, in test_local_snapshot_is_successful
    self.assertEqual(proc.returncode, 0, msg)
AssertionError: 1 != 0 : Returncode: 1
stderr: Traceback (most recent call last):
  File "/usr/lib/python3.10/site-packages/dbus/bus.py", line 177, in activate_name_owner
    return self.get_name_owner(bus_name)
  File "/usr/lib/python3.10/site-packages/dbus/bus.py", line 361, in get_name_owner
    return self.call_blocking(BUS_DAEMON_NAME, BUS_DAEMON_PATH,
  File "/usr/lib/python3.10/site-packages/dbus/connection.py", line 652, in call_blocking
    reply_message = self.send_message_with_reply_and_block(
dbus.exceptions.DBusException: org.freedesktop.DBus.Error.NameHasNoOwner: Could not get owner of name 'net.launchpad.backintime.serviceHelper': no such name

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/kleysley/.cache/yay/backintime/src/backintime-1.3.1/common/backintime.py", line 1165, in <module>
    startApp()
  File "/home/kleysley/.cache/yay/backintime/src/backintime-1.3.1/common/backintime.py", line 517, in startApp
    args.func(args)
[...]

PR #1335, opened 2022-10-16
Fix bug: BiT does not start (DBus exception NameHasNoOwner)

aryoda:
Fix bug: BiT does not start with DBus exception NameHasNoOwner (partial fix for #921). Also fixes Arch Linux build error (#1233).

* Fix the BiT CLI side of the bug #921
* serviceHelper system D-Bus daemon issues are still under investigation
* Fix Arch Linux build test and installation failure (#1233)
* Add better documentation for installation on Arch using AUR package to README

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -2,19 +2,27 @@ Back In Time
 
 Upcoming Release
 * New feature: Command line argument "--diagnostics" to show helpful info for better issue support (#1100)
-* GUI change: remove Exit button from the toolbar (#172)
-* GUI change: define accelerator keys for menu bar and tabs, as well as toolbar shortcuts (#1104)
-* Desktop integration: update .desktop file to mark Back In Time as a single main window program (#1258)
+* GUI change: Remove Exit button from the toolbar (#172)
+* GUI change: Define accelerator keys for menu bar and tabs, as well as toolbar shortcuts (#1104)
+* Desktop integration: Update .desktop file to mark Back In Time as a single main window program (#1258)
 * Bugfix: AttributeError in "Diff Options" dialog (#898)
 * Bugfix: Settings GUI: "Save password to Keyring" was disabled due to "no appropriate keyring found" (#1321)
+* Bugfix: Back in Time did not start with D-Bus error
+          "dbus.exceptions.DBusException: org.freedesktop.DBus.Error.NameHasNoOwner:
+           Could not get owner of name 'net.launchpad.backintime.serviceHelper': no such name"
+           (fixes client-side part of #921 - system D-Bus part of the Udev serviceHelper is still under investigation).
 * Bugfix: Avoid logging errors while waiting for a target drive to be mounted (#1142, #1143, #1328)
-* Documentation update: correct description of profile<N>.schedule.time in backintime-config manpage (#1270)
+* Bugfix: [Arch Linux] AUR pkg "backintime-git": Build tests fails and installation is aborted (#1233, fixed with #921)
+* Documentation update: Correct description of profile<N>.schedule.time in backintime-config manpage (#1270)
 * Translation update: Brazilian Portuguese (#1267)
 * Translation update: Italian (#1110, #1123)
 * Translation update: French (#1077)
-* Testing: fix a test fail when dealing with an empty crontab (#1181)
-* Testing: fix a test fail when dealing with an empty config file (#1305)
-* Testing: numerous fixes and extensions to testing (#1115, #1213, #1279, #1280, #1281, #1285, #1288, #1290, #1293, #1309)
+* Testing: Fix a test fail when dealing with an empty crontab (#1181)
+* Testing: Fix a test fail when dealing with an empty config file (#1305)
+* Testing: Skip "test_quiet_mode" (does not work reliably)
+* Testing: Improve "test_diagnostics_arg" (introduced with #1100) to no longer fail
+           when JSON output was mixed with logging output (part of #921, fixes #1233)
+* Testing: Numerous fixes and extensions to testing (#1115, #1213, #1279, #1280, #1281, #1285, #1288, #1290, #1293, #1309)
 
 Version 1.3.2 (2022-03-12)
 * Fix bug: Tests no longer work with Python 3.10 (https://github.com/bit-team/backintime/issues/1175)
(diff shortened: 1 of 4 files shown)


aryoda:
Also closes #1333

buhtzz:
Is it the DBus exception that pollutes the JSON output? Something like this (quick & dirty handwritten)?

```

$ backintime --diagnostics
  File "/usr/lib/python3/dist-packages/dbus/bus.py", line 175, in activate_name_owner
    return self.get_name_owner(bus_name)
  File "/usr/lib/python3/dist-packages/dbus/bus.py", line 361, in get_name_owner
    's', (bus_name,), **keywords)
  File "/usr/lib/python3/dist-packages/dbus/connection.py", line 651, in call_blocking
    message, timeout)
dbus.exceptions.DBusException: org.freedesktop.DBus.Error.NameHasNoOwner: Could not get owner of name 'net.launchpad.backintime.serviceHelper': no such name

   "backintime": {
        "name": "Back In Time",
        "version": "1.3.2"
    },
    "host-setup": {
        "platform": "Linux-5.10.0-18-amd64-x86_64-with-glibc2.31",
    },
    "python-setup": {
        "python": "3.9.2 default Feb 28 2021 17:03:44 CPython GCC 10.2.1 20210110",
        ],
    },
    "external-programs": {
        "rsync": "3.2.3",
        "ssh": "OpenSSH_8.4p1 Debian-5+deb11u1, OpenSSL 1.1.1n  15 Mar 2022",
        "sshfs": "3.7.1",
        "encfs": "1.9.5",
        "shell": "/bin/bash",
        "shell-version": "GNU bash, Version 5.1.4(1)-release (x86_64-pc-linux-gnu)"
    }
}

```

I don't understand enough about DBus, the dbus issue and how it is involved with BIT. And I don't need to. :smile: I trust you at this point. From my understanding the DBus problem isn't fixed yet. Am I right?

Am I also right to say when it is fixed that the json output of `--diagnostics` won't be polluted anymore?

If I'm wrong cut the rest. ;)

# IMHO
The "problem" with this PR is that it kind of "deactivates" the integration test `test_backintime.py::TestBackInTime::test_diagnostics_arg()`.

That test is an integration test because it doesn't test one isolated function/method but a whole feature (`--diagnostics`). Does this feature work means: _Does `--diagnostics` give me JSON output._ Technically this means: _Does `json.loads(output)` raise a parsing exception or not?_ That's it. Nothing more.

On the other hand we test the content of the output. That is a unit test because we can do this quit isolated. We only test the function `common/diagnostics.py::collect_diagnostics()`. This is done in `test_diagnostics.py::Diagnostics::*`. Here we check the content of the JSON output. Are the right keys/items in it? etc

# Consequence
I understand that the test `test_backintime.py::TestBackInTime::test_diagnostics_arg()` does make problems because the dbus issue isn't fixed yet. Because of that I would suggest to really deactivate that test in the meaning of out-comment it with `# `. And add a link to the DBUS issue. We also could open an extra issue about Re-activating the text when dbus is fixed.

aryoda:
> Is it the DBus exception that pollutes the JSON output?

The problem is the added log output of the captured exception (the exception itself is handled and does not bubble up).
The old implementation was totally silent (swallowed the exception) which made it difficult to diagnose.

> Am I also right to say when it is fixed that the json output of --diagnostics won't be polluted anymore?

For this unit test I guess yes, but the new unit tests that eg. combine `--diagnostics --debug` would fail too (reason see below).

> The "problem" with this PR is that it kind of "deactivates" the integration test
> test_backintime.py::TestBackInTime::test_diagnostics_arg().
> That test is an integration test because it doesn't test one isolated function/method but a whole feature (--diagnostics).

I think the integration test is still done in my PR because `--diagnostics` is still executed (and eg. every RTE makes the test fail).
Only the assertion checks in the expected results are based on text strings now instead of parsed JSON.

> I would suggest to really deactivate that test in the meaning of out-comment it with # 

I'd prefer to keep it running due to its integrative character.

> We also could open an extra issue about Re-activating the text when dbus is fixed.

There will be no better final D-Bus fix on the BiT client-side (a connection error may always happen) and BiT was (theoretically) and is (now with this PR more reliably) able to work without it.

**Opening a new issue for the real underlying problem sounds good:**

Our logging system pollutes the normal (non-logging) stdout output with logging output, we need a clear separation here
(eg. logging to stderr instead of stdout) otherwise we can never separate normal from (partially random) logging output in unit tests. I think you have mentioned this problem too during the implementation of `--diagnostics` (when called together with `--debug`).

In the new issue we could also request new features
- for logging into a file via a new command line argument, env var or config file
- to add a new command line argument(s) for choosing the log level or do this in a conf file (currently we only have `--debug` and everything above is always logged to stdout)
- using the python standard logger instead of the custom implementation...

aryoda:
@buhtzz 

> Our logging system pollutes the normal (non-logging) stdout output with logging output, we need a clear separation here
> (eg. logging to stderr instead of stdout) otherwise we can never separate normal from (partially random) logging output in unit tests.

Thanks for tackling me ;-) I think I can change the unit test so that it uses JSON again by calling `subprocess.check_output` instead of `subprocess.getoutput` (the former contains only the `stdout`, the latter also the `stderr`).

I also just saw that the custom logger function already do use `stderr` **partially** and my new warnings are luckily written to `stderr`.

https://github.com/bit-team/backintime/blob/b4dd2d1737c5031d264d667c18f6987a1916762b/common/logger.py#L48-L53

Interestingly the info and debug output is written to `stdout` which should be `stderr` too IMHO as standard python does:

> If you call the functions [debug()](https://docs.python.org/3/library/logging.html#logging.debug), [info()](https://docs.python.org/3/library/logging.html#logging.info), [warning()](https://docs.python.org/3/library/logging.html#logging.warning), [error()](https://docs.python.org/3/library/logging.html#logging.error) and [critical()](https://docs.python.org/3/library/logging.html#logging.critical), they will check to see if no destination is set; and if one is not set, they will set a destination of the console (sys.stderr)

https://docs.python.org/3/howto/logging.html

I will update my PR!

aryoda:
> Interestingly the info and debug output is written to `stdout` which should be `stderr` too IMHO as standard python doe

As internal PoC I have just redirected our logger class output to `stderr` for ALL serverity levels and only one (!) unit test fails.

I think this is worth another PR ;-)

buhtzz:
Great to hear.

Btw: The replacement of `common/logger.py` with pythons own `logging` package is also on my todo list. Very important and will save a lot of work and will make other solutions incl. unittesting much easier to control.

PR #1337, opened 2022-10-18
Write all log output to stderr (do not pollute stdout)

aryoda:
* Affects logger.info and logger.debug
* complies now to the behavior of pythons's standard logging package

### Background

Our logging system pollutes the normal (non-logging) stdout output with logging output, we need a clear separation here
(eg. logging to stderr instead of stdout) otherwise we can never separate normal from (partially random) logging output in unit tests.

Standard python's logger package says:

> If you call the functions [debug()](https://docs.python.org/3/library/logging.html#logging.debug), [info()](https://docs.python.org/3/library/logging.html#logging.info), [warning()](https://docs.python.org/3/library/logging.html#logging.warning), [error()](https://docs.python.org/3/library/logging.html#logging.error) and [critical()](https://docs.python.org/3/library/logging.html#logging.critical), they will check to see if no destination is set; and if one is not set, they will set a destination of the console (sys.stderr)

https://docs.python.org/3/howto/logging.html

### DIFF
--- a/common/logger.py
+++ b/common/logger.py
@@ -23,8 +23,8 @@ import atexit
 import tools
 import bcolors
 
-DEBUG = False
-APP_NAME = 'backintime'
+DEBUG = False            # Set to "True" when passing "--debug" as cmd arg
+APP_NAME = 'backintime'  # TODO Duplicated code (see Config.APP_NAME)
 
 def openlog():
     name = os.getenv('LOGNAME', 'unknown')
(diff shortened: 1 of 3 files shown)


buhtzz (approved):

common/logger.py
```diff
@@ -55,14 +55,16 @@ def warning(msg , parent = None, traceDepth = 0):
...
-    print('%sINFO%s: %s' %(bcolors.OKGREEN, bcolors.ENDC, msg), file=sys.stdout)
+    print('%sINFO%s: %s' %(bcolors.OKGREEN, bcolors.ENDC, msg), file=sys.stderr)
     for line in tools.wrapLine(msg):
         syslog.syslog(syslog.LOG_INFO, 'INFO: ' + line)
 
 def debug(msg, parent = None, traceDepth = 0):
     if DEBUG:
         msg = '%s %s' %(_debugHeader(parent, traceDepth), msg)
-        print('%sDEBUG%s: %s' %(bcolors.OKBLUE, bcolors.ENDC, msg), file = sys.stdout)
+        # Why does this code differ from eg. "error()"
```
buhtzz, line 65: In `error()` an error message is written to syslog no matter if debug mode is on or not. But `if DEBUG` then debug infos are **added** to the error messgae.

In `debug()` there is no need to **add** debug infos to the message because the message itself is of type `DEBUG`.

common/logger.py
```diff
@@ -55,14 +55,16 @@ def warning(msg , parent = None, traceDepth = 0):
 def info(msg , parent = None, traceDepth = 0):
     if DEBUG:
         msg = '%s %s' %(_debugHeader(parent, traceDepth), msg)
-    print('%sINFO%s: %s' %(bcolors.OKGREEN, bcolors.ENDC, msg), file=sys.stdout)
+    print('%sINFO%s: %s' %(bcolors.OKGREEN, bcolors.ENDC, msg), file=sys.stderr)
     for line in tools.wrapLine(msg):
         syslog.syslog(syslog.LOG_INFO, 'INFO: ' + line)
 
 def debug(msg, parent = None, traceDepth = 0):
     if DEBUG:
```
buhtzz, line 63: I would revert that if.

```
def debug()
    if not DEBUG:
        return
```

In that case you can save the indention of the following lines.
Clearer code IMHO.

aryoda: I think this would change the existing logging logic (different output) so I refrain from that.

common/logger.py
aryoda: What is irritating to me is that DEBUG is **not** a severity threshold for logging but a flag to inject additional call stack information ("_debugHeader") but for a single severity level `debug()` it is also a threshold.
I do not know any logging framework that works like that (and don't understand why this makes sense).
Perhaps the idea was to work like `-v` vs. `-vv` verbosity switches (the more "v"s the chattier is the output ;-)

buhtzz: Maybe we misunderstand here each other? e.g. in Python's own `logging` package `logging.DEBUG` is a threshold.
But you are deeper in that part of the code. It was just a quick and dirty night shift thought. :smile:

Issue #1338, opened 2022-10-20
[Ubuntu] Key is stored in legacy trusted.gpg keyring (apt warning)


gander:
On Ubuntu, I am getting such an error from APT

> W: http://ppa.launchpad.net/bit-team/stable/ubuntu/dists/jammy/InRelease: Key is stored in legacy trusted.gpg keyring (/etc/apt/trusted.gpg), see the DEPRECATION section in apt-key(8) for details.

freiheitsnetz:
I agree with @aryoda : This sounds exactly like the changes which where done in apt and is not a problem of Back In Time.
There are many threads on askubuntu on this issue and you can follow the guidelines given in https://askubuntu.com/questions/1398344/apt-key-deprecation-warning-when-updating-system to fix it.

gander:
I was sure it was something with this repository, because I installed this program on a fresh installation of Linux Mint 21 (based on Ubuntu 22.04), and only in this repository I have this deprecation

gander:
I corrected these settings in file `bit-team-stable-jammy.list` as it was in the article.

Before:
```
deb http://ppa.launchpad.net/bit-team/stable/ubuntu jammy main
```
After:
```
deb [arch=amd64 signed-by=/usr/share/keyrings/launchpad.gpg] http://ppa.launchpad.net/bit-team/stable/ubuntu jammy main
```

---

However, this does not fix the problem as this file was created during the installation of the application.

gander:
I have already fixed this problem by following [these instructions](https://askubuntu.com/a/1398346/678427). I am no longer seeing these warnings.

As already mentioned, I use Linux Mint, and here I have a Software Manager (mintinstall) with the package `backintime-qt` available.

```
Package: backintime-qt                   
Version: 1.3.2~jammy
State: installed
Automatically installed: no
Priority: extra
Section: utils
Maintainer: BIT Team <bit-team@lists.launchpad.net>
Architecture: all
Uncompressed Size: 418 k
Depends: x11-utils, libnotify-bin, python3-pyqt5, python3-dbus.mainloop.pyqt5, policykit-1, backintime-common (>= 1.3.2~jammy~), python3:any (>= 3.3~)
Recommends: python3-secretstorage
Suggests: meld | kompare
Conflicts: backintime-gnome (< 1.3.2~jammy~), backintime-kde (< 1.3.2~jammy~), backintime-kde4 (< 1.3.2~jammy~), backintime-notify (< 1.3.2~jammy~),
[...]

PR #1344, opened 2022-10-24
Add "warning: apt-key is deprecated" to known issues (#1338)

aryoda:


buhtzz (approved):

emtiu:
Thanks for this! Let's leave #1338 open until we know what to recommend to users instead of the current procedure ;)

PR #1347, opened 2022-10-30
Fix root launcher

phparkle:
Back In Time (root) fails to launch on Ubuntu Desktop 22.04.

Fix is borrowed from https://github.com/linuxmint/timeshift/blob/master/src/timeshift-launcher

### DIFF
--- a/qt/backintime-qt_polkit
+++ b/qt/backintime-qt_polkit
@@ -1,9 +1,30 @@
-#!/bin/sh
+#!/bin/bash
 
-if [ "x$XDG_SESSION_TYPE" = "xwayland" ]; then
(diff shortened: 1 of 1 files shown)


buhtzz:
Dear @phparkle ,
thanks for your PR.

Can you please link to the Issue or describe the problem that this PR tries to solve. I'm not aware of a problem with "starting BIT (root) on Ubuntu 22".

phparkle:
Hi @buhtzz, thanks for looking into this.

#1348 describes the issue exactly.

I wasn't aware of the ubuntu ticket before but I believe it is referring to the same issue.

I don't know how this will impact other distros. I just copied the way timeshift launches a root GUI and it happened to work, but I am not sure if it is a proper fix.

Issue #1320, opened 2022-10-05
Cannot delete snapshots when using --old-args with backintime


petaramesh:
Hello,

I use backintime-qt 1.3.2 with rsync v3.2.6.

At first it wouldn't work until I found in another post the suggestion to add “--old-args” in the advanced rsync options, that fixed it.

However, I now can create snapshots, but I cannot delete snapshots, and running backintime in debug mode when I try to delete snapshots, shows me errors similar to what I saw before using the “--old-args”.

So it appears to me that the “--old-args” additional options is not passed when deleting snapshots.

I assume this is a bug that would need fixing ?

Kind regards.

buhtzz:
Dear @petaramesh ,
thank you very much for this report.

I also assume that you are right and this is related to #1247 which is about the _new default argument protection_ of rsync. I'm still working on that problem and will also take your side problem into account.

I just have a solution but it will take some time to implement this without blowing everything away. The new maintaining team is just forming and we are getting just warm with the code of BIT. But in context of #1247 your problem has the highest priority.

Tag: Bug, High, Extern/Rsync

petaramesh:
Hi,
I'm not sure where is the best place to set these rsync options (given that I wouldn't want them to affect my usual rsync commands).

I tried to put “Exec=RSYNC_OLD_ARGS=1 RSYNC_PROTECT_ARGS=0 backintime-qt” in /usr/share/applications/backintime-qt.desktop, but it doesn't seem to help. Or should I rather use some “export” command ?

buhtzz:
> I wouldn't want them to affect my usual rsync commands).

Very important point. I should have mentioned that myself. Sorry.

> I tried to put “Exec=RSYNC_OLD_ARGS=1 RSYNC_PROTECT_ARGS=0 backintime-qt” in /usr/share/applications/backintime-qt.desktop, but it doesn't seem to help. Or should I rather use some “export” command ?

I'm not well with desktop-files. But on a usual shell you would do it via `export RSYNC_OLD_ARGS=1`.

Another place where you could do this is the "entry point" (a start script).
https://github.com/bit-team/backintime/blob/master/qt/backintime-qt

You my find it at `/usr/bin/backintime-qt` on your system.

buhtzz:
If it is possible for you and your environment, can you please check out the PR #1351 and test if your problem is fixed?

emtiu:
Closing this Issue, because #1351 was merged and should have eliminated the problem. Please re-open if it causes trouble again!

Issue #1248, opened 2022-04-18
Will EncFs be replaced?


1fexd:
Hi,

will EncFs ever be replaced with a more secure alternative in the future, or should I switch to another solution for encrypted remote backups?

mossroy:
There had been some discussion on replacing encfs by [gocryptfs](https://github.com/rfjakob/gocryptfs): https://github.com/bit-team/backintime/issues/644, and some initial work in branch gocrypt: https://github.com/bit-team/backintime/tree/gocrypt

But it seems stalled (no change in the last 4 years).

Encfs probably has flaws, but is still better than nothing. If it's not enough for you, consider switching to another solution

Codeberg-AsGithubAlternative-buhtz:
Would one of you please describe a use-case for EncFS/GoCryptFS?

I am not aware of it and don'T understand where it belongs in the context of BackInTime.

When it comes to encrypted file system I would assume that this is the responsibility of the operating system and not BIT. But it seems I'm wrong here. So please teach me. ;)

mossroy:
It's usually interesting when you put your backup in a remote location that you cannot fully trust.
An administrator of the remote server could read your backup for example (even with an encrypted filesystem)
By locally encrypting your data before sending it, you can safely put it in this remote location.

In our case, EncFS is known to probably have flaws, so the "safely" is not really true (even if it's still better than putting the data in cleartext)

Codeberg-AsGithubAlternative-buhtz:
OK I understand.

But how does this integrate in BIT?

From my understanding you "mount" (e.g. to `/mnt/mysecrets`) the encrypted FS on a operating system level and then start BIT using that mount point as a destination.

I don't see why BIT use EncFs "inside" and depends on that.

Taking a snapshot involves a lot of steps for BIT. On which of that steps is EncFs involved?

Maybe I am wrong: But in my experience BIT doesn't mount something for you. You have to use the `user-callbacks` to handle that.

mossroy:
When you configure an encryption password in BiT, it does all the necessary for you: it mounts a temporary location using encFS, that relies on a local or sshfs location.

Codeberg-AsGithubAlternative-buhtz:
This here?
https://backintime.readthedocs.io/en/latest/settings.html#local-encrypted

Please let me ask:
Does BiT really need that feature? There are a lot of solutions out there creating encrypted FS. Why not let them (and the operating system) take care of that?

I don't mean this feature is useless, stupid or luxury!
[...]

Issue #1121, opened 2020-12-04
Blocking error not shown in GUI


mat-m:
Hello

I had an issue where snapshots were not taken. Using the GUI, I had no information and no log was generated.

I decided to run in CLI and got:
```
DEBUG: [common/snapshots.py:1751 Snapshots.flockExclusive] Set flock /tmp/backintime.lock
Traceback (most recent call last):
  File "/usr/share/backintime/common/backintime.py", line 1165, in <module>
    startApp()
  File "/usr/share/backintime/common/backintime.py", line 517, in startApp
    args.func(args)
  File "/usr/share/backintime/common/backintime.py", line 739, in backup
    ret = takeSnapshot(cfg, force)
  File "/usr/share/backintime/common/backintime.py", line 94, in takeSnapshot
    ret = snapshots.Snapshots(cfg).backup(force)
  File "/usr/share/backintime/common/snapshots.py", line 633, in backup
    self.flockExclusive()
  File "/usr/share/backintime/common/snapshots.py", line 1752, in flockExclusive
    self.flock = open(self.GLOBAL_FLOCK, 'w')
PermissionError: [Errno 13] Permission denied: '/tmp/backintime.lock'
 ```

Is there a way to push these errors in the GUI, or at least ask to run the CLI to check ?

Codeberg-AsGithubAlternative-buhtz:
Maybe related to #1122 

How can you reproduce this behaviour/error?

mat-m:
On the same host, I defined two users lauching a different backup at the same time. In my case, it was `@reboot`, but I think it would trigger it also when setting the same custom hours, or even with a simultaneous command line

buhtzz:
@mat-m There is a new forming maintaining team and we do review all issues. Is this problem still relevant for you or did you find a solution?

Maybe a duplicate or a neighbor problem.

Tag: Bug, Medium, Scheduling, Feedback

mat-m:
@buhtzz Glad a new team is coming ! Thanks so much for this !
I worked it around by using different times. But the lock name is too generic to handle mutli-users  or multi-targets. I actually don't know what you want to lock. In my opinion, the lock should be on the target dir

mat-m:
Can be considered as dup of #1122, where I will repost my last message.

mat-m:
Ah, no.  My point here is that the error was not pushed to the GUI. I had to run BIT in console to see it and understand why it blocked trough the GUI

emtiu:
[...]

Issue #907, opened 2018-07-07
Very hidden error message


Azimoth:
Hi, 
I'm running Backintime 1.2.0_pre20180221 on Gentoo. And had the following situation:

Backintime was running fine and apparently suddenly stopped working. Clicking the icon in Gnome lead to a loading circle, but it vanished without any window opening and running "backintime" on the CLI only printed version and licencse information, while backintime-qt addtionally said "The application is already running! (pid: 21895)".

Only when I killed the running backintime processes and ran backintime-qt in the CLI once again, I was prompted 

> The authenticity of host '192.168.x.x (192.168.x.x)' can't be established.
> ECDSA key fingerprint is SHA256:<randomgibberish>.
> Are you sure you want to continue connecting (yes/no)? ``´´´´

and realized I accidentally had removed this address from known hosts.

I suggest to handle this problem in a more visibile error, preferably in a GUI window.

emtiu:
Related Issues or possible duplicates:
* #84 
* #1121

PR #1351, opened 2022-11-04
Fix: Incompatibility with rsync 3.2.4 or newer

### LINKED ISSUE #1247: Incompatibility with rsync 3.2.4

buhtz:
# Summary
This fix #1247 (and kind of #1320). The core problem is that BIT do call `rsync` in a way that is not compatible with _newer_ versions of `rsync`. The version `3.2.4` of rsync [activated a new method of _argument protection_](https://download.samba.org/pub/rsync/NEWS#BEHAVIOR_CHANGES-3.2.4). This PR does fix the `rsync` calls and did all necessary modifications. **Important** to know is that BIT still work with _old_ rsync versions, too.

From my point of view that solution is final/stable. I introduced two new unit tests in former PR's (#1340 #1334). I did systematic manual tests with different GNU/Linux distributions including different `rsync` versions. I used all four types of snapshot profiles (local, SSH, local encrypted, SSH encrypted) in that test. I used `git grep` to find possibly relevant parts of the code. All that steps brought up extra issues/problems I deald with. **But** theoretically there still is a risk of side effects.

**Because of that please carefully review that PR and also do multiple and pedantic tests yourself.**

## Short description of the solution
In the past BIT did call `rsync` (simplified) like this:
```
rsync user@localhost:"/home/user/folder with blank in name"
```
Now it is done without the quoting of the path via `"` and with the new flag `-s`:
```
rsync -s user@localhost:/home/user/folder with blank in name
```
# The solution in detail
The `rsync` [maintainer has checked and confirmed my approach](https://lists.samba.org/archive/rsync/2022-September/032882.html).

The goals are
 - Not using workarounds like `--old-args` flag or environment variables like `RSYNC_OLD_ARGS` or `RSYNC_PROTECT_ARGS` to force newer `rsync` versions to use the _old way_ of argument protection.
 - Use only one way to call `rsync` and don't use `if` statements and/or check the version of the current installed `rsync`.
 - The solution should work in the future, too.

## About the quoting
In most cases quoting a path in BIT is done by the method [`Snapshots.rsyncRemotePath()`](https://github.com/bit-team/backintime/blob/49db33e62eaa8501120d85ffcaa6443492df7422/common/snapshots.py#L1730). It has an argument `quote` which is by default `quote='"'`.

But I didn't modify that default value because of possible side effects.
I modified the calls of `rsyncRemotePath()` to explicit set that `quote` argument to an empty string (`quote=''`).

After doing some tests and `git grep` searches I can say that the productive BIT code never call that method with its `quote` default value. That is how it should be. There are only some unittest doing it. And I wouldn't touch that tests yet.

# Background
## Rsync
The _new argument protection_ was made the [default behaviour](https://download.samba.org/pub/rsync/NEWS#BEHAVIOR_CHANGES-3.2.4) with `rsync` version 3.2.4 but it was [introduced](https://download.samba.org/pub/rsync/NEWS#ENHANCEMENTS-3.0.0) (but optional) with version 3.0.0 in the year 2008.

The flag `-s` also have the synonym flags `--protect-args` and `--secluded-args` ([read further](https://download.samba.org/pub/rsync/NEWS#ENHANCEMENTS-3.2.6)). The latter two are not consistent exist in all versions. Only `-s` exists in all versions (since 3.0.0). That is why that short form is used in that approach.
 
Further reading about argument protection: [ADVANCED USAGE section in rsync manpage](https://download.samba.org/pub/rsync/rsync.1#ADVANCED_USAGE).

# Interactions with workarounds
Because of #1247 we advised the users a workaround. They should use  `--old-args` flag (in BIT Expert options) or the environment variables like `RSYNC_OLD_ARGS` or `RSYNC_PROTECT_ARGS` to force newer `rsync` versions to use the _old way_ of argument protection. Do this workarounds conflicts with that PR?

I checked it out. The environment variables having a lower priority then flags on commandline. It means `-s` will _overwrite_ that environment variables. No problem here.

The `--old-args` combined with `-s` will cause an "conflict error" by rsync. Because of that this PR does remove `--old-args` automatically from the `rsyncOptions` config variable and give a WARNING message about it on the terminal (see `config.py::Config.rsyncOptions()`). IMHO it is OK not to inform the users via GUI in a more offensive way. It only affects a small number of users and won't hurt them or their data if we do it that way.
# About `Snapshot.backupConfig()` and `tools.Execute()`
It is a general problem of BIT that `tools.Execute()` do only show a WARNING when the command had a problem. When using an encrypted SSH snapshot profile that `rsync` call there faild because of #1247 and no one noticed it. It even took me multiple manual tests in `--debug` mode to recognize that WARNING line in between the hundreds of other debug output lines.

I canceled my first intention to create a unit test for that case because the encryption (EncFS) need be replaced or rewritten in the near future. I choose a more simple approach and simply "transformed" the WARNING (only in that specific situation) into an (red) ERROR. But with this PR this ERROR shouldn't come up anymore.
# Testing
## TravisCI
I tested that PR in my own forked repo first. Travis gave me all lights green there. Here on bit-team the Travis account is currently out of credits and won't run.
## Manual tests
 - tested all four types of snapshot profiles:
    - local
    - local encrypted
    - SSH
    - SSH encrypted
 - create snapshots (100 snapshots, one on each day; using `faketimelib`)
 - manually delete snapshots
 - smart (auto) delete snapshots
 - restore snapshots

Distros
 - Debian 11 stable (using the "old" rsync)
 - Debian (12) testing (using the "new" rsync)
 - Manjaro (Arch based)
 - Ubuntu 22
 - a special Debian 11 (on ARM)
 - via Travis: Ubuntu 20

Any suggestions about other important distros that are not based on Debian or Arch?


emtiu:
Congratulations! Our first big bugfix – thanks, @buhtzz :1st_place_medal:

> It is a general problem of BIT that `tools.Execute()` do only show a WARNING when the command had a problem. When using an encrypted SSH snapshot profile that `rsync` call there faild because of #1247 and no one noticed it. It even took me multiple manual tests in `--debug` mode to recognize that WARNING line in between the hundreds of other debug output lines.

Out of interest: Would you say that this is same problem described in #84 and #907 and #1121? If yes, then my understanding is: Your solution is a small improvement, because the error becomes more visible in the console/logs. But it's not yet a fix, the error is still invisible from the GUI/desktop. Correct?

> I canceled my first intention to create a unit test for that case because the encryption (EncFS) need be replaced or rewritten in the near future. I choose a more simple approach and simply "transformed" the WARNING (only in that specific situation) into an (red) ERROR. But with this PR this ERROR shouldn't come up anymore.

Just my personal opinion on this (main discussion in #1248): There will be no replacement for encfs in the near future. Yes, it's known to be insecure under certain circumstances, but it's very well-tested and robust. The only possible replacement seems to be gocryptfs, and it would be a lot of work to integrate.

buhtzz:
> Out of interest: Would you say that this is same problem described in #84 and #907 and #1121?

Technically I would say "no". The problem is that `tools.Excecute()` do run shell commands in a generalized way but did not take care if they fail or not. To be really sure we could implement "the hammer" and add a dirty and hard `raise RuntimeException` in there when the return code is `!0`. But of course this is a bit to much. But it is maybe an option in debug situations.

Conceptual "yes". It is part of the problem that the core of BIT doesn't communicate with its other parts e.g. the GUI. We could open a meta issue about it. But I suspect it is also "to meta" for an issue and should better be discussed on [bit-dev](https://mail.python.org/mailman3/lists/bit-dev.python.org/). IMHO it is part of a re-design and not just adding a new "communication system" (e.g. DBUS or other IPC solutions). BIT just shouldn't call itself via `subprocess.Popen()`. There are better ways.

> my understanding is: Your solution is a small improvement, because the error becomes more visible in the console/logs. But it's not yet a fix, the error is still invisible from the GUI/desktop. Correct?

Correct. Not a fix. Just a workaround and only relevant in that very special use case where an "SSH encrypted" snapshot profile is used.

buhtzz:
Sorry I couldn't resists. Just read [that heise-Article](https://heise.de/-7330320) about a GPT-3 based algorithm generate summary text from code diffs for PRs. I see no value in such a tool even if it would work 100% perfect. I just played a bit arround.

emtiu:
> Sorry I couldn't resists. Just read [that heise-Article](https://heise.de/-7330320) about a GPT-3 based algorithm generate summary text from code diffs for PRs. I see no value in such a tool even if it would work 100% perfect. I just played a bit arround.

I was honestly surprised at the quality of the result :grinning:

aryoda (commented):
I have reviewed the code and found only one location that possibly could have been forgotten to change:

https://github.com/bit-team/backintime/blob/ac6933fcb9a715524bed1d820f63bf0b5fc2dc3d/common/snapshots.py#L1473

@buhtz Could you please check this LOC which still contains an oddly escaped quote...? THX!

buhtz:
> 1. `-s` is supported since 2008 (v3.0.0) in rsync with the same semantics
> 2. `rsync -s user@localhost:/home/user/folder with blank in name` (with `-s` but without surrounding quotes) is the "new" way of doing it in BiT?

Yes, it is just that simple.

But to more precise with point 2. In BIT it is done that way

`subprocess.Popen(['rsync', '-s', 'user@localhost:/home/user/folder with blank in name'])`

So `Popen()` itself will surround the command elements with quotes if it sees the need (e.g. blanks) for it.

> Which scenarios do you consider as the main risks to be tested by us?

Good question. I don't have a special scenario in mind. I would say that scenarios that seems to be _not special_. ;)
Technically most of the magic happens with SSH profiles. But I would always say to test nearly everything because I'm not enough into all details and features of BIT to judge that point.

> I have reviewed the code and found only one location that possibly could have been forgotten to change:
> https://github.com/bit-team/backintime/blob/ac6933fcb9a715524bed1d820f63bf0b5fc2dc3d/common/snapshots.py#L1473
> @buhtz Could you please check this LOC which still contains an oddly escaped quote...? THX!

I have an idea what's going on there and why it works. Removing a snapshot is done on the local filesystem because the remote is mounted. But I will take a closer look to it.

buhtz:
> https://github.com/bit-team/backintime/blob/ac6933fcb9a715524bed1d820f63bf0b5fc2dc3d/common/snapshots.py#L1473
> 
> @buhtz Could you please check this LOC which still contains an oddly escaped quote...? THX!

OK, I got it. When you look into the resulting rsync-path-string (when using _smartRemove in background_ in an SSH snapshot profile) you can see it is not quoted.

This is the call
```
remote = self.rsyncRemotePath(sid.path(use_mode = ['ssh', 'ssh_encfs']), use_mode = [], quote = '\\\"')
````

Looking into 
```
    def rsyncRemotePath(self, path, use_mode = ['ssh', 'ssh_encfs'], quote = '"'):
        mode = self.config.snapshotsMode()

        if mode in ['ssh', 'ssh_encfs'] and mode in use_mode:
            user = self.config.sshUser()
            host = tools.escapeIPv6Address(self.config.sshHost())
            return '%(u)s@%(h)s:%(q)s%(p)s%(q)s' %{'u': user,
                                                   'h': host,
                                                   'q': quote,
                                                   'p': path}
        else:
            return path
```

`mode is "ssh"` but `use_mode is []`. Because of that the `if`-block is never entered and the `quote` argument is ignored.
I don't really understand why it is intended that way but I would work on that in separate in #1359.

So I would say the PR is "clean" again and don't have any unclear problems.
But I would keep it open for some more weeks. It doesn't hurry. I want to make sure.
I will refresh my request for testing to the other reports and on the ML in some days.

aryoda:
I'd also say "go" for a merge (the earlier we use it the earlier we may find hidden issues).

I want to set up kinda integration test for myself (taking backups, restoring them and do a folder compare) but this will take a few more weeks (not much time ATM).

Any vetos?

emtiu:
No veto. I haven't found time to test this myself, but I trust the procedures @buhtz described enough to merge :)

Germar:
Good job Christian! Thank you!
I'd suggest to add a deprecation warning in `Snapshots.rsyncRemotePath` if `quote` is not empty. This way we would find missed calls for this easily.

PR #1352, opened 2022-11-04
Fix: Invalid escape sequences

buhtzz:
This fix `DeprecationWarnings` about _invalid escape sequences_. Most of that warnings are caught when running `pytest`. That warnings will become serious errors in future Python versions.

This where the warnings. Sorry for providing a screenshot instead of plain text but I still have problems copy & past from a Windows terminal window running ssh client and tmux.
image

Of course, I checked all problems and what that strings really need to do.
Most of the warnings are about `\$` which I modified to `\\$`. It is like escaping and escape. Sounds wired but it is intended to be that way. Because that string is handed over (via `subprocess.Popen()`) to a shell command (ssh) which need to receive a string containing an escape.

### DIFF
--- a/common/snapshots.py
+++ b/common/snapshots.py
@@ -423,7 +423,7 @@ class Snapshots:
                                         restore to original destiantion
             delete (bool):              delete newer files which are not in the
                                         snapshot
-            backup (bool):              create backup files (\*.backup.YYYYMMDD)
+            backup (bool):              create backup files (*.backup.YYYYMMDD)
                                         before changing or deleting local files.
             only_new (bool):            Only restore files which does not exist
                                         or are newer than those in destination.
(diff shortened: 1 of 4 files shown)


common/snapshots.py
```diff
@@ -1405,22 +1405,25 @@ def smartRemove(self, del_snapshots, log = None):
             additionalChars = len(self.config.sshPrefixCmd(cmd_type = str))
 
             head = 'screen -d -m bash -c "('
-            head += 'TMP=\$(mktemp -d); '                      #create temp dir used for delete files with rsync
-            head += 'test -z \\\"\$TMP\\\" && exit 1; '        #make sure $TMP dir was created
-            head += 'test -n \\\"\$(ls \$TMP)\\\" && exit 1; ' #make sure $TMP is empty
+            # create temp dir used for delete files with rsync
```
buhtzz, line 1408: Here it is hard to see. In all three lines the `\$` was modified to `\\$`.

common/snapshots.py
```diff
@@ -1405,22 +1405,25 @@ def smartRemove(self, del_snapshots, log = None):
...
+            head += 'TMP=\\$(mktemp -d); '
+            # make sure $TMP dir was created
+            head += 'test -z \\\"\\$TMP\\\" && exit 1; '
+            # make sure $TMP is empty
+            head += 'test -n \\\"\\$(ls \\$TMP)\\\" && exit 1; '
```
buhtzz, line 1413: Sorry it is hard to see here.
In all three lines the `\$` was modified to `\\$`.

common/snapshots.py
```diff
@@ -1634,7 +1637,7 @@ def statFreeSpaceSsh(self):
         # /tmp           127266564 115596412   5182296  96% /
         #                                      ^^^^^^^
         for line in output.split(b'\n'):
-            m = re.match(b'^.*?\s+\d+\s+\d+\s+(\d+)\s+\d+%', line, re.M)
+            m = re.match(r'^.*?\s+\d+\s+\d+\s+(\d+)\s+\d+%', line.decode(), re.M)
```
buhtzz, line 1640: The output from shell was catched as binary instead of a string. No need for that.
Converted that to a real string (via ` .decode()`) and run a raw regex string (via `r''`) on it.

buhtzz:
Sorry I couldn't resists. Just read [that heise-Article](https://heise.de/-7330320) about a GPT-3 based algorithm generate summary text from code diffs for PRs. I see no value in such a tool even if it would work 100% perfect. I just played a bit arround.

emtiu (approved):

aryoda:
@buhtzz If you want you could also fix a few more "invalid escape sequence" warning candidates in:

- settingsdialog.py: 855 and 861
- create-manpage-backintime-config.py#166

buhtzz:
> @buhtzz If you want you could also fix a few more "invalid escape sequence" warning candidates

Good idea. In a short test I can find much more via `flake8`. I will cover this in the next PR.

aryoda:
> > @buhtzz If you want you could also fix a few more "invalid escape sequence" warning candidates
> 
> Good idea. In a short test I can find much more via `flake8`. I will cover this in the next PR.

That's fine, no urgent need so far.

PR #1353, opened 2022-11-04
Fix #1333: arch Linux build fails

### LINKED ISSUE #1333: [Arch] Backintime fails to build

aryoda:
I have fixed all known issues on our side an try to contact the package maintainer now asking
- to fix the missing dependencies
- to reconsider setting LC_ALL=C for `make`ing

On my arch VM the build & unit tests are working now (after manually installing the required dependencies)...

Fixes #1333

### DIFF
--- a/common/diagnostics.py
+++ b/common/diagnostics.py
@@ -138,7 +138,7 @@ def collect_diagnostics():
     # rsync <= (somewhere near) 3.1.3: -VV doesn't exists
     # rsync == 3.1.3 (Ubuntu 20 LTS) doesn't even know '-V'
 
-    # This work when rsync understand -VV and return json or human readable
+    # This works when rsync understands -VV and returns json or human readable
     result['external-programs']['rsync'] = _get_extern_versions(
         ['rsync', '-VV'],
         r'rsync  version (.*)  protocol version',
@@ -189,15 +189,19 @@ def _get_extern_versions(cmd,
                          error_pattern=None):
     """Get the version of an external tools using ``subprocess.Popen()``.
 
-    Args:
-        cmd (list): Commandline arguments that will be passed to `Popen()`.
-        pattern (str): A regex pattern to extract the version string from the
-                       commands output.
-        try_json (bool): Interpet the output as json first (default: False).
-        error_pattern (str): Regex pattern to identify a message in the output
-                             that indicates an error.
+     Args:
+         cmd (list): Commandline arguments that will be passed to `Popen()`.
+         pattern (str): A regex pattern to extract the version string from the
+                        commands output.
+         try_json (bool): Interpret the output as json first (default: False).
+                          If it could be parsed the result is a dict
+         error_pattern (str): Regex pattern to identify a message in the output
+                              that indicates an error.
 
-    """
+     Returns:
+         Version information as string (or dict if was JSON and parsed successfully).
+         `None` if the error_pattern did match (to indicate an error)
+     """
 
     try:
         # as context manager to prevent ResourceWarning's
@@ -219,7 +223,7 @@ def _get_extern_versions(cmd,
             if match:
                 return None
 
-        # some tools use "stderr" for version infos
+        # some tools use "stderr" for version info
         if not std_output:
             result = error_output
         else:
@@ -236,14 +240,13 @@ def _get_extern_versions(cmd,
                 pass
 
             else:
-                # No regex parsing because it was json
-                pattern = None
+                return result  # as JSON
 
         # extract version string
         if pattern:
             result = re.findall(pattern, result)[0]
 
-    return result.strip()
+    return result.strip()  # as string
 
 
 def get_git_repository_info(path=None):
--- a/common/diagnostics.py
+++ b/common/diagnostics.py
@@ -77,7 +77,14 @@ def collect_diagnostics():
         'XDG_SESSION_TYPE', '($XDG_SESSION_TYPE not set)')
 
     # locale (system language etc)
-    result['host-setup']['locale'] = ', '.join(locale.getlocale())
+    #
+    # Implementation note: With env var "LC_ALL=C" getlocale() will return (None, None).
+    # This throws an error in "join()":
+    #   TypeError: sequence item 0: expected str instance, NoneType found
+    my_locale = locale.getlocale()
+    if all(x is None for x in my_locale):
+        my_locale = ["(Unknown)"]
+    result['host-setup']['locale'] = ', '.join(my_locale)
 
     # PATH environment variable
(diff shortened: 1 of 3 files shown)


buhtzz:
I cloned directly from "aryoda:issue/1333_Arch_build_fails".

Testing on Debian 11 (stable) via `./configure && make && make install && make test` runs fine (even in an SSH terminal without X11 or Wayland involved).

I also used my Manjaro VM for testing. When I'm on a temrinal emulator in the desktop environment `make test` is fine.
But when I log into that same machine via SSH in a real terminal (without X11/Wayland involved) I got an error about `QPainter` which polutes the output and make a test fail.

Why is Qt involved here? We don't have any tests for Qt. Where is it imported?

```
user@Alf ~/sfd/backintime/common $ make test
python3 -m unittest  -b test/test_applicationinstance.py
...................
Ran 19 tests in 0.048s

OK
python3 -m unittest  -b test/test_argparser.py
.....................
Ran 21 tests in 3.355s

OK
python3 -m unittest  -b test/test_backintime.py
WARNING: Failed to connect to Udev serviceHelper daemon via D-Bus: org.freedesktop.DBus.Error.Spawn.ChildExited
WARNING: D-Bus message: Launch helper exited with unknown return code 1
WARNING: Udev-based profiles cannot be changed or checked due to Udev serviceHelper connection failure
.Fs
======================================================================
FAIL: test_local_snapshot_is_successful (test.test_backintime.TestBackInTime)
end to end test - from BIT initialization through snapshot
  File "/media/sf_Downloads/backintime/common/test/test_backintime.py", line 146, in test_local_snapshot_is_successful
    self.assertRegex(filtered_log_output, re.compile(r'''INFO: Lock
AssertionError: Regex didn't match: 'INFO: Lock\nINFO: Take a new snapshot. Profile: 1 Main profile\nINFO: Call rsync to take the snapshot\nINFO: Save config file\nINFO: Save permissions\nINFO: Create info file\nINFO: Unlock\n' not found in 'INFO: Lock\nINFO: Take a new snapshot. Profile: 1 Main profile\nINFO: Call rsync to take the snapshot\nINFO: Save config file\nINFO: Save permissions\nINFO: Create info file\nQPainter::begin: Paint device returned engine == 0, type: 2\nQWidget::render: Cannot render with an inactive painter\nINFO: Unlock\n'

Ran 3 tests in 4.166s

FAILED (failures=1, skipped=1)
make: *** [Makefile:570: unittest] Fehler 1
```

aryoda:
Very good test, this looks like a new bug.

> But when I log into that same machine via SSH in a real terminal (without X11/Wayland involved)
> I got an error about QPainter which polutes the output and make a test fail.

Did I introduce this with my PR or does it happen with `master` too?

I can't see a line number for this QPainter message so I need a way to reproduce this.
Do you have a minimal setup to reproduce this (eg. without `ssh`?).
Meanwhile I will try to boot Linux into a console instead of a desktop environment, perhaps this is showing the message too...

> I was wondering why Qt is involved here. But I saw that Qt is imported in `diagnostics.py`

Yes, diagnostics does show the PytQt5 version number, but only when "--diagnostics" is called.
And this is the most plausible reason since some unit tests do use "--diagnostics".

https://github.com/bit-team/backintime/blob/2121c60c882e8fa16de6439a27e91b046118291d/common/diagnostics.py#L121-L130

Perhaps it would be better to use PyQt5 only if `--diagnostics` was started from `backintime-qt`
(I saw a "flag" to recognize this in the code - I am searching...)

Could you please do a quick check via the console that printed the "QPainter" message:

`python3 -c "import PyQt5.QtCore; print(PyQt5.QtCore.PYQT_VERSION_STR)"`

If this also writes the QPainter message we would have found the culprit.

aryoda:
When I boot into a Manjaro console via run-level 3 in grub the same unit test fails but with another actual/expected difference:
One line too much in the output:

> WARNING: Inhibit Suspend failed.

If I ignore this warning by filtering it like the DBus warnings (test_backintime staring line 147) the unit tests succeed.

I could not yet reproduce the `QPainter` message (I am seaching the web for the same message ATM)...

buhtzz:
I'm still confused and not sure if "my problem" is relevant here or should block the merge. I tend to merge that PR and see what happens in the wildlife out there. :tiger2: 

> Could you please do a quick check via the console that printed the "QPainter" message:
> `python3 -c "import PyQt5.QtCore; print(PyQt5.QtCore.PYQT_VERSION_STR)"`

No errors with that. I just get the version number back.

The error happens in a unittest where `--diagnostics` isn't called. That confuse me, too.

# How I reproduce
The `make test` isn't needed. I can also reproduce it via `pytest`.

`pytest test/test_backintime.py::TestBackInTime::test_local_snapshot_is_successful`

It is Manjaro on a `tty` terminal, on a `ssh` terminal and on a `tmux` terminal. The problem doesn't happen when using a terminal emulator in a desktop environment.

buhtzz:
Btw: It seems that you edited my own comment instead of quote and answer it. ;)

> Do you get the `QPainter` message also when running BiT `master` instead of my PR version?

Yes I can reproduce this with `master` also. This indicates to me that it isn't a serious problem because it was there before but no one noticed or reported about it.

aryoda:
> Btw: It seems that you edited my own comment instead of quote and answer it. ;)

Damn it, I shouldn't do two things in parallel :-)

> > Do you get the `QPainter` message also when running BiT `master` instead of my PR version?

> Yes I can reproduce this with `master` also. This indicates to me that it isn't a serious problem
>  because it was there before but no one noticed or reported about it.

OK, then I suggest to merge now and I do the analysis until EoW.

buhtzz (approved):

Issue #1354, opened 2022-11-05
Decide about manpage "backintime-config" and the generating script "create-manpage-backintime-config.py"


buhtz:
# Current situation
 - BIT offers the [manpage `backintime-config`](https://manpages.ubuntu.com/manpages/bionic/man1/backintime-config.1.html) describing all(?) variables used in the config file.
 - There is the script [`common/create-manpage-backintime-config.py`](https://github.com/bit-team/backintime/blob/master/common/create-manpage-backintime-config.py) that intends to generate the manpage out of the real existing config file.
 - The script is nearly undocumented and full of bugs.
# Open questions
Because of the bugs I assume that that script wasn't used in the last years. Looking into the commit history of that file it seems to me that only some minor changes (copyright and version strings) where made.
# To discuss
I tend to remove that script and the `backintime-config`manpage. Users don't need to edit the config file by hand. The infos from that manpage are technical documentation and should go somewhere else. Or if the config.py code has good in-code docstrings there is no more need for an extern technical documentation about the config file. So my primary argument is that we (or the user) don't need a config manpage.

Fixing the script would take much resources. I suspect that writing it new from scratch would be faster. But even this does take resources that are invested better in other tasks.

aryoda:
Good point. I just realized my manpage fix

https://github.com/bit-team/backintime/commit/3c1e08e110246e2159f6666ad309664098cbb6eb

will most probably be overwritten if the py script is really used as man page generator.

> if the config.py code has good in-code docstrings there is no more need for an extern technical documentation about the config file.
> So my primary argument is that we (or the user) don't need a config manpage.

When I was fixing the man page due to an issue the OP was really using the doc to write his BiT config file "by hand".
This is a rare but existing use case and the documentation is IMHO still required.

As end user I'd prefer to **not** read the documentation as docstrings in a python file (which is BTW not installed and therefore
not accessible for end users!).

emtiu:
> When I was fixing the man page due to an issue the OP was really using the doc to write his BiT config file "by hand".
> This is a rare but existing use case and the documentation is IMHO still required.

I agree, this is a legitimate use-case. It seems to be very rare that backintime is used headless, but I like that it's possible, and we should continue supporting it.

> Good point. I just realized my manpage fix [3c1e08e](https://github.com/bit-team/backintime/commit/3c1e08e110246e2159f6666ad309664098cbb6eb) will most probably be overwritten if the py script is really used as man page generator.

[...]

PR #1355, opened 2022-11-05
Fix: Invalid escape sequence warnings and some PEP8

buhtzz:
- Eliminated possible deprecation warnings about invalid escape sequences ([W605](https://www.flake8rules.com/rules/W605.html))
- Eliminated round about 800 PEP8 problems in `settingsdialog.py`
- I didn't touched `create-manpage-backintime-config.py` because of #1354 

I was search for invalid escape sequences via `flake8`.

### DIFF
--- a/qt/settingsdialog.py
+++ b/qt/settingsdialog.py
@@ -1,5 +1,6 @@
 #    Back In Time
-#    Copyright (C) 2008-2022 Oprea Dan, Bart de Koning, Richard Bailey, Germar Reitze, Taylor Raack
+#    Copyright (C) 2008-2022 Oprea Dan, Bart de Koning, Richard Bailey,
+#                            Germar Reitze, Taylor Raack
 #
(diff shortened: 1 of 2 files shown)


qt/settingsdialog.py
```diff
@@ -23,9 +24,40 @@
 import grp
 import re
 
-from PyQt5.QtGui import *
-from PyQt5.QtWidgets import *
-from PyQt5.QtCore import *
+from PyQt5.QtGui import QIcon, QFont, QPalette, QBrush, QColor
```
aryoda, line 27: Import * was a bad practice and should be replaced be importing the module (see https://google.github.io/styleguide/pyguide.html#22-imports).

Since this would break a lot of code that uses the imported classes I consider this change as the best work-around for now until we do a full redesign.

aryoda (approved):
This PR does not change how the code works (except the more explicit `from ... import` statements) so I think this can be merged now.

Issue #1270, opened 2022-08-15
schedule.time should specify 'hour' but seems to specify 'minute'


Whytey:
Using backintime version 1.3.2 from the command line only.  No GUI installed.
```
root@server1:~/backintime# backintime --version
backintime 1.3.2
```

I have three profiles.  I intended to run profile 1 and 2 daily at 1am and 2am respectfully.  Profile 3 should be run every Sunday at 3am.

Config snippet:
```
profile1.name=filesystem
profile1.schedule.mode=20
profile1.schedule.time=1

profile2.name=databases
profile2.schedule.mode=20
profile2.schedule.time=2

profile3.name=containers
profile3.schedule.mode=30
profile3.schedule.time=03
profile3.schedule.weekday=7
```

When I run `check_config`, it goes through the process of creating the crontab entries, but seems to specify the times as 1, 2 and 3 minutes past midnight:

```
root@server1:~/backintime# crontab -l
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin

#Back In Time system entry, this will be edited by the gui:
1 0 * * * /usr/bin/nice -n19 /usr/bin/ionice -c2 -n7 /usr/bin/backintime --config /root/backintime/backintime.config backup-job >/dev/null 2>&1
#Back In Time system entry, this will be edited by the gui:
2 0 * * * /usr/bin/nice -n19 /usr/bin/ionice -c2 -n7 /usr/bin/backintime --profile-id 2 --config /root/backintime/backintime.config backup-job >/dev/null 2>&1
#Back In Time system entry, this will be edited by the gui:
3 0 * * 7 /usr/bin/nice -n19 /usr/bin/ionice -c2 -n7 /usr/bin/backintime --profile-id 3 --config /root/backintime/backintime.config backup-job >/dev/null 2>&1
```

Seems like a bug to me.

Cheers,
Whytey

aryoda:
Did you create the profile manually? And what does the BiT GUI show as schedule time for this config/profile?

IMHO the time is a number but **NOT** the hour but a decimal-codes system of the form:

    hhmm

So 1 o'clock should be 'profile2.schedule.time=200' (for hour = 2 and minutes = 00).

See the code for this: https://github.com/bit-team/backintime/blob/master/common/config.py#L1507

Please set the Hour of the schedule in the BiT GUI under "Settings" and check how the `schedule.time` is modified then
[...]

PR #1360, opened 2022-11-09
Code comments and PEP8 prettification.

buhtz:
While investiagting the manpage creation script some comments where added. Also some code formatting was done to improve readability and better fitting PEP8 rules.

Also repaired special #? comments accidentially removed in PR #1313 before.

Related to #1354

### DIFF
--- a/common/config.py
+++ b/common/config.py
@@ -1,5 +1,6 @@
 #    Back In Time
-#    Copyright (C) 2008-2022 Oprea Dan, Bart de Koning, Richard Bailey, Germar Reitze
+#    Copyright (C) 2008-2022 Oprea Dan, Bart de Koning, Richard Bailey,
+#    Germar Reitze
 #
 #    This program is free software; you can redistribute it and/or modify
 #    it under the terms of the GNU General Public License as published by
@@ -15,6 +16,18 @@
 #    with this program; if not, write to the Free Software Foundation, Inc.,
 #    51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 
+"""Configuration logic.
+
+That modul and its `Config` handling the appliation logic round about the
+configuration of Back In Time. The handling of the configuruation file itself
+is separated in the module :py:module:`configfile`.
+
+Development notes:
+    Some of the methods do have code comments starting with `#? ` instead of
+    `# `. That special comments are used as text in the manpage
+    `backintime-config`. The script `create-manpage-backintime-config.py` do
+    parse this module for that.
+"""
 
 import os
 import sys
@@ -445,13 +458,8 @@ class Config(configfile.ConfigFileWithProfiles):
         return True
 
     def snapshotsMode(self, profile_id=None):
-        """Use mode (or backend) for this snapshot.
-
-        Look at 'man backintime' section 'Modes'.
-
-        Returns:
-            str: Possible values are local|local_encfs|ssh|ssh_encfs.
-        """
+        #? Use mode (or backend) for this snapshot. Look at 'man backintime'
+        #? section 'Modes'.;local|local_encfs|ssh|ssh_encfs
         return self.profileStrValue('snapshots.mode', 'local', profile_id)
 
     def setSnapshotsMode(self, value, profile_id = None):
--- a/common/config.py
+++ b/common/config.py
@@ -18,15 +18,15 @@
 
 """Configuration logic.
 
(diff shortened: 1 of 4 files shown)


common/config.py
```diff
@@ -15,6 +16,18 @@
...
+That modul and its `Config` handling the appliation logic round about the
+configuration of Back In Time. The handling of the configuruation file itself
+is separated in the module :py:module:`configfile`.
+
+Development notes:
+    Some of the methods do have code comments starting with `#? ` instead of
+    `# `. That special comments are used as text in the manpage
+    `backintime-config`. The script `create-manpage-backintime-config.py` do
+    parse this module for that.
+"""
```
emtiu, line 30:
```suggestion
"""Configuration logic.

This module and its `Config` class contain the appliation logic handling the
configuration of Back In Time. The handling of the configuration file itself
is separated in the module :py:module:`configfile`.

Development notes:
    Some of the methods have code comments starting with `#? ` instead of
    `# `. These special comments are used to generate the manpage
    `backintime-config`. The script `create-manpage-backintime-config.py`
    parses this module for that.
"""
```

common/create-manpage-backintime-config.py
```diff
@@ -15,6 +16,37 @@
...
+    `commentline`.
+  - If a line starts with `#` it will be skipped.
+  - Next the script searches for lines which ``retur`` the config value (like
+    `snapshots.ssh.port`. There it will extract the
+    key/name (`snapshots.ssh.port`), the default value (`22`),
+    the instance (`Int`) and if it is a profile or general value.
+  - If the line contain a `List` value like `snapshots.include` it will
+    process all values for the list like `snapshots.include.<I>.value` and
+    `snapshots.include.<I>.type`.  Also it will add the size like
+    `snapshots.include.size`.
```
emtiu, line 42:
```suggestion
The script reads every line and tries to analyse it:
  - It searches for `DEFAULT` and puts those into a `dict` for later replacing
    the variable with the value.
  - If that didn't match it will look for lines starting with `#?` which is
    basically my own description for the manpage-entry.
    Multiple lines will get merged and stored in `commentline` until the
    processing of the current config option is done. That will reset
    `commentline`.
  - If a line starts with `#` it will be skipped.
  - Next the script searches for lines which ``return`` the config value (like
    `snapshots.ssh.port`. There it will extract the
    key/name (`snapshots.ssh.port`), the default value (`22`),
    the instance (`Int`) and if it is a profile or general value.
  - If the line contains a `List` value like `snapshots.include` it will
    process all values for the list like `snapshots.include.<I>.value` and
    `snapshots.include.<I>.type`.  Also it will add the size like
    `snapshots.include.size`.
```

common/create-manpage-backintime-config.py
```diff
@@ -63,150 +96,267 @@
...
+                if commentline \
+                   and ';' not in commentline \
+                   and not commentline.endswith('\\n'):
                     commentline += ' '
+
                 commentline += line.lstrip('#?').rstrip('\n')
+
                 continue
+
+            # Usual comments ignored
```
emtiu, line 256:
```suggestion
            # Simple comments are ignored
```

emtiu:
Some minor suggestions, but looks good otherwise :)

Out of curiosity: Is the manpage generation covered by testing somehow? Or does it only happen when we run the script manually?

buhtz:
> Out of curiosity: Is the manpage generation covered by testing somehow?

No.

> Or does it only happen when we run the script manually?

Only, when we run it manually. It is a "helper" script and conceptually don't belong to the release version of BIT. But looking into my Debian installation of BIT that py file is shipped with the release version.

aryoda:
I think my fix for https://github.com/bit-team/backintime/issues/1270 (schedule.time doc outdated) is still lost if we create the man pages again.

My changes from https://github.com/bit-team/backintime/commit/3c1e08e110246e2159f6666ad309664098cbb6eb need to be added here:

https://github.com/bit-team/backintime/blob/775c9c38eafb55bb2f3b91dbeef7db17ef5f2af5/common/config.py#L852-L855

Shall I commit to your PR or could you do the change for me (saves me the setup overhead ;-) ?

buhtz:
> Shall I commit to your PR or could you do the change for me (saves me the setup overhead ;-) ?

You mean this?
https://github.com/bit-team/backintime/issues/1270#issuecomment-1224344899

I'll do it for you. ❤️

aryoda:
Can we merge this?

Issue #1364, opened 2022-11-17
[OpenSuSE rolling] Missing icons in SSH private key buttons (BiT root)


ont4300:
Fresh install of 1.3.2 on Opensuse Tumbleweed.

The symbols in the push buttons beside the private key input box are missing. This applies to backintime-qt_polkit. The symbols are there when running backintime_qt.

![backman missing icons](https://user-images.githubusercontent.com/24478202/202551742-449e47d0-a641-48d1-8dd0-91a59c5a6f45.png)

ont4300:
I should add that this occurs when running as root...  Tom.

emtiu:
Thank you for your report! This is likely related to #1306, so maybe @aryoda could have a look :)

ont4300:
I've been thinking though (always dangerous at best!).

If I run backintime-qt_polkit, I am running the backup as root. In this case, I should be using the ssh keys for the root user and not any other user. I cannot change the private key settings. That is, the keys have to be from the root directory. In this case, not showing the symbols for "searching for a key", or "adding a key" would properly be blanked out.

In addition, if I click on where the button where the directory icon would be, the program presents a a directory window that can be scrolled through. However, a selected file is not moved to the private key window, and no hidden files or directories are shown.

I think this is what is going on. Whoever wrote the code was correct in preventing us from making a mistake.

Perhaps some mention of this should be made in the documentation.

Thanks... Tom.

aryoda:
Which desktop environment do you use? KDE, Gnome3...?

Could you please post the output of this diagnostics code here (copy into a terminal):

```
# if you are using X11:
pkexec env DISPLAY=$DISPLAY XDG_RUNTIME_DIR=$XDG_RUNTIME_DIR XAUTHORITY=$XAUTHORITY python3 -c "from PyQt5.QtGui import QIcon; from PyQt5.QtWidgets import QSystemTrayIcon,QApplication; app = QApplication(['']); print('isSystemTrayAvailable: ' + str(QSystemTrayIcon.isSystemTrayAvailable())); print('Theme name: ' + QIcon.themeName()); print('has theme icon <document-save>: ' + str(QIcon.hasThemeIcon('document-save'))); print('themeSearchPaths: ' + str(QIcon.themeSearchPaths())); print('fallbackSearchPaths: ' + str(QIcon.fallbackSearchPaths()))"

# if you are using Wayland:
[...]

PR #1367, opened 2022-11-19
Add logging for Qt5 platform, style, theme and system tray.

aryoda:
* Show QT QPA platform plugin in app window title for "--debug"
* Show "root" in app window title if running as root

This is a partial "fix" for #1364 by adding debug output and documenting icon problems as known issue

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,6 +144,22 @@ to support all Ubuntu distros (older and newer ones).
 This issue is tracked in [#1338](https://github.com/bit-team/backintime/issues/1338).
 
 
+
+#### Tray icon or other icons not shown correctly
+
+This effect can be caused by missing installations of Qt5-supported
+themes and icons. Back In Time may activate the wrong theme in this
+case leading to some missing icons. A fix for the next release is in preparation.
+
+As clean solution please check your Linux settings (Appearance, Styles, Icons)
+and install all themes and icons packages for your preferred style via
+your package manager.
+
+See issues [#1306](https://github.com/bit-team/backintime/issues/1306)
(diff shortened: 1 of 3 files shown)


buhtz:
This is your PR, run as normal user with `--debug` on Debian Testing (12).
grafik

`backintime-qt_polkit` doesn't run directly out of the repo because of the hardcoded /usr/bin/backintime-qt path in that script. I have to do the configure-make-install first. I can live with that short coming but if there is a better and more flexible solution...

This is your PR, after configure-make-install and run as root with `--debug` on Debian Testing (12).
grafik

Is "xcp" value you would expect?

buhtz (approved):

aryoda:
> Is "xcp" value you would expect?

Yes, it is the name of the Qt5 QPA plugin. Just showing "X11" or "Wayland" wouldn't be enough since this plugin
affects the Qt5 themes (incl. icons) too (eg. for X11 also `qt5ct` could be possible.

More documentation for developers will be contained in may upcoming PR in `$prj_root/doc-dev` (first draft is pushed to my personal repo only ATM: https://github.com/aryoda/backintime/blob/PR/add_doc_dev_with_wayland_doc/doc-dev/From_X11_to_Wayland_with_Qt5.md)

```
DEBUG: [qt/qttools.py:179 createQApplication] QT active style: fusion
DEBUG: [qt/qttools.py:180 createQApplication] QT fallback style: hicolor
DEBUG: [qt/qttools.py:181 createQApplication] QT supported styles: ['cleanlooks', 'gtk2', 'cde', 'motif', 'plastique', 'Windows', 'Fusion']
```

PR #1376, opened 2022-11-27
"Shutdown" translation not clear enought in French

jej:
"Arrêt" means "Stop" and is not evident (stop what?). I clicked on it, thinking I would stop the backup, and I was very surprised to see the machine going down...

Shutdown is stronger, so here "Éteindre" is better for the French translation. This translation is already used just below:
https://github.com/bit-team/backintime/blob/3132f4dd4fc627aef46e152696f5b98fb3db0298/common/po/fr.po#L701-L702

Hope it helps,
J.

### DIFF
--- a/common/po/fr.po
+++ b/common/po/fr.po
@@ -695,7 +695,7 @@ msgstr "Préférences"
 
 #: ../../qt/app.py:142
 msgid "Shutdown"
-msgstr "Arrêt"
+msgstr "Éteindre"
 
 #: ../../qt/app.py:143
 msgid "Shutdown system after snapshot has finished."


buhtz:
Dear @jej ,
thank you very much for your contribution.

We need to update the contribution section of README.md about translations.

I take your PR and modified the [original po file on launchpad](https://translations.launchpad.net/backintime/trunk/+pots/back-in-time/fr/130/+translate). The po-files from launchpad will be copied into that GitHub repo before each release. The po-files in our GitHub repo are not the "original files" and they will be overwritten the next time.

The team is aware that this isn't an elegant solution, and we will change that after the next release.

I'll keep this PR open until someone who is able to understand French did a review. ;)

emtiu (approved):
My French is rather basic, but this makes sense. Also, it is consistent with uses of "Arrêt" and "Éteindre" that I've found by a quick search. So, I approve :)

buhtz:
Again, thank you very much for your contribution.

We have [8 untranslated items in _French_ left](https://translations.launchpad.net/backintime/trunk/+pots/back-in-time/fr/+translate?start=0&batch=10&show=untranslated&field.alternative_language=&field.alternative_language-empty-marker=1&old_show=all). Maybe you can give it a shot? 😄

jej:
Ok, I will have a look for the 8 strings and the whole. I need to use backintime a bit more to see if there are some weird strings... At a first glance it was mostly smooth.

Issue #1401, opened 2023-01-29
New Version Won't Open: 1.3.3 (ppa:bit-team/stable, Debian unstable/bookworm, Ubuntu unstable/Lunar Lobster)


ryankingart:
I upgraded to the latest version of Back in Time via the update manager, but when I try to open Back in Time, or Back in Time Root, it simply will not open. I tried re-installing it, and I also tried re-starting my computer, but that doesn't work either. The program will just not open.

My OS: Linux Mint 19.3 Cinnamon

Is there a possibility that the new version of back in time doesn't work on my OS? I know that my current Linux Mint version is old, and will be unsupported in april. So I plan on installing the latest version of Linux Mint soon. But I need to backup my files before I do that.

If anyone can help, I'd appreciate it.
Thanks!

aryoda:
> I upgraded to the latest version of Back in Time via the update manager

Did you use [our PPA](https://github.com/bit-team/backintime#ubuntu-ppa) for that or another source?

Could you please add the output of

`backintime --debug`

and

`backintime --diagnostics`

here to make it possible for me to dig deeper into the problem?

It may just be a dependency issue (or even a too old python version)...

ryankingart:
Hello. Thanks for the quick reply!

I installed the program with these commands:
`sudo add-apt-repository ppa:bit-team/stable`
`sudo apt-get update`
`sudo apt-get install backintime-qt`

**Here is the terminal output of: backintime --debug**

Traceback (most recent call last):
  File "/usr/share/backintime/common/backintime.py", line 28, in <module>
    import config
  File "/usr/share/backintime/common/config.py", line 45, in <module>
    import tools
  File "/usr/share/backintime/common/tools.py", line 37, in <module>
    from packaging.version import Version
ModuleNotFoundError: No module named 'packaging'

**Here is the terminal output of: backintime --diagnostics**

Traceback (most recent call last):
  File "/usr/share/backintime/common/backintime.py", line 28, in <module>
    import config
  File "/usr/share/backintime/common/config.py", line 45, in <module>
    import tools
  File "/usr/share/backintime/common/tools.py", line 37, in <module>
    from packaging.version import Version
ModuleNotFoundError: No module named 'packaging'

aryoda:
This is a missing dependency. Please also install:

`sudo apt install python3-packaging`

and try it again (start GUI and in case of problems post the `backintime --debug` output here again.

ryankingart:
[...]

Issue #990, opened 2019-05-01
"Save password to keyring" option is disabled


Azimoth:
I'm having this issue using version 1.2.0~alpha0 on Gentoo. 
The thing is: It WAS working until I installed something else that required PyQt5. 

Then backintime stopped working properly by complaining about being unable to access the password, so I removed it manually from kwallet, expecting to be asked for it again.
Then the GUI didn't come up anymore, which I could fix by installing xdpyinfo.

Now Backintime is working again, but the "Save password to keyring" option is disabled.

I also tried re-installing Backintime (also removing ~/.config/backintime and ~/.local/share/backintime), but it still doesn't work.

> backintime --debug

doesn't show anything that seems to be out of the ordinary...

I am well aware of #545, #587, #509 and #534 but the solutions presented there didn't work.+
`python-secretstorage` is installed, 

```
$ python -c "import keyring; print(keyring.get_keyring())"
keyring.backends.chainer.ChainerBackend (priority: 10)
```
```
$ backintime pw-cache stop

Back In Time
Version: 1.2.0~alpha0

Back In Time comes with ABSOLUTELY NO WARRANTY.
This is free software, and you are welcome to redistribute it
under certain conditions; type `backintime --license' for details.

WARNING: Failed to read PID and process name from /home/user/.local/share/backintime/password_cache/PID: [2] No such file or directory
ERROR: pidfile /home/user/.local/share/backintime/password_cache/PID does not exist. Daemon not running?
```
Since I added the password to my ssh-agent (via ksshaskpass) I don't have to enter it, when I open back in time manually. Only taking snapshots or changing settings require a password. (When cancelling without password, there's a message the password is "wrong or not available for cron".)

What am I missing here?

joaoherberto:
I am having a similar issue, I am on an Arch based system, current BIT version is 1.1.12.  I am wondering if the dependency of python3-keyring may be the issue. on an Arch system, the files are named python-keyring (it installs python3). Perhaps one of the devs could way in on this.

Does BIT look for python3 or just python paths?

Azimoth:
I just noticed that 

`$ backintime backup`

also works without manually entering a password. So the issue only exists using the qt gui...

bentolor:
Same here on Ubuntu 18.10 / Gnome Shell. My error is:

     [common/tools.py:822 keyringSupported] No appropriate keyring found. 'keyring.backends.chainer' can't be used with BackInTime

[...]

PR #1402, opened 2023-01-30
Hotfix for issue #1401 (missing dependencies in Ubuntu PPA package)

aryoda:
Proposed hotfix for #1401.

Needs to be merged into

* main
* dev (must be done in the cmd line I guess since the Github GUI does support only one target branch for a PR):
   `git switch dev` + `git merge main`

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -9,7 +9,7 @@ Homepage: https://github.com/bit-team/backintime
 
 Package: backintime-common
 Architecture: all
-Depends: rsync, cron-daemon, openssh-client, python3-keyring, python3-dbus, ${python3:Depends}, ${misc:Depends}
+Depends: rsync, cron-daemon, openssh-client, python3-keyring, python3-dbus, python3-packaging, ${python3:Depends}, ${misc:Depends}
 Recommends: sshfs, encfs
 Conflicts: backintime
 Replaces: backintime
@@ -19,7 +19,7 @@ Description: Simple backup system (common)
 Package: backintime-qt
 Architecture: all
 Depends: x11-utils, libnotify-bin, python3-pyqt5, python3-dbus.mainloop.pyqt5, policykit-1, backintime-common (>= ${source:Version}~), ${python3:Depends}, ${misc:Depends}
-Recommends: python3-secretstorage
+Recommends: python3-secretstorage, python3-keyring-kwallet, python3-gnomekeyring
 Suggests: meld | kompare
 Conflicts: backintime-qt4 (<< ${source:Version}~), backintime-kde (<< ${source:Version}~), backintime-kde4 (<< ${source:Version}~), backintime-gnome (<< ${source:Version}~), backintime-notify (<< ${source:Version}~)
 Replaces:  backintime-qt4 (<< ${source:Version}~), backintime-kde (<< ${source:Version}~), backintime-kde4 (<< ${source:Version}~), backintime-gnome (<< ${source:Version}~), backintime-notify (<< ${source:Version}~)
--- a/debian/control
+++ b/debian/control
@@ -19,7 +19,7 @@ Description: Simple backup system (common)
 Package: backintime-qt
 Architecture: all
 Depends: x11-utils, libnotify-bin, python3-pyqt5, python3-dbus.mainloop.pyqt5, policykit-1, backintime-common (>= ${source:Version}~), ${python3:Depends}, ${misc:Depends}
-Recommends: python3-secretstorage, python3-keyring-kwallet, python3-gnomekeyring
+Recommends: python3-secretstorage
 Suggests: meld | kompare
 Conflicts: backintime-qt4 (<< ${source:Version}~), backintime-kde (<< ${source:Version}~), backintime-kde4 (<< ${source:Version}~), backintime-gnome (<< ${source:Version}~), backintime-notify (<< ${source:Version}~)
 Replaces:  backintime-qt4 (<< ${source:Version}~), backintime-kde (<< ${source:Version}~), backintime-kde4 (<< ${source:Version}~), backintime-gnome (<< ${source:Version}~), backintime-notify (<< ${source:Version}~)


aryoda:
@Germar If you update our `ppa:testing` after this hotfix made its way into the `main` branch I can test the package installation of my Ubuntu systems ...

buhtz:
Please don't merge this yet because this needs more discussion, also about some basic ("grundsätzliche") topics. :)

Conceptual the PPA is not "upstream". It is a distro-specific and community-maintained source. That one of the upstream maintainers (Germar) do manage that PPA is a coincidence but doesn't mean that it is the same.

IMHO the GitHub repo and the PPA need to be treated different and don't follow the same rules.

On upstream I vote to merge the dependency thing only to `dev`. There is no need for a hotfix on upstream for this.

On PPA of course there is also no need for a "hotfix" but for a fix of the package. The version keeps 1.3.3 but the package version need to increment. "1.3.3-2" would be the name. The files in `/debian` don't belong to upstream because they are distro specific. Changing them won't change upstream code.
That this folder is there just has historical and organizational reasons. For example all the PPA related stuff could be in its own repo on Launchpad. But this can be discussed later on `bit-dev`.

About dependencies themself. I assume that the kdewallet and gnome wallet recommendations won't make it into an official Debian package. "recommends" are selected and installed by default. In that case the user would trigger the installation of a lot of KDE and Gnome packages depending on those booth. I'm also not sure about the consequences of having this in the PPA package.

aryoda:
> Please don't merge this yet because this needs more discussion

Yes, I see your points, thanks for raising the red flag :-)

> Conceptual the PPA is not "upstream". It is a distro-specific and community-maintained source. 
> [...]
> The files in /debian don't belong to upstream because they are distro specific (Ubuntu-PPA only. Debian does ignore our /debian.). Changing them won't change upstream code.
That this folder is there just has historical and organizational reasons. For example all the PPA related stuff could be in its own repo on Launchpad.

Correct, this must be clearly separated from our BiT source code in the future.

> On PPA of course there is also no need for a "hotfix" but for a fix of the package.

Please ignore my "dramatic" wording "hotfix" here, I just wanted to make sure

- we fix this soon
- the change must be merged into `dev` and `main` (that is how a hotfix works)

I personally consider this as an "hotfix" for the PPA package, not BiT itself (as you also do).

> On upstream I vote to merge the dependency thing only to dev. There is no need for a hotfix on upstream for this.

I am not quite sure about our workflow to create a new package on our Ubuntu PPA,
but do we really want to update our PPA:stable package from a `dev` branch (not from `main`)`?

> About dependencies themself. I assume that the kdewallet and gnome wallet recommendations won't make it into an official Debian package. "recommends" are selected and installed by default. In that case the user would trigger the installation of a lot of KDE and Gnome packages depending on those booth.

After reconsidering this I think I should indeed remove these both recommended packages again because the installation may probably cause that `keyring` is choosing another default keyring so that credentials that are already stored in one keyring implementation may not be found anymore in the new default keyring (see https://github.com/bit-team/backintime/issues/990#issuecomment-1272014311).

Our BiT CLI should indeed not depend on any UI stuff (KDE or Gnome).

> After some quick n dirty research I'm also [not sure that this two packages even exist in Debian](https://tracker.debian.org/) and [maybe Ubuntu](https://packages.ubuntu.com/). Where do you get this names from?

These are Ubuntu packages but Ubuntu uses DEB files requiring the dependencies in the `control` file.
So I think our folder name `Debian` is somehow misleading if the content is only used to update our **Ubuntu** PPA...

**So: On what next steps can we agree?**

- I need confirmation that the changed `control` file is indeed only used to feed our Ubuntu PPA
- I would like to remove at least the "Recommend" changes from this PR
- ...

buhtz:
Moin again,

I understand the term "hotfix" in the context of our branching model. In that case a hotfix implicates a new version number. So if you would "hotfix" it means you will create a new release "1.3.4". That is why it is hot 🌶️ . 😄 

> I am not quite sure about our workflow to create a new package on our Ubuntu PPA,
but do we really want to update our PPA:stable package from a dev branch (not from main)`?

This is up to Germar to decide. I assume it isn't a big problem to create a deb file based on the `dev` branch.
On the other hand I see no big problem to integrate that modification also into `main`. This would IMHO be an acceptable exception to our branching model (where usually `main` commits are releases only). But just with a simple commit, not with a release tag.

> These are Ubuntu packages but Ubuntu uses DEB files requiring the dependencies in the control file.
> So I think our folder name Debian is somehow misleading if the content is only used to update our Ubuntu PPA...

deb-files (for Ubuntu) are created with Debian tools. They expect a `/debian` folder. I'm not sure of course but assume that all debian-based Distributions not named "Debian" do use such folders.

> I need confirmation that the changed control file is indeed only used to feed our Ubuntu PPA

I can't confirm exactly that but confirm that Debian defenitly doesn't use that folder.
See the [`debian` branch](https://salsa.debian.org/jmw/pkg-backintime/-/tree/debian/) of the backintime package on the "Salsa" (Debians own GitLab) instance. In the same repo there is the upstream source (a 1:1 copy of our GitHub repo) in the [`upstream` branch](https://salsa.debian.org/jmw/pkg-backintime/-/tree/upstream). The `debian` folder in the latter is our upstream-`debian`-folder that shouldn't belong there as we discussed. Comparing [our upstream `control` file](https://salsa.debian.org/jmw/pkg-backintime/-/blob/upstream/debian/control) and Debian's own [`control` file](https://salsa.debian.org/jmw/pkg-backintime/-/blob/debian/debian/control) you see there is much difference.

Beside this Salsa instance you can find on the [Debians package tracker](https://tracker.debian.org/pkg/backintime) a link called [browse source code](https://sources.debian.org/src/backintime/unstable/) where you can look into the source of the selected package.

buhtz:
> * I would like to remove at least the "Recommend" changes from this PR

You could move the two packages from "Recommend" into "Suggested". See https://www.debian.org/doc/debian-policy/ch-relationships.html#syntax-of-relationship-fields

Germar:
> > I am not quite sure about our workflow to create a new package on our Ubuntu PPA,
> > but do we really want to update our PPA:stable package from a dev branch (not from main)`?
> 
> This is up to Germar to decide. I assume it isn't a big problem to create a deb file based on the `dev` branch. On the other hand I see no big problem to integrate that modification also into `main`. This would IMHO be an acceptable exception to our branching model (where usually `main` commits are releases only). But just with a simple commit, not with a release tag.

I create the PPA from a local checkout. So for the PPA it doesn't matter if you update `main` or just `dev`. But we provide the possibility to build `.deb` files from source with `makebeb.sh`. And because of this I'd recommend to update both.



> > These are Ubuntu packages but Ubuntu uses DEB files requiring the dependencies in the control file.
> > So I think our folder name Debian is somehow misleading if the content is only used to update our Ubuntu PPA...
> 
> deb-files (for Ubuntu) are created with Debian tools. They expect a `/debian` folder. I'm not sure of course but assume that all debian-based Distributions not named "Debian" do use such folders.

`/debian` is crucial for the `.deb` build. Please do not rename.

> > * I would like to remove at least the "Recommend" changes from this PR
> 
> You could move the two packages from "Recommend" into "Suggested".

Please don't set those dependencies. Neither as recommend nor suggested. `keyring-secretstorrage` is the minimum that is required. `gnome-keyring` or `kde-keyring` will be installed by the desktop environment.

aryoda:
OK, I have removed the keyring dependencies again, please review.

> I create the PPA from a local checkout. So for the PPA it doesn't matter if you update main or just dev. But we provide the possibility to build .deb files from source with makedeb.sh.

THX, no I understand the build workflow. It uses `dpkg-buildpackage` internally which `chroot`s into the `debian` folder (which is why the folder must be named as it is).

I have put this topic on my TODO list (to document how to build the DEB pkg for Ubuntu).

Last missing part (for me) is how to upload this to launchpad.

Germar:
I just pushed the new version to the testing PPA.

> Last missing part (for me) is how to upload this to launchpad.

For this I set the new version in `VERSION` and `CHANGES` and run `updateversion.sh`. This will change the version in every other file and create a new `debian/changelog` with my name at the end. This is crucial for the build process later on.

Next I `git commit` those changes and create a new tar archive with `make-tarball.sh` which will actually `git clone` current branch into a new folder `../backintime-$VERSION` and than make an archive from that new folder. This way we can be sure there are no left-over files inside the release nor in the PPA.

Inside this new folder I run `makedeb-src.sh` which will create all the necessary files for the build service and sign it with my personal GPG key. That's why it is important that `debian/changelog` end with my name. To publish those files I run `dput ppa:bit-team/testing *.changes`. Without my signature the Launchpad servers will reject those files.

aryoda:
> I just pushed the new version to the testing PPA.

Did you manually merge my PR into your checked-out source code before building and pushing to PPA?

I think the control file of the DEB pkg does still not contain the missing `python3-packaging` dependency (I did a manual download and opened it via my Archive Manager to decompress it...)

PS: THX for describing the PPA publishing workflow, I have added this missing part to my TODO list for `dev-doc` too.

aryoda:
@Germar OK, I have merged this PR with the missing dependency into `main` and `dev` now.

Did you manually merge my PR into your checked-out source code before building and pushing to PPA?

I think the control file of the DEB pkg does still not contain the missing `python3-packaging` dependency (I did a manual download and opened it via my Archive Manager to decompress it...)

Could someone please check this?

Germar:
> Did you manually merge my PR into your checked-out source code before building and pushing to PPA?
> 
> I think the control file of the DEB pkg does still not contain the missing `python3-packaging` dependency

No, I changed it manually in my local clone. But before launching `make-tarball.sh` I remembered, that I've to commit changes before. I didn't want to commit it to `main`, so I reverted all changes, created a new branche and did all the changes again. Looks like I forgot the most important change the 2nd time 🤦‍♂️, sorry.
I'll make a new release tomorrow.

Germar:
I pushed it again to both testing and stable PPA

aryoda:
> I pushed it again to both testing and stable PPA

THX a lot, it looks good now (includes the missing dependency):

https://launchpadlibrarian.net/649192535/backintime_1.3.3~kinetic_1.3.3-3~kinetic.diff.gz

Issue #1409, opened 2023-02-24
backintime-cli fails unit tests (Gtk-WARNING ... Theme parsing error)


KaibutsuX:
Attempting to install v1.3.3-3

Both backintime and backintime-cli fail with the same test failure.

OS: Arch/Xfce

```
======================================================================
FAIL: test_local_snapshot_is_successful (test.test_backintime.TestBackInTime)
end to end test - from BIT initialization through snapshot
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/kaibutsux/.cache/yay/backintime/src/backintime-1.3.3/common/test/test_backintime.py", line 147, in test_local_snapshot_is_successful
    self.assertRegex(filtered_log_output, re.compile(r'''INFO: Lock
AssertionError: Regex didn't match: 'INFO: Lock\nINFO: Take a new snapshot. Profile: 1 Main profile\nINFO: Call rsync to take the snapshot\nINFO: Save config file\nINFO: Save permissions\nINFO: Create info file\nINFO: Unlock\n' not found in "INFO: Lock\nINFO: Take a new snapshot. Profile: 1 Main profile\nINFO: Call rsync to take the snapshot\nINFO: Save config file\nINFO: Save permissions\nINFO: Create info file\n\n(python:2194333): Gtk-WARNING **: 10:20:50.946: Theme parsing error: gtk.css:4419:85: '/*' in comment block\n\n(python:2194333): Gtk-WARNING **: 10:20:50.946: Theme parsing error: gtk.css:4512:64: '/*' in comment block\nINFO: Unlock\n"

----------------------------------------------------------------------
Ran 3 tests in 3.388s

FAILED (failures=1, skipped=1)
make: *** [Makefile:570: unittest] Error 1
==> ERROR: A failure occurred in check().
    Aborting...
 -> error making: backintime (backintime-cli backintime)
```

buhtz:
Just to improve reading I reformatted the output

```
AssertionError: Regex didn't match:
    'INFO: Lock\n
     INFO: Take a new snapshot. Profile: 1 Main profile\n
     INFO: Call rsync to take the snapshot\n
     INFO: Save config file\n
     INFO: Save permissions\n
     INFO: Create info file\n
     INFO: Unlock\n'

     not found in

     "INFO: Lock\n
      INFO: Take a new snapshot. Profile: 1 Main profile\n
      INFO: Call rsync to take the snapshot\n
      INFO: Save config file\n
[...]

PR #1411, opened 2023-02-24
Fixed test with polluted output by gtk-warning messages

KaibutsuX:
Related to #1409

PR #1412, opened 2023-02-25
Unittest: Workaround fix parsing output

buhtz:
This should fix the problem with polluted output with Gtk-Warnings while testing (see #1409).

It is still a workaround. The whole test should be refactored. But it is an *integration test* or somewhere higher in the *test pyramid*. It means before refactoring this test a lot of other *unit test* lower in the test pyramid and there affected units need to refactored.

The current workaround is to exclude lines from the catched output before testing that output via `assertRegex`. Some lines removed because they do begin with a string (DBus stuff). Others excluded because they contain a string (Gtk-Warnings).

The filter and exclude mechanic now is more efficient using `filter()` and list comprehension instead of explicit `for` loops.

Because #1409 is not reproducible currently I'm not able to test that.

### DIFF
--- a/common/test/test_backintime.py
+++ b/common/test/test_backintime.py
@@ -138,10 +138,8 @@ under certain conditions; type `backintime --license' for details.
         #       regex but if the BiT serviceHelper.py DBus daemon is not
         #       installed at all the warnings also occur in the middle of below
         #       expected INFO log lines so they are removed by filtering here.
-        # TODO If filtering output rows should become a common testing pattern
-        #      this code should be refactored into a function for re-usability.
+        #       The same goes with Gtk warnings.
 
-        # checked via str.startswith()
         line_beginnings_to_exclude = [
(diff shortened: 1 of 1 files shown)


buhtz:
The build error can be ignored. It is just one job (at the pce64 machine) and is IMHO not related to BIT itself. Something is wrong with the travis machine while cloning our repo.

aryoda:
clever solution, using anonymous functions (`lambda`s) for filtering :-)

aryoda:
Is this PR `xor` to #1411 (which could be closed without merging then)?

buhtz:
> Is this PR `xor` to #1411 (which could be closed without merging then)?

Yes correct. That was my plan.
If there is no hurry on your site I would wait some days (and sleepy nights) before doing this. It is just a minor modification but in my experience it is better to let things hang out a bit ("Abhängen lassen" i.S. von Reifen). :smile:

PR #1413, opened 2023-03-03
Add ChainerBackend keyring support to fix password lookup error (#1410) 

### LINKED ISSUE #1410: backintime-qt: Fails to access password with new keyring package (ChainerBackend not supported)
## Overview

A [bug report at Debian](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=998105) (many thanks to Sven Geuer) and also others here (see tag [`Keyrings`](https://github.com/bit-team/backintime/labels/Keyrings)) showed that
the new `ChainerBackend` of the `keyring` package is now the default backend and it is not supported in BiT
(which white-lists the supported keyring backends).

## Symptoms

Logging output (DEBUG log level):

> No appropriate keyring found. 'keyring.backends.chainer' can't be used with BackInTime

For details see also this FR #1330 which explains the backgrounds.

## Proposed fix

I propose to fix this bug ASAP by enabling `ChainerBackend` support in BiT.
I had already documented and prepared this fix three months ago:

https://github.com/bit-team/backintime/blob/e22c7f253fa048fab9119f4cbea34c0fa8d1aba6/common/tools.py#L889-L897

`Chainer` will loop over all installed keyring backends (password safes) and tries to find the requested password:

https://github.com/jaraco/keyring/blob/7ffff5b285d6f9d9b8cba33a9d5e9071cdd507b9/keyring/backends/chainer.py#L47-L51

**Possibly unwanted side-effect:** Setting a password uses the highest-ranked password safe (if more than one is installed) so updating a password may leave the old password in a lower-ranked password safe untouched and create a new entry in another higher-ranked password safe. This "change in password-safe ranking" may e.g. occur if another password safe is installed...

https://github.com/jaraco/keyring/blob/7ffff5b285d6f9d9b8cba33a9d5e9071cdd507b9/keyring/backends/chainer.py#L53-L58

## Alternatives

- Probe the supported backends for already stored BiT credentials and choose the matching backend automatically
  (if only one: How to cope with password in different password safes if BiT does only pick one?)

- BiT could update an existing password in the same backend where the old password exists
(may not be the same backend chosen by ChainerBackend!)

aryoda:
Fixes #1410 

Urgently required for the upcoming Debian BiT release.

Successfully ested via my local `openSSH` server and a password on my private key file...

For testing you can change the default keyring backend via a config file
(see instructions: https://github.com/bit-team/backintime#non-working-password-safe-and-bit-forgets-passwords-keyring-backend-issues)

Please review.

Further tests welcome.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -2,6 +2,7 @@ Back In Time
 
 Version 1.3.4-dev (development of upcoming release)
 * Project: Renamed branch "master" to "main" and started "gitflow" branching model.
+* Fix bug: Add support for ChainerBackend class as keyring which iterates over all supported keyring backends (#1410)
 
 Version 1.3.3 (2023-01-04)
 * New feature: Command line argument "--diagnostics" to show helpful info for better issue support (#1100)
--- a/common/tools.py
+++ b/common/tools.py
@@ -42,6 +42,7 @@ try:
     if os.getenv('BIT_USE_KEYRING', 'true') == 'true' and os.geteuid() != 0:
         import keyring
         from keyring import backend
+        import keyring.util.platform_
 except:
     keyring = None
     os.putenv('BIT_USE_KEYRING', 'false')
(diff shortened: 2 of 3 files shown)


aryoda:
@buhtz One of the Travis CI builds did not succeed (looks like a connection timeout). Do you have the rights to trigger the build again to be sure no unit test is failing in our build matrix? THX!

```
$ git clone --depth=50 https://github.com/bit-team/backintime.git bit-team/backintime
No output has been received in the last 10m0s, this potentially indicates a stalled build
or something wrong with the build itself.
...
The build has been terminated
```

buhtz:
You had a long night. 👍 

About TravisCI: The problem is known but not understood yet. Happens only on "ppc64le" machine. I retriggered and it worked now. The Travis support told me to logout and login again. That worked also. If this happens more often I will recontact the support. But maybe we should switch from Ubuntu20 to Ubuntu22 after the Debian release.
You can create your own TravisCI account if you want. I assume the GitHub-Team-setup is reflected in Travis and you should be able to trigger builds then.

About your PR: Looks OK for me but I'm not a keyring user or pro. My review isn't much of a value. I would only say the code is not PEP8 conform (e.g. length of lines). But we have bigger "problems". 😄

aryoda:
> I would only say the code is not PEP8 conform (e.g. length of lines)

Yes, I have to admit I quite often ignore the PEP8 max length of line rule for better readability (my quite personal and biased opinion ;-)

I am merging now, for Debian a PR containing this code as patch is on its way:

https://salsa.debian.org/jmw/pkg-backintime/-/merge_requests/9

PR #1414, opened 2023-03-04
Fix typos

kianmeng:
Found via `codespell -S *.po,CHANGES -L dum,clude,nin,sistem`


common/cli.py
```diff
@@ -97,7 +97,7 @@ def errorHandler(msg):
             return False
         okay()
 
-    test = 'Check/prepair snapshot path'
+    test = 'Check/prepare snapshot path'
```
emtiu, line 100: Do we need to take care of translations here? Or is this a debug message that never gets translated?

emtiu, line 100: Do we need to take care of translations here? Or is this a debug message that never gets translated?

common/po/messages.pot
```diff
@@ -950,7 +950,7 @@ msgid "Changes"
 msgstr ""
 
 #: ../../qt/logviewdialog.py:100
-msgid "Informations"
+msgid "Information"
```
emtiu, line 953: Pretty sure that this would mess up translations.

common/tools.py
```diff
@@ -1666,12 +1666,12 @@ def __init__(self, callback = None, overwrite = True):
         Args:
             callback: Function to call when the timer ran down
                       (ensure calling only reentrant code).
-                      Use ``None`` toi throws a ``Timeout`` exception instead.
+                      Use ``None`` to throws a ``Timeout`` exception instead.
```
emtiu:
```suggestion
                      Use ``None`` to throw a ``Timeout`` exception instead.
```

emtiu (changes requested):
Thanks, I've reviewed all the suggested changes, and there's:

- 1 possible improvement
- 1 thing we should examine for potential trouble
- 1 thing that would certainly cause trouble with translations

buhtz:
Dear Kian-Meng,

thanks a lot for your contribution. I like people taking care of such details!

I set the PR in "Draft" mode to prevent accidentally merging this into `main`. I also changed the target branch from `main` to `dev`. Please don't worry. The "problem" is on the site of the maintaining team. We still try to figuring out how to conduct contributors into the right direction.

I'll explain but I also have the opinion that contributors don't have to know about this. There must be a way to prevent situations like this and I'm working on it figuring this out.

Because of the situation your contribution also do improve our maintaining workflow. Thanks for that also! :heart: 

First of all all PR's should be targeting the `dev` branch not `main`. This is because of our branching model known as [git-flow](https://nvie.com/posts/a-successful-git-branching-model), where `dev` always represent the latest development state and `main` do contain the stable releases only.

When cloning/forking from our repo I assume you create your `fix-typos` branch based on `main` not `dev`, right? This also causing problems when targeting the PR to `dev`. For example you now can see in the "Commits" section that there is a new commit not created from you.

common/cli.py
buhtz: Makes me a headache diving deeper in this code and code that is using this part. :smile: 

In the context of this PR I would say it is OK and we should leave it.

common/po/messages.pot
buhtz: "mess up" is a bit to much. Yes, it will modify the translation and do open the need to translate that part again.

But we do this in the `dev` branch and the next release is far away. There is enough time to update the translation. We still do have some work todo with the translation and the related workflow. I see no need to make this a part of that PR.

buhtz:
OK, I think I fixed the branching-problem. Remember the problem was that `fix-typos` branches was based on `main`. What I've done so far (but snipping some details):

 1. I cloned `kianmeng`'s repo.
 2. Add `bit-team/backintime` as _remote upstream_ (via `git remote add upstream git@....`)
 3. Then I merged the `dev` branch into `kianmeng:fix-typos` (`git checkout fix-typos && git merge remotes/upstream/dev`)
 4. And then `push`this back to `kianmeng:fix-typos` which also automatically updates this PR.

I carefully reviewed again the _Files Changed_ section to be sure that no harm is done to our repo.

common/po/messages.pot
emtiu: Okay, I think I understand. If you know what's going on here, I'm fine with following your plan. I'd just like to avoid leaving this "hanging", for fear that it might come back to "bite" us later when we've forgotten about this change ;)

buhtz: It will just appear as a new & untranslated string in the translation frontend (currently launchpad).

kianmeng:
@buhtz Thanks for the details explanation, perhaps extract the https://github.com/bit-team/backintime#guidelines--rules into a separate `CONTRIBUTING.md` guide somehow?

emtiu (approved):

buhtz:
Just FYI:
Creating the commit message I tried to follow the [Conventional Commits Guideline](https://www.conventionalcommits.org/en/v1.0.0/) in combination with [similar guideline from Angular project](https://github.com/angular/angular/blob/22b96b9/CONTRIBUTING.md#type).

Here I was using `style: ` as prefix of first commit line.

kianmeng:
:partying_face: :partying_face: :partying_face: :partying_face: :partying_face:

PR #1415, opened 2023-03-04
Test TravisCI with Ubuntu Jammy 22.04

buhtz:
Not intended to merge this. It just a test.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ arch :
 
 # ensures that we have UUID filesystem mounts for proper testing
 sudo: true
-dist: focal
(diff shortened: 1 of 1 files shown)


buhtz:
OK, what we have learned from this test
 
1. It isn't without problems moving TravisCI from Ubuntu 20 to 22.
2. Ubuntu 22 doesn't support Python 3.6. Because of that we need an extra Ubuntu 20 test to include Python 3.6.
3. The unclear problem with the "ppce64le" machine persists also in the Ubuntu 22. I don't know why but I'm in contact with support. Our codebase is not the problem.

PR #1433, opened 2023-05-06
fix: Extra backslash in config-manpage (Debian Lintian error 734: bad argument name 'P)

### LINKED ISSUE #1398: manpage autogeneration: Problem with escaping backslashes
Related to the [packaging process at Debian](https://salsa.debian.org/jmw/pkg-backintime/-/merge_requests/6) I was pointed to a tiny problem with the manpages which maybe caused by the script [`create-manpage-backintime-config.py`](https://github.com/bit-team/backintime/blob/master/common/create-manpage-backintime-config.py) that auto-generates the manpage.

https://github.com/bit-team/backintime/blob/979b757d2cbbe65b332c5fd15244d400a9d80983/common/man/C/backintime-config.1#L728-L735

Line 734 has three `\` but it should be only two.

    Default: 'PATH=/opt/bin:/opt/sbin:\\\$PATH'
                                      ^^^

That line is generated from [that value](https://github.com/bit-team/backintime/blob/979b757d2cbbe65b332c5fd15244d400a9d80983/common/config.py#L144) where are only two `\`:

    DEFAULT_SSH_PREFIX = 'PATH=/opt/bin:/opt/sbin:\\$PATH'
                                                  ^^

I will look into the script later and fix it.

Lintian does use this command to check such manpages.
```
$ LC_ALL=C.UTF-8 MANROFFSEQ='' MANWIDTH=80 man --warnings -E UTF-8 -l -Tutf8 -Z "backintime-config.1" > /dev/null

troff: <standard input>:734: bad argument name 'P'
```

buhtz:
The manpage-generating script no more add an extra backslash to `\\$`. The generated manpage now has no syntax errors.

Technical details described in the issue #1398.

Looking into the diff it might be radical just to remove the `replace()` call. Comparing the manpages generated before and after I couldn't see a negative effect. Also in the future the manpage generation process will get replaced by a new script using pythons code introspection mechanic.

Fix #1398


buhtz:
Just merging this because the core code base isn't involved and I see risk.

PR #1436, opened 2023-05-07
feat: Exclude swapfile by default.

### LINKED ISSUE #1035: Exclude /swapfile by default
This is a default on my Ubuntu install, and I think definitely should be excluded in some way by default, and referenced in the suggestion list of what to exclude. Otherwise you're backing up a huge always-changing file.

(I realize that swap partitions are better - I have my reasons for not using them on this system)

buhtz:
When creating a new snapshot profile "swapfile" is excluded by default.

Resolves: #1035

I checked that this list of default values doesn't affect existing snapshot profiles but only the creation of new ones.


common/config.py
```diff
@@ -129,10 +129,28 @@ class Config(configfile.ConfigFileWithProfiles):
...
+        '/dev/*',
+        '/run/*',
+        '/etc/mtab',
+        '/var/cache/apt/archives/*.deb',
+        'lost+found/*',
+        '/tmp/*',
+        '/var/tmp/*',
+        '/var/backups/*',
+        '.Private',
+        'swapfile',
```
aryoda: Should the exclude pattern be restricted to the root folder (= default location of the swap file)?

buhtz: I thought about that, too. But I assume that some people do locate it in secondary mounted volume (e.g. `/mnt/anotherhdd/swapfile`) or something like that.

On the other hand I thought if there are files named "swapfile" somewhere else with different purposes.

Maybe it is saver to restrict it to root folder with `/swapfile` ?

CHANGES
```diff
@@ -4,6 +4,7 @@ Version 1.3.4-dev (development of upcoming release)
 * Project: Renamed branch "master" to "main" and started "gitflow" branching model.
 * Fix bug: Add support for ChainerBackend class as keyring which iterates over all supported keyring backends (#1410)
 # Fix bug: Unit test fails on some machines due to warning "Ignoring XDG_SESSION_TYPE=wayland on Gnome..." (#1429)
+* Feature: Exclude swapfile by default (#1053)
```
```suggestion
* Feature: Exclude /swapfile by default (#1053)
```

emtiu (commented):
Looks good, let's reflect the decision to exclude /swapfile only in the CHANGES.

Issue #1358, opened 2022-11-07
Python 3.5 support broken


Germar:
Hi,
I just fired up my (quite old) Test-VM for BiT. It is still running with Ubuntu 16.04 which comes with Python 3.5.2
Commit 8e0397f791803f2b0f3e5e2e119fde49807b59b1 comes with a new string format I didn't know before which breakes Python 3.5 support.

```
$ python3
Python 3.5.2 (default, Sep 16 2022, 21:16:04) 
[GCC 5.4.0 20160609] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> print(f"foo")
  File "<stdin>", line 1
    print(f"foo")
               ^
SyntaxError: invalid syntax
```

As Ubuntu 16.04 is not End of Life yet (April 2026) and we still build the PPA for 16.04 I'd prefere to keep compatibility. As far as I can see `f"foo {bar()}"` is just a replacement for `"foo {}".format(bar())`

aryoda:
I think we have also introduced new package dependencies that require at least Python 3.6 so there'd waiting more work for us to be backward compatible.

@buhtz Can you remember which ones?

I am not sure but LTE normally meant only 5 years but there seams to be another 5 years "Extended Security Maintenance (ESM)" period now:

https://ubuntu.com/16-04

Shall we really support such an old version with new updates (the old packaged BiT version should still work; we discussed a few weeks ago not to backport fixes to old versions due to the high maintenance efforts).

Any ideas/comments/proposals?

emtiu:
This was discussed [here on the mailing list](https://mail.python.org/archives/list/bit-dev@python.org/thread/OA64YEC3YYVUG5OOIZP26QMLMKMMKG6H/#OA64YEC3YYVUG5OOIZP26QMLMKMMKG6H) a couple of weeks ago. We didn't make a fixed policy decision, but the general consensus was that 3.6 (and therefore also 3.5) was too old to support in active development.

buhtz:
# The facts
## Python Software Foundation
Official details about Python versions and their support can be found at https://devguide.python.org/versions/.

- Python 3.5: Latest [release](https://www.python.org/downloads/release/python-3510/) was Sept. 2020. Reached end of life.
- Python 3.6: Latest [release](https://www.python.org/downloads/release/python-3615/) (and only for security reasons) was Sept. 2021 (3.6.15). It's latest _regular_ bugfix release (3.6.8) was Dec. 2018. So official support was droped 4 years ago.
[...]

PR #1438, opened 2023-05-16
build(deps): Drop support Python 3.7 and older

buhtz:
- Build requirement Python 3.8 or newer. **But** final decision about the version is pending. See #1358
- Fixed `pytest` support for `make test`.

### TODO

- [ ] Add minimal Python version to README and CONTRIBUTE docu.
- [ ] Adapt TravisCI configuration.


aryoda:
Looks good (but I didn't configure & build to test it).

Travis CI will now fail with Python 3.6, we should update the build matrix too (remove Python 3.6 as build target)...

aryoda (approved):
Python 3.6 must be removed from travis.yml to prevent permanent build errors

emtiu (approved):

PR #1439, opened 2023-05-22
fix: FAQ Url

buhtz:
The manpage and the Help menu in the GUI where pointing to the old FAQ in the Wiki. Now they point to the latest version of `FAQ.md` in the repo.

Beside this PR I also fixed the Wiki that way that users of older BIT versions are redirected to the correct FAQ file. The FAQ page from Wiki is removed and now the users do see the landing page of the Wiki which points to the FAQ.md file.

@Germar One question to you. Do I see it right that the manpage `backintime.1` is written manually and there is not autogeneration/script involved?

### DIFF
--- a/common/man/C/backintime.1
+++ b/common/man/C/backintime.1
@@ -145,7 +145,7 @@ remote host (via interactive ssh session) you are ready to go.
 .PP
 If you have questions on how to install and configure the Optware please refer
 to the community of your device. You can also take a look on Back In Time FAQ on
-GitHub https://github.com/bit-team/backintime/wiki/FAQ
+GitHub https://github.com/bit-team/backintime/blob/-/FAQ.md
 .PP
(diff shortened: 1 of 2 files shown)


aryoda:
Looks good, can be merge.

> https://github.com/bit-team/backintime/blob/-/FAQ.md

BTW: What does the "-" in the URL mean? It seems to be resolved to the "dev" branch. Is this a "github" convention?

buhtz:
> > `https://github.com/bit-team/backintime/blob/-/FAQ.md`
> 
> BTW: What does the "-" in the URL mean? It seems to be resolved to the "dev" branch. Is this a "github" convention?

This resolves to the default branch no matter what name it is. To my knowledge it is specific to Microsoft GitHub. I wasn't able to do the same with Gitea/Forgejo.

PR #1441, opened 2023-05-26
Use updateversion.sh script to update version in documentation (Sphinx configuration)

buhtz:
The `updateversion.sh` script is enhanced and now updates the sphinx configuration file to use the correct version number.

Using such scripts was usual in the early days of Python because there wasn't a good solution for this. As I wrote in the scripts docstring this all can be removed in the future when BIT migrates to Python Packaging standards. For example Sphinx then won't read a version from a file but from `backintime` packages meta data after importing it **without** manipulating the `sys.path`.

**EDIT**: As you can see in the commit history of this PR I reset `$MAINTANER` back to Germar where it was before. That variable is used only for Debian (PPA) related packaging files and not related to upstream. And Germar is maintainer of the PPA.


updateversion.sh
```diff
@@ -28,13 +37,15 @@
...
-    # TODO Decide on a maintainer name and email address
-    MAINTAINER="BIT Team <dan@le-web.org>"
-fi
+MAINTAINER="Germar Reitze <germar.reitze@gmail.com>"
+# MAINTAINER="BIT Team <dan@le-web.org>"
+# MAINTAINER="BIT Team <bit-dev@python.org>"
+
+update_sphinx_config () {
+  echo "Update '$1'"
+  sed -e "s/^\(\s*\)version = '.*'$/\1version = '$VERSION'/" \
```
aryoda, line 46: woo, regexp crack at work :-)

Would it make sense to add a comment with an example text what the first `sed` part shall match (to be replaced)?

I wasn't expecting `\(` and `\)` which AFAIR match the opening/closing parenthesis (since they are escaped)...
This looks wrong.

Furthermore `\s` matches any whitespace AFAIR and `\1` shall keep this first match group in the replacement I guess so I am quite confused what happens here and if it is as intended...

Are you trying to match the indentation of the `version` keyword string?

aryoda: I have just found out that the parentheses in sed requires escaping to the regexp looks good (to match optional leading white space) and it also looks the same way in other functions of the script (which obviously **did** work in the past :-)

https://stackoverflow.com/questions/11650940/sed-how-to-do-regex-groups-using-sed

```
update_man_page () {
  echo "Update '$1'"
  sed -e "s/\.TH\(.*\)\"version\([^\"]*\)\"\(.*\)$/.TH\1\"version $VERSION\"\3/" \
      -i $1
}
```

aryoda (approved):

updateversion.sh
buhtz: Yeah, I was sloopy. I just copy & paste the regex from the other parts of the script.

Usually I "test" regex via regex101.com .

PR #1448, opened 2023-06-15
fix: Ease translation strings

buhtz:
Now translatable strings are simplified and easier to understand for non-technical translators using our platform.
On the other side all translations now need a rework. But IMHO this is OK because there is no near release planned.

- "Cryptic" (format) characters and escape sequences are removed from the strings where ever possible. E.g. `'You don\'t can'` is now `"You don't can"`
- Strings are generalized where every possible to reduce the amount of translatable strings. E.g. `_("Profile:")` now is `_("Profile") + ":"`
- The context of a strings is now IMHO easier to understand. E.g. `_("Can not load %s")` now is `_("Can not load {path}").format(path=self.fancyvar)`
- Some translations are totally removed; e.g. Exceptions don't need to be translated because they are not intended to be read by users (if we do a good job), and crypto-alogrithm-names like SSH256 don't need a translation.

After this PR I can do a fresh transfer between the repo and weblate to see how many strings are left. Before this PR we do have 354 translatable strings.

I will do some more live tests on a virtual machine because this type of modification is not covered by unittests.

There is a high risk of bugs introduced. But they should be of minor type; e.g. typos or problems with parenthesis.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -9,6 +9,7 @@ Version 1.3.4-dev (development of upcoming release)
 * Feature: Exclude /swapfile by default (#1053)
 * Documentation: Removed outdated docbook (#1345).
 * Build: Introduced .readthedocs.yaml as asked by ReadTheDocs.org (#1443).
+* Translation: Strings to translate now easier to understand for translators (#1448).
 
 Version 1.3.3 (2023-01-04)
 * New feature: Command line argument "--diagnostics" to show helpful info for better issue support (#1100)
--- a/common/config.py
+++ b/common/config.py
@@ -95,13 +95,13 @@ class Config(configfile.ConfigFileWithProfiles):
                 NONE: _('Disabled'),
                 AT_EVERY_BOOT: _('At every boot/reboot'),
                 _5_MIN: _('Every 5 minutes'),
-                _10_MIN: _('Every 10 minutes'),
-                _30_MIN: _('Every 30 minutes'),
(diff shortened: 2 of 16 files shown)


aryoda:
Great work towards better internationalization!

> There is a high risk of bugs introduced. But they should be of minor type; e.g. typos or problems with parenthesis.

I think syntax errors are the biggest risk (esp. since they only occur when the LOC is executed so if there is no unit test or thorough manual testing BiT may stop with an RTE.

I will try to review the code with that in mind (before merging it) to reduce this risk.

**Edit:** It is impossible to review so many changes while staying fully concentrated ;-) I give up and would say let's take the risk...

buhtz:
> **Edit:** It is impossible to review so many changes while staying fully concentrated ;-) I give up and would say let's take the risk...

Yes, it is the same to me.

In my other (more PEP8 conform) projects I do use `pycodestyle` (state of the art PEP8 linter) and `ruff` (Rust-written linter combining multiple rule sets) as unittests. Others might use pre-commit hooks or such fancy things but I do it in unittests.
With BIT this isn't possible because pycodestyle gives us 3701 warnings and errors and ruff 787.

But I ran both of them manually, filtered out some minor errors and identify problems related to my PR. As you can see in the latest commit I found two syntax problems I could fix.

I also found two other issues not related to my PR and opened extra issues for that.

PR #1452, opened 2023-06-19
Translation: Updated pot and po files

buhtz:
Updated message template (pot) and language files (po) because of PR #1448.

Interesting to know:
Strings that do not exist in the python sources are removed from `pot` file.
But in the `po` language files that strings are not removed but just out commented.


buhtz:
Now I updated our weblate project. And I'm a bit confused.

Weblate don't show me a reduced number of strings but an increased. It increased from 354 to 437.

I checked the number of strings in the pot file in our repo (via `cat common/po/messages.pot | grep msgid | wc -l`) which gives me 321 strings. This is more realistic to me.

I also count the number of strings in the language files (via `cat common/po/de.po | grep "^msgid" | wc -l`) which also gets me 321 strings.

So I think it is fine. I'll do further research about why weblate react that way.

buhtz:
I fixed it and also have an idea why it happend. Complex but boils down to the fact that our weblate project is not connected to our Microsoft GitHub repo. Weblate can pull from our repo but is not able to push/commit/PR something.

I will monitor that when transfering between GitHub and Weblate the next time. Then I will update our developers docu about it.

I think the point is that the usual "Update" (or "Pull") do not work as expected. We have to "Reset" the weblate project everytime we updated the po files.

image

I count 321 strings on shell in our upstream repo but also in the cloned weblate repo. But Weblate frontend shows me 319. 😄

PR #1456, opened 2023-06-20
Update Turkish translation

ghost:


buhtz:
Dear Sabril,
thank you very much for your contribution and answering my request to translate.

The problem is that you modified the po file directly. As I wrote in my email we have a [translation platform](https://translate.codeberg.org/languages/tr/backintime/) for this. Direct editing the po files do create interference.

I will keep this PR open but in Draft state.
I will try to merge your po-file next time when updating all po-files from our translation platform.

So your work won't be lost!

Please let me know if you wan't to be named in the [TRANSLATIONS](https://github.com/bit-team/backintime/blob/dev/TRANSLATIONS) file? This names appear in the about dialog for example.

ghost:
Thanks, uploaded to the https://translate.codeberg.org/translate/backintime/common/tr/

PR #1458, opened 2023-06-22
fix: Missing return value when smartremove

### LINKED ISSUE #1392: function smartRemoveList missing return value
That was [originally reported](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=973760) at Debian Bug Tracker on Nov. 2020.
Nothing serious.

> Package: backintime-common
> Version: 1.2.1-2
> Severity: normal
> 
> Dear Maintainer,
> 
> Happened to be borwsing the class Snapshots in function smartRemoveList 
> 
>        if len(snapshots) <= 1:
>             logger.debug("There is only one snapshots, so keep it", self)
>             return
> 
> should maybe return the empty list

Related code location
https://github.com/bit-team/backintime/blob/e36d8dc85dd0181a38e174efc6d1f774be99ff4e/common/snapshots.py#L1350-L1354

buhtz:
The method computing a list of snapshots that are ready to get removed now returns an empty list if it can't find a snapshot to remove. Minor PEP8 formatting.

Fix #1392 

I checked involved code about possible bad implications of that fix.


Germar:
I would like to suggest, to splitt future formating changes into separate PRs. The mixture of formating and functional changes makes it hard to comprehend.

common/snapshots.py
```diff
@@ -1382,76 +1382,86 @@ def smartRemoveList(self,
...
         """
         snapshots = listSnapshots(self.config)
-        logger.debug("Considered: %s" %snapshots, self)
+        logger.debug('Considered: {snapshots}', self)
+
         if len(snapshots) <= 1:
-            logger.debug("There is only one snapshots, so keep it", self)
-            return
+            logger.debug('There is only one snapshots, so keep it', self)
+            return []
```
buhtz, line 1389: That is the core of the modification.
The rest is PEPificaton.

buhtz:
> splitt future formating changes into separate PRs. The mixture of formating and functional changes makes it hard to comprehend.

I totally agree. We also discussed and consented this in the team in the past. Because of that I always try to follow this standard but it isn't always possible in the wild because of interfering needs. In this specific PR I had to review some parts of the code to decide about the impact of the modification. To better understand the code and not missing something it need to be formatted well. I only modified two methods here.

Germar:
Maybe you could run that formating script for all `.py` files at once and commit those changes

buhtz:
Hello Germar,

thanks for your reply.

> Maybe you could run that formatting script for all `.py` files at once and commit those changes

I never use auto-formatters like `black` or something else. I don't agree with the concept behind them and it is definitely to risky. I only use "passive" linter.

But, as an anecdote, I tested `black` and used BIT as a test case. Black does break the code and introducing bugs. It is not invented auto-fixing py-files having that much PEP8 problems. That is why I state in our [CONTRIBUTE.md](https://github.com/bit-team/backintime/blob/dev/CONTRIBUTING.md) being careful with black & co and be transparent when  using it when opening an PR.

It is to risky fix all PEP8 problems at once. It is IMHO better doing it on the fly while "becoming warm" with the codebase and learning how and why it works. Which of course violate the rule to create atomar PRs. So I try to find a why between.

Even doing it manually is risky. Every modification of code keeps the risk introducing bugs. For example #1461 fixing PEP8 problems in two quite small files. I had to do this very focused and with patient. Work like this took me some hours. But the benefit is code easier to understand which increase its maintainability and also attracts future contributors.

PR #1459, opened 2023-06-22
fix: Warnings by Sphinx

buhtz:
Fixed two simple warnings in Sphinx's doc build. I encountered them while using `-W` (warnings to errors) flag in the build process.


buhtz:
If someone noticed the Sphinx build on ReadTheDocs was broken in the last days.
The major reason is a curios restriction for projects created before 20th October 2020.
The problem now is [fixed](https://github.com/readthedocs/readthedocs.org/issues/10464) exclusively for our project.

PR #1462, opened 2023-06-23
trans: High quality review of source strings

buhtz:
A qualified linguist (Justin B Rye) from Debian mailing list [debian-l10n-english](https://lists.debian.org/debian-l10n-english) reviewed the English strings in our code. He provided patches via Email.

**EDIT**: Learned something about git. 😄 That is why there are so many extra commits. Now it is in the state it has to be.

## Diff with comments
The reviewer provided comments directly in his diff files. Here they are.

```
  diff --git a/common/backintime.py b/common/backintime.py
  index 346ca3b6..8f2ba433 100644
  --- a/common/backintime.py
  +++ b/common/backintime.py
  @@ -225,7 +225,7 @@ def createParsers(app_name = 'backintime'):
       nargs = 0
       aliases = [(command, nargs), ('b', nargs)]
       description = 'Take a new snapshot. Ignore if the profile ' +\
  -                  'is not scheduled or if the machine runs on battery.'
  +                  'is not scheduled or if the machine is running on battery.'
       backupCP =             subparsers.add_parser(command,
                                                    parents = [rsyncArgsParser],
                                                    epilog = epilogCommon,

Simple past "runs" here would mean "if the machine always runs on
battery" or perhaps "is capable of running on battery".

  @@ -263,7 +263,7 @@ def createParsers(app_name = 'backintime'):
                                                    action = 'store',
                                                    default = 40,
                                                    nargs = '?',
  -                                                 help = 'File size used to for benchmark.')
  +                                                 help = 'File size used for benchmark.')
   
       command = 'check-config'
       description = 'Check the profiles configuration and install crontab entries.'

Typo.

  @@ -400,17 +400,17 @@ def createParsers(app_name = 'backintime'):
   
       backupGroup.add_argument                    ('--no-local-backup',
                                                    action = 'store_true',
  -                                                 help = 'Temporary disable creation of backup files before changing local files. ' +\
  -                                                 'This can be switched of permanently in Settings, too.')
  +                                                 help = 'Temporarily disable creation of backup files before changing local files. ' +\
  +                                                 'This can be switched off permanently in Settings, too.')

A couple more typos.
   
       restoreCP.add_argument                      ('--only-new',
                                                    action = 'store_true',
  -                                                 help = 'Only restore files which does not exist or are newer than ' +\
  +                                                 help = 'Only restore files which do not exist or are newer than ' +\
                                                           'those in destination. Using "rsync --update" option.')

Plural subject agreement fix.

   
       command = 'shutdown'
       nargs = 0
  -    description = 'Shutdown the computer after the snapshot is done.'
  +    description = 'Shut down the computer after the snapshot is done.'
       shutdownCP =           subparsers.add_parser(command,
                                                    epilog = epilogCommon,
                                                    help = description,

One of the cases where the verb is two words (proof: you can "shut the
computer down").  You can get away with this when you mean "running
/usr/sbin/shutdown", but then it doesn't get a direct object.

  @@ -431,7 +431,7 @@ def createParsers(app_name = 'backintime'):
       command = 'snapshots-list'
       nargs = 0
       aliases.append((command, nargs))
  -    description = 'Show a list of snapshots IDs.'
  +    description = 'Show a list of snapshot IDs.'
       snapshotsListCP =      subparsers.add_parser(command,
                                                    parents = [snapshotPathParser],
                                                    epilog = epilogCommon,

If they're IDs for snapshots, they're snapshot IDs - when you pile up
nouns like this they tend to lose their plural endings.

  @@ -443,7 +443,7 @@ def createParsers(app_name = 'backintime'):
       command = 'snapshots-list-path'
       nargs = 0
       aliases.append((command, nargs))
  -    description = "Show the path's to snapshots."
  +    description = "Show the paths to snapshots."
       snapshotsListPathCP =  subparsers.add_parser(command,
                                                    parents = [snapshotPathParser],
                                                    epilog = epilogCommon,
Plain plural nouns don't need an apostrophe.

  @@ -514,7 +514,7 @@ def startApp(app_name = 'backintime'):
       if tools.usingSudo() and os.getenv('BIT_SUDO_WARNING_PRINTED', 'false') == 'false':
           os.putenv('BIT_SUDO_WARNING_PRINTED', 'true')
           logger.warning("It looks like you're using 'sudo' to start %(app)s. "
  -                       "This will cause some troubles. Please use either 'sudo -i %(app_name)s' "
  +                       "This will cause some trouble. Please use either 'sudo -i %(app_name)s' "
                          "or 'pkexec %(app_name)s'."
                          %{'app_name': app_name, 'app': config.Config.APP_NAME})

"Trouble" is usually an uncountable noun.
   
  diff --git a/common/cli.py b/common/cli.py
  index 8c45e2a9..f1c3c1da 100644
  --- a/common/cli.py
  +++ b/common/cli.py
  @@ -46,7 +46,7 @@ def remove(cfg, snapshot_ids = None, force = None):
       sids = [selectSnapshot(snapshotsList, cfg, sid, 'SnapshotID to remove') for sid in snapshot_ids]
   
       if not force:
  -        print('Do you really want to remove this snapshots?')
  +        print('Do you really want to remove these snapshots?')
           [print(sid.displayName) for sid in sids]
           if not 'yes' == input('(no/yes): '):
               return
  diff --git a/common/config.py b/common/config.py
  index 0ffffb57..fab23e13 100644

"This" and "that" have special plural forms.  I assume you're sure the
snapshots are plural.

  --- a/common/config.py
  +++ b/common/config.py
  @@ -105,7 +105,7 @@ class Config(configfile.ConfigFileWithProfiles):
                   CUSTOM_HOUR: _('Custom Hours'),
                   DAY: _('Every Day'),
                   REPEATEDLY: _('Repeatedly (anacron)'),
  -                UDEV: _('When drive get connected (udev)'),
  +                UDEV: _('When drive gets connected (udev)'),
                   WEEK: _('Every Week'),
                   MONTH: _('Every Month'),
                   YEAR: _('Every Year')

Number agreement fix.

  @@ -355,7 +355,7 @@ class Config(configfile.ConfigFileWithProfiles):
                   self.notifyError(
                       '{}\n{}'.format(
                           _('Profile: "{name}"').format(name=profile_name),
  -                        _('You must select at least one folder to backup!')
  +                        _('You must select at least one folder to back up!')
                       )
                   )
   
Again, a one-word noun ("the backup") but a two-word verb ("to back
up").

  @@ -470,7 +470,7 @@ class Config(configfile.ConfigFileWithProfiles):
                   try:
                       os.chmod(p, 0o777)
                   except PermissionError as e:
  -                    msg = "Failed to change permissions world writeable for '{}': {}"
  +                    msg = "Failed to set permissions world-writable for '{}': {}"
                       logger.warning(msg.format(p, str(e)), self)
   
           # Test filesystem

Elsewhere in these messages you use the more conventional spelling
"writable".  While I'm fixing that, "change" is the wrong word (it
doesn't convey the fact that you're changing it *to* world-writable).
"Make" might also work, but "set" fits better with "for $target".

  @@ -487,7 +487,7 @@ class Config(configfile.ConfigFileWithProfiles):
   
           elif fs == 'cifs' and not self.copyLinks():
               self.notifyError(_(
  -                'Destination filesystem for {path} is a SMB mounted share. '
  +                'Destination filesystem for {path} is an SMB-mounted share. '
                   'Please make sure the remote SMB server supports symlinks or '
                   'activate {copyLinks} in {expertOptions}.')
                   .format(path=value,

"Ess-em-bee" begins with a vowel sound.  Then while I'm fixing that,
this isn't a mounted share that's SMB, it's a share that's
SMB-mounted, with a hyphen.

  @@ -496,7 +496,7 @@ class Config(configfile.ConfigFileWithProfiles):
   
           elif fs == 'fuse.sshfs' and mode not in ('ssh', 'ssh_encfs'):
               self.notifyError(_(
  -                "Destination filesystem for {path} is a sshfs mounted share."
  +                "Destination filesystem for {path} is an sshfs-mounted share."
                   " sshfs doesn't support hard-links. "
                   "Please use mode 'SSH' instead.")
                   .format(path=value))

As above.

  @@ -635,7 +635,7 @@ class Config(configfile.ConfigFileWithProfiles):
           #?0 = unlimited;0, >700
           value = self.profileIntValue('snapshots.ssh.max_arg_length', 0, profile_id)
           if value and value < 700:
  -            raise ValueError('SSH max arg length %s is to low to run commands' % value)
  +            raise ValueError('SSH max arg length %s is too low to run commands' % value)
           return value
   
       def setSshMaxArgLength(self, value, profile_id = None):

An easy typo to miss.

  @@ -1250,7 +1250,7 @@ class Config(configfile.ConfigFileWithProfiles):
           if '--old-args' in val:
               logger.warning(
                   'Found rsync flag "--old-args". That flag will be removed '
  -                'from the options because it does conflict with '
  +                'from the options because it conflicts with '
                   'the flag "-s" (also known as "--secluded-args" or '
                   '"--protected-args") which is used by Back In Time to force '
                   'the "new form of argument protection" in rsync.'

You'd only say "does conflict" if you were arguing with somebody who
had just said that it didn't.  (Does this message need re-wrapping?)

  @@ -1610,7 +1610,7 @@ class Config(configfile.ConfigFileWithProfiles):
               # entries if there is no automatic entry.
               newCrontab.append(self.SYSTEM_ENTRY_MESSAGE)
               newCrontab.append("#Please don't delete these two lines, or all custom backintime "
  -                              "entries are going to be deleted next time you call the gui options!")
  +                              "entries will be deleted next time you call the gui options!")
           return newCrontab
   
       def cronLine(self, profile_id):

This is a context where "going to" has too much implication of being
determined by external forces.

  diff --git a/common/mount.py b/common/mount.py
  index 819faba4..25e52fc3 100644
  --- a/common/mount.py
  +++ b/common/mount.py
  @@ -425,7 +425,7 @@ class MountControl(object):
                       raise HashCollision(
                           f'Hash collision occurred in hash_id {self.hash_id}. '
                           'Incrementing global value hash_collision and '
  -                        'try again.')
  +                        'trying again.')
   
                   logger.info('Mountpoint {} is already mounted'
                               .format(self.currentMountpoint),

If this is something the user needs to do then it should be "increment
it and try again"; if it's reporting what backintime is doing then
it's "incrementing it and trying again".

  diff --git a/common/snapshots.py b/common/snapshots.py
  index a9b99c16..24a5e181 100644
  --- a/common/snapshots.py
  +++ b/common/snapshots.py
  @@ -120,7 +120,7 @@ class Snapshots:
           try:
               mid = int(items[0])
           except Exception as e:
  -            logger.debug('Failed extract message ID from %s: %s'
  +            logger.debug('Failed to extract message ID from %s: %s'
                            %(items[0], str(e)),
                            self)
   
Missing word.

  @@ -746,7 +746,7 @@ restore is done. The pid of the already running restore is in %s.  Maybe delete
                       if not self.config.canBackup(profile_id):
                           if self.config.PLUGIN_MANAGER.hasGuiPlugins and self.config.notify():
                               self.setTakeSnapshotMessage(1,
  -                                    _('Can\'t find snapshots folder.\nIf it is on a removable drive please plug it.') +
  +                                    _('Can\'t find snapshots folder.\nIf it is on a removable drive please plug it in.') +
                                       '\n' +
                                       gettext.ngettext('Waiting %s second.', 'Waiting %s seconds.', 30) % 30,
                                       30)

Plugging a thing either means blocking it up or appearing on TV saying
how good it is, neither of which would help much in this case; what
they need to do is plug it in.  (I wonder why people never write it as
"to plugin"?)

  @@ -962,7 +962,7 @@ restore is done. The pid of the already running restore is in %s.  Maybe delete
                   # But we need a real ERROR here.
                   if rc != 0:
                       logger.error(
  -                        f'Backup the config in "{self.config.snapshotsMode()}"'
  +                        f'Backing up the config in "{self.config.snapshotsMode()}"'
                           f' mode failed! The return code was {rc} and the'
                           f' command was {cmd}. Also see the previous '
                           'WARNING message for a more details.', parent=self)

If it's trying to report that "doing something failed", it needs to be
the verb, with "-ing" attached.

  @@ -1384,7 +1384,7 @@ restore is done. The pid of the already running restore is in %s.  Maybe delete
           snapshots = listSnapshots(self.config)
           logger.debug("Considered: %s" %snapshots, self)
           if len(snapshots) <= 1:
  -            logger.debug("There is only one snapshots, so keep it", self)
  +            logger.debug("There is only one snapshot, so keep it", self)
               return
   
           if now_full is None:

One is the only number that doesn't go with a plural inflection.

  @@ -2673,12 +2673,12 @@ class NewSnapshot(GenericNonSnapshot):
                   with open(flag, 'wt') as f:
                       pass
               except Exception as e:
  -                logger.error("Failed to set 'save_to_continue' flag: %s" %str(e))
  +                logger.error("Failed to set 'save_to_continue' flag: %s" %str(e)) # should be "safe", throughout
           elif os.path.exists(flag):
               try:
                   os.remove(flag)
               except Exception as e:
  -                logger.error("Failed to remove 'save_to_continue' flag: %s" %str(e))
  +                logger.error("Failed to remove 'save_to_continue' flag: %s" %str(e)) # should be "safe", throughout
   
       @property
       def hasChanges(self):

I haven't dared fix this one.  You've got a spelling error in the
code, which shouldn't matter except that you then expose the name of
the flag to users (um, except that in fact the code doesn't use that
name).  What do translations do with this?

  @@ -2783,7 +2783,7 @@ def iterSnapshots(cfg, includeNewSnapshot = False):
                   yield sid
           except Exception as e:
               if not isinstance(e, LastSnapshotSymlink):
  -                logger.debug("'{}' is no snapshot ID: {}".format(item, str(e)))
  +                logger.debug("'{}' is not a snapshot ID: {}".format(item, str(e)))
   
   
   def listSnapshots(cfg, includeNewSnapshot = False, reverse = True):

Unless you mean this to be a dramatic "X is no TRUE snapshot!"...

  diff --git a/common/sshtools.py b/common/sshtools.py
  index 49058085..216d3e56 100644
  --- a/common/sshtools.py
  +++ b/common/sshtools.py
  @@ -740,7 +740,7 @@ class SSH(MountControl):
                   raise MountException(
                       "Checking commands on remote host didn't return any "
                       "output. We already checked the maximum argument length "
  -                    "but it seem like there is an other problem")
  +                    "but it seems like there is another problem")
   
               logger.warning(
                   'Looks like the command was to long for remote SSHd. '

Minor English glitches.

  @@ -897,7 +897,7 @@ class SSH(MountControl):
                   # Argument list too long
                   if e.errno == 7:
                       logger.debug(
  -                        'Argument list too log (Python exception)', self)
  +                        'Argument list too long (Python exception)', self)
   
                       return maxArg()

Debug message too short.

  @@ -1013,7 +1013,7 @@ def sshKeyGen(keyfile):
       if proc.returncode:
           logger.error('Failed to create a new ssh-key: {}'.format(err))
       else:
  -        logger.info('Successfully create new ssh-key "{}"'.format(keyfile))
  +        logger.info('Successfully created new ssh-key "{}"'.format(keyfile))
   
       return not proc.returncode
   
I'm assuming this is the opposite of the error, reporting something
that happened.

  index 3c954fad..db043561 100644
  --- a/common/tools.py
  +++ b/common/tools.py
  @@ -850,7 +850,7 @@ def keyringSupported():
           logger.debug('No keyring due to import error.')
           return False
   
  -    keyring_config_file_folder = "Unkown"
  +    keyring_config_file_folder = "Unknown"
       try:
           keyring_config_file_folder = keyring.util.platform_.config_root()
       except:

I hope this doesn't break everybody's folder setups.

  @@ -1418,7 +1418,7 @@ def readCrontab():
               crontab = [x.strip() for x in out.strip('\n').split('\n')]
               if crontab == ['']:  # Fixes issue #1181 (line count of empty crontab was 1 instead of 0)
                   crontab = []
  -            logger.debug('Read %s lines from users crontab'
  +            logger.debug('Read %s lines from user crontab'
                            %len(crontab))
               return crontab
   
  @@ -1455,7 +1455,7 @@ def writeCrontab(lines):
                        %(proc.returncode, err))
           return False
       else:
  -        logger.debug('Wrote %s lines to users crontab'
  +        logger.debug('Wrote %s lines to user crontab'
                        %len(lines))
           return True
   
In both these cases, it's easier to treat them as stacked uninflected
nouns instead of a possessive phrase.

  @@ -1613,13 +1613,13 @@ class UniquenessSet:
               size,inode  = dum.st_size, dum.st_ino
               # is it a hlink ?
               if (size, inode) in self._size_inode:
  -                logger.debug("[deep test] : skip, it's a duplicate (size, inode)", self)
  +                logger.debug("[deep test]: skip, it's a duplicate (size, inode)", self)
                   return False
               self._size_inode.add((size,inode))
               if size not in self._uniq_dict:
                   # first item of that size
                   unique_key = size
  -                logger.debug("[deep test] : store current size ?", self)
  +                logger.debug("[deep test]: store current size?", self)
               else:
                   prev = self._uniq_dict[size]
                   if prev:
  @@ -1627,16 +1627,16 @@ class UniquenessSet:
                       md5sum_prev = md5sum(prev)
                       self._uniq_dict[size] = None
                       self._uniq_dict[md5sum_prev] = prev
  -                    logger.debug("[deep test] : size duplicate, remove the size, store prev md5sum", self)
  +                    logger.debug("[deep test]: size duplicate, remove the size, store prev md5sum", self)
                   unique_key = md5sum(path)
  -                logger.debug("[deep test] : store current md5sum ?", self)
  +                logger.debug("[deep test]: store current md5sum?", self)
           else:
               # store a tuple of (size, modification time)
               obj  = os.stat(path)
               unique_key = (obj.st_size, int(obj.st_mtime))
           # store if not already present, then return True
           if unique_key not in self._uniq_dict:
  -            logger.debug(" >> ok, store !", self)
  +            logger.debug(" >> ok, store!", self)
               self._uniq_dict[unique_key] = path
               return True
           logger.debug(" >> skip (it's a duplicate)", self)

All through the above I'm being pedantic about the rules for spaces
before "!/?/:/;" - but you may notice I ended up not bothering to fix
the ones with a space before "!!!", since after all that can take
cover behind the fact that multiple exclamation marks are one of those
things that most style guides don't approve of in the first place.

  @@ -2347,7 +2347,7 @@ class Daemon:
           """
           # Check for a pidfile to see if the daemon already runs
           if self.pidfile and not self.appInstance.check():
  -            message = "pidfile %s already exist. Daemon already running?\n"
  +            message = "pidfile %s already exists. Daemon already running?\n"
               logger.error(message % self.pidfile, self)
               sys.exit(1)
   
Singular agreement.  I notice you're also a bit inconsistent about
whether it's "pidfile" or "PID file", but I've left that for now.

  diff --git a/qt/app.py b/qt/app.py
  index 9d12d3e9..33e54363 100644
  --- a/qt/app.py
  +++ b/qt/app.py
  @@ -145,7 +145,7 @@ class MainWindow(QMainWindow):
           self.mainToolbar.addSeparator()
   
           self.btnShutdown = self.mainToolbar.addAction(icon.SHUTDOWN, _('Shutdown'))
  -        self.btnShutdown.setToolTip(_('Shutdown system after snapshot has finished.'))
  +        self.btnShutdown.setToolTip(_('Shut down system after snapshot has finished.'))
           self.btnShutdown.setCheckable(True)
           self.shutdown = tools.ShutDown()
           self.btnShutdown.setEnabled(self.shutdown.canShutdown())

Another case of "one word noun, two words verb".

  @@ -242,13 +242,13 @@ class MainWindow(QMainWindow):
           self.menuRestore.addSeparator()
           self.btnRestoreParent = self.menuRestore.addAction(icon.RESTORE, '')
           self.btnRestoreParent.setToolTip(_('Restore the currently shown '
  -                                           'folder and all its content to '
  +                                           'folder and all its contents to '
                                              'the original destination.'))
           self.btnRestoreParent.triggered.connect(self.restoreParent)
           self.btnRestoreParentTo = self.menuRestore.addAction(
               icon.RESTORE_TO, '')
           self.btnRestoreParentTo.setToolTip(_('Restore the currently shown '
  -                                             'folder and all its content '
  +                                             'folder and all its contents '
                                                'to a new destination.'))
           self.btnRestoreParentTo.triggered.connect(self.restoreParentTo)
           self.menuRestore.addSeparator()

Folders have plural contents, though they aren't made up of many
individual things each of which is a content.  There's also an
orthodox uncountable noun "content", but that means something
different.

  @@ -519,7 +519,7 @@ class MainWindow(QMainWindow):
           if not config.canBackup(profile_id):
               messagebox.critical(self, _(
                   "Can't find snapshots folder.\nIf it is on a removable "
  -                "drive please plug it and then press OK."))
  +                "drive please plug it in and then press OK."))
   
           self.filesViewProxyModel.layoutChanged.connect(self.dirListerCompleted)
   
You plug things with bath plugs; what you do with power plugs is plug
things in.

  @@ -555,7 +555,7 @@ class MainWindow(QMainWindow):
       def closeEvent(self, event):
           if self.shutdown.askBeforeQuit():
               msg = _('If you close this window Back In Time will not be able '
  -                    'to shutdown your system when the snapshot has finished.'
  +                    'to shut down your system when the snapshot has finished.'
                       '\nDo you really want to close?')
               if QMessageBox.Yes != messagebox.warningYesNo(self, msg):
                   return event.ignore()

The same old English grammar gotcha; then this next one has another
case of "backup/back up", but gets more complicated:

 @@ -1073,8 +1073,8 @@ class MainWindow(QMainWindow):

       def backupOnRestore(self):
           cb = QCheckBox(_(
  -            'Backup local files before overwriting or\nremoving with '
  -            'trailing {suffix}.').format(
  +            'Create backup copies with trailing {suffix} before\n'
  +            'overwriting or removing local files.').format(
                   suffx=self.snapshots.backupSuffix()))

           cb.setChecked(self.config.backupOnRestore())

It wasn't clear what "with trailing {suffix}" was doing in that
sentence.  You're not overwriting files with some other thing, or
selecting and overwriting the files with some feature, you're backing
files up by creating copies with new suffixed filenames - right?

  @@ -1131,7 +1131,7 @@ files that the receiver requests to be transferred.""")
                           'to the original destination and\n'
                           'delete files/folders which are '
                           'not in the snapshot.\n'
  -                        'This will delete files/folders which where '
  +                        'This will delete files/folders which were '
                           'excluded during taking the snapshot!\n'
                           'Be extremely careful!!!'))
           return {'widget': cb, 'retFunc': cb.isChecked, 'id': 'delete'}

A common typo.

  @@ -1139,11 +1139,11 @@ files that the receiver requests to be transferred.""")
       def confirmRestore(self, paths, restoreTo = None):
           if restoreTo:
               msg = '{}:'.format(
  -                _('Do you really want to restore this files(s)\ninto '
  +                _('Do you really want to restore the files(s)\ninto '
                     'new folder {path}').format(path=restoreTo))
           else:
               msg = '{}:'.format(
  -                _('Do you really want to restore this files(s)'))
  +                _('Do you really want to restore the files(s)'))
   
           confirm, opt = messagebox.warningYesNoOptions(self,
                                                         msg,

Avoid the issue of "this/these file(s)" by just using the definite
article.

  diff --git a/qt/settingsdialog.py b/qt/settingsdialog.py
  index f9f02844..d5d29a5c 100644
  --- a/qt/settingsdialog.py
  +++ b/qt/settingsdialog.py
  @@ -501,8 +501,8 @@ class SettingsDialog(QDialog):
   
           self.lblSshEncfsExcludeWarning = QLabel(
               _('<b>Warning:</b> Wildcards (\'foo*\', \'[fF]oo\', \'fo?\') '
  -              'will be ignored with mode \'SSH encrypted\'.\nOnly separate '
  -              'asterisk are allowed (\'foo/*\', \'foo/**/bar\')'), self)
  +              'will be ignored with mode \'SSH encrypted\'.\nOnly single or '
  +              'double asterisks are allowed (\'foo/*\', \'foo/**/bar\')'), self)
           self.lblSshEncfsExcludeWarning.setWordWrap(True)
           layout.addWidget(self.lblSshEncfsExcludeWarning)
   
Plural "asterisks", and I don't think you mean "separate" - you mean
"just * or **", so this is the best way I can find of saying it.

  @@ -566,9 +566,9 @@ class SettingsDialog(QDialog):
                 'new files\n'
                 'because for rsync this is a transfer option, not an '
                 'exclude option.\n'
  -              'So big files that has been backed up before will remain '
  +              'So big files that have been backed up before will remain '
                 'in snapshots\n'
  -              'even if they had changed.' % {'prefix': 'MiB'})
  +              'even if they have changed.' % {'prefix': 'MiB'})
           )
           hlayout.addWidget(self.cbExcludeBySize)
           self.spbExcludeBySize = QSpinBox(self)

A number agreement error and a tense problem: we're not talking about
big files that had already changed before the last time they were
backed up, we're talking about ones that have changed since then.

  @@ -719,7 +719,7 @@ class SettingsDialog(QDialog):
           self.cbGlobalFlock.setToolTip(
               _('Other snapshots will be blocked until the current snapshot '
                 'is done.\n'
  -              'This is a global option. So it will effect all profiles '
  +              'This is a global option. So it will affect all profiles '
                 'for this user.\n'
                 'But you need to activate this for all other users, too.')
           )

A different verb-versus-noun oddity!  It's noun "the effect" but verb
"to affect" (with rare exceptions: https://xkcd.com/326)

  @@ -747,7 +747,7 @@ class SettingsDialog(QDialog):
           layout.addWidget(self.cbUseChecksum)
   
           self.cbTakeSnapshotRegardlessOfChanges = QCheckBox(
  -            _('Take a new snapshot regardless of there were changes or not.'))
  +            _('Take a new snapshot regardless of whether there were changes or not.'))
           layout.addWidget(self.cbTakeSnapshotRegardlessOfChanges)
   
           # log level

In fact I think I'd say this with just "whether" and no "regardless
of", but you can't do the reverse.

  @@ -780,7 +780,7 @@ class SettingsDialog(QDialog):
   
           label = QLabel(
               _('Change these options only if you really know what '
  -              'you are doing !'),
  +              'you are doing!'),
               self)
           qttools.setFontBold(label)
           layout.addWidget(label)

The no-space-before-!?:; rule.

  @@ -849,7 +849,7 @@ class SettingsDialog(QDialog):
               self)
           self.cbRedirectStdoutInCron.setToolTip(
               'cron will automatically send an email with attached output '
  -            'of cronjobs if a MTA is installed.')
  +            'of cronjobs if an MTA is installed.')
           layout.addWidget(self.cbRedirectStdoutInCron)
   
           self.cbRedirectStderrInCron = QCheckBox(
  @@ -858,7 +858,7 @@ class SettingsDialog(QDialog):
               self)
           self.cbRedirectStderrInCron.setToolTip(
               'cron will automatically send an email with attached errors '
  -            'of cronjobs if a MTA is installed.')
  +            'of cronjobs if an MTA is installed.')
           layout.addWidget(self.cbRedirectStderrInCron)
   
           # bwlimit

Because "em-tee-ay" begins with a vowel.

  @@ -1019,7 +1019,7 @@ class SettingsDialog(QDialog):
                 'is not available, this could lead to some\n'
                 'weird errors.'))
           self.cbSshCheckCommands = QCheckBox(
  -            _('Check if remote host support all necessary commands'))
  +            _('Check if remote host supports all necessary commands'))
           self.cbSshCheckCommands.setToolTip(
               _('Warning: if disabled and the remote host\n'
                 'does not support all necessary commands,\n'

Number agreement.

  @@ -1853,7 +1853,7 @@ class SettingsDialog(QDialog):
                   question_msg = _(
                       '"{path}" is a symlink. The linked target will not be '
                         'backed up until you include it, too.\nWould you like '
  -                      'to include the symlinks target instead?'
  +                      'to include the symlink target instead?'
                   ).format(path=path)
   
                   if self.questionHandler(question_msg):
  @@ -1878,7 +1878,7 @@ class SettingsDialog(QDialog):
                   if self.questionHandler(
                       _('"%s" is a symlink. The linked target will not be '
                         'backed up until you include it, too.\nWould you like '
  -                      'to include the symlinks target instead?') % path):
  +                      'to include the symlink target instead?') % path):
                       path = os.path.realpath(path)
   
               path = self.config.preparePath(path)

I don't see how that "-s" got there, but it should go away.

  @@ -2114,7 +2114,7 @@ class RestoreConfigDialog(QDialog):
               "Please navigate to the snapshot from which you want to restore "
               "{appName}'s configuration. The path may look like:\n"
               "{samplePath}\n\nIf your snapshots are on a remote drive or if "
  -            "they are ncrypted you need to manually mount them first. "
  +            "they are encrypted you need to manually mount them first. "
               "If you use Mode SSH you also may need to set up public key "
               "login to the remote host{addFuse}.\n"
               "Take a look at 'man backintime'.")

Typo?  (Oh, a nice idiomatic split infinitive though!)

  diff --git a/qt/snapshotsdialog.py b/qt/snapshotsdialog.py
  index c62f8f0d..5d64fcc3 100644
  --- a/qt/snapshotsdialog.py
  +++ b/qt/snapshotsdialog.py
  @@ -376,7 +376,7 @@ class SnapshotsDialog(QDialog):
                           file=self.path, count=len(items))
   
           msg = '{}\n{}: {}'.format(
  -            msg, _('WARNING'), _('This can not be revoked!'))
  +            msg, _('WARNING'), _('This cannot be revoked!'))
   
           if QMessageBox.Yes == messagebox.warningYesNo(self, msg):
               for item in items:

"Can not" gets ambiguous ("you can not go" may mean "it's prohibited"
or "it's non-mandatory"), so stick to "cannot".
```


common/config.py
```diff
@@ -1626,7 +1626,7 @@ def createNewCrontab(self, oldCrontab):
             # entries if there is no automatic entry.
             newCrontab.append(self.SYSTEM_ENTRY_MESSAGE)
             newCrontab.append("#Please don't delete these two lines, or all custom backintime "
-                              "entries are going to be deleted next time you call the gui options!")
+                              "entries will be deleted next time you call the gui options!")
```
Germar, line 1634: Wow, this is a high quality review! Thank you!

BiT uses this phrase to test for own entries in `crontab`. We need to check if changing this phrase will break that test!

buhtz: Thanks for reviewing so carefully and bringing this up. That situation is also [mentioned in the FAQ](https://github.com/bit-team/backintime/blob/dev/FAQ.md#if-i-edit-my-crontab-and-add-additional-entries-will-that-be-a-problem-for-bit-as-long-as-i-dont-touch-its-entries-what-does-it-look-for-in-the-crontab-to-find-its-own-entries). After testing around and checking the code I'm quit sure there is no risk because `self.SYSTEM_ENTRY_MESSAGE` only is checked. If none of the existing snapshot profiles are scheduled the crontab has this two lines

    #Back In Time system entry, this will be edited by the gui:                                                                                                                           
    #Please don't delete these two lines, or all custom backintime entries are going to be deleted next time you call the gui options!

BIT only checks for the first of this lines which value is stored in a [module constant](https://github.com/bit-team/backintime/blob/72ced78a5ee3c7803e0cba16e11698af4d883535/common/config.py#L1540).

https://github.com/bit-team/backintime/blob/72ced78a5ee3c7803e0cba16e11698af4d883535/common/config.py#L1593C21-L1593C42

I added a user defined crontab line (`1 0 * * * backintime`). It was deleted only when the first of the comment lines are removed.

To prevent accidentally modifying that string we may should add a comment to the definition of `self.SYSTEM_ENTRY_MESSAGE`. On step more would be to create a "unit test" just checking the value of that string. Not elegant but would do it's job preventing accidents.

I also did some `git grep` searching but couldn't find any indication of more problems.

buhtz: I added a docstring incl. WARNING to SYSTEM_ENTRY_MESSAGE.

PR #1467, opened 2023-07-04
refactor: Introduce GNU gettext class-based API

buhtz:
This PR introduce the class-based API for GNU gettext ([docu](https://docs.python.org/3/library/gettext.html#class-based-api) & [example](https://docs.python.org/3/library/gettext.html#localizing-your-application)). It eliminates the need to `import gettext` and declaring `_()` in each module. Now it is done on a global level (in `builtins` namespace). The behavior of BIT do not change. It is just refactoring.

_Why we need this_: I want to introduce a feature to modify the language used by BIT. This puts more value to the translations because there are areas in the world where it is not unusual to use multiple languages on the same computer system independent from the locale that is setup in the OS. It also helps with manual GUI testing.

~There might be a problem with `qt/qtsystrayicon.py`. [I'm not sure how it is invoked or where it is used](https://mail.python.org/archives/list/bit-dev@python.org/message/V3TLXN3QXZTDMFQZV72SYXUTM2TL5NUK/). Depending on the answer the use of `gettext` need to be altered in that file, too.~

_Technical roadmap_: The language code will be stored via `config.Config`. That is why `tools.initiate_translation()` is done in the end of `config.Config.__init__()`. In the future the language code will be read from the config and handed over as an argument to `initiate_translation()`. IMHO it is not ideal doing this inside the `Config` class but it is a pragmatic solution without changing to much of the code(structure). Something similar will be done with `qt/qttools.py::translator()`.

### DIFF
--- a/common/config.py
+++ b/common/config.py
@@ -41,6 +41,14 @@ except ImportError:
     import getpass
     pwd = None
 
+# Workaround: Mostly relevant on TravisCI but not exclusivley.
+# While unittesting and without regular invocation of BIT the GNU gettext
+# class-based API isn't setup yet.
+try:
+    _('Foo')
+except NameError:
+    _ = lambda val: val
+
 import tools
 import configfile
 import logger
--- a/common/config.py
+++ b/common/config.py
@@ -32,7 +32,6 @@ Development notes:
 import os
 import sys
 import datetime
-import gettext
 import socket
 import random
 import shlex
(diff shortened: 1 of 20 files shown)


common/config.py
```diff
@@ -342,6 +344,8 @@ def __init__(self, config_path=None, data_path=None):
         self.inhibitCookie = None
         self.setupUdev = tools.SetupUdev()
 
+        tools.initiate_translation()
```
buhtz, line 347: Here happens the _magic_. That is the key of this PR.

buhtz:
I found the answer about `qt/qtsystrayicon.py`. It is invoked via a `Popen()` in `qt/plugins/qt4plugin.py`.
Translation works also well in there because `snapshots` is imported which itself imports `config`. The latter do the translation setup.

Issue #1468, opened 2023-07-06
Why use setThemeName() and why is it never reached?


buhtz:
I found a piece of code and don't understand what it does or why it is there. Maybe someone can light me up with it. I have some PR's on my todo list relating to this.

https://github.com/bit-team/backintime/blob/e44e9d64b972be2dab3aa0f94af34d44e80eabf9/qt/icon.py#L19-L26

It was introduced 9 years ago (and slightly modified since then) via c95b616.

The `setThemeName()` is never reached on my tests. What is the purpose and intended behavior here?
Can't exactly see how #1364 and #1306 is related to it.

buhtz:
After a cup of coffee...

Please correct me if my assumptions not correct here. Qt do use the currents system theme by default. But sometimes an icon is not available in that theme. In this case this code do check other (well known) themes (3 in this piece of code) if the icon is available there. If it is the case then this alternative theme is set for BIT no matter what the current systems theme is. Right?

Some thoughts about that
 - First of all I'm interested why icons not available in a theme? We don't use such exotic icons in BIT. Maybe something is wrong with the icon handling and not with the theme.
 - Why is `document-save` used in this piece of code. I would bet this is an "extrem standard icon" available in every icon theme ever used. It will never be unavailable in any theme.
 - Using an alternative theme is IMHO not a good behavior. It seems more appropriate to fall back to Qt's in build icons available via `QStyle` in that case. There are not much and they are not pretty. But they are there.

aryoda:
> Why use setThemeName() and why is it never reached?

This LOC **is** executed but only if none of the white-listed themes is installed (so the last white-listed theme "oxygen" is activated then even though it is not installed - this is a bug and I have fixed it in my upcoming PR)

buhtz:
Did you know about [Qt's in-build icons](https://www.pythonguis.com/faq/built-in-qicons-pyqt/)?

> this is a bug and I have fixed it in my upcoming PR)

Ah, good to know. Then I will wait with refactoring `qt/icon.py` to prevent locale namespace imports happening here:

```
$ git grep "import icon" **/*.py
qt/app.py:77:        import icon
qt/app.py:1502:        import icon
qt/logviewdialog.py:54:        import icon
qt/messagebox.py:31:    import icon
qt/qtsystrayicon.py:61:        import icon
qt/restoredialog.py:44:        import icon
qt/settingsdialog.py:84:        import icon
qt/settingsdialog.py:2097:        import icon
[...]

PR #1469, opened 2023-07-06
feat: Diagnostics about Qt theme

buhtz:
Improved "--diagnostics" with Qt Theme information.

Related to #1468 

Output (running with X):

image

Output in terminal (without X):

        "qt": {
            "Version": "PyQt 5.15.2 / Qt 5.15.2"
        }

Question by the way: Does someone know what `:/icons` mean a list of paths?

### DIFF
--- a/common/diagnostics.py
+++ b/common/diagnostics.py
@@ -201,7 +201,7 @@ def _get_qt_information():
 
     # Themes
     theme_info = {}
-    if 'DISPLAY' in os.environ:
+    if tools.checkXServer():
         qapp = PyQt5.QtWidgets.QApplication([])
         theme_info = {
             'Theme': PyQt5.QtGui.QIcon.themeName(),
--- a/common/diagnostics.py
+++ b/common/diagnostics.py
@@ -197,7 +197,7 @@ def _get_qt_information():
         import PyQt5.QtGui
         import PyQt5.QtWidgets
     except ImportError:
-        return '(Can not import PyQt5)'
+        return '(Cannot import PyQt5)'
 
     # Themes
     theme_info = {}
--- a/common/diagnostics.py
+++ b/common/diagnostics.py
@@ -130,16 +130,7 @@ def collect_diagnostics():
 
     result['python-setup']['sys.path'] = sys.path
 
-    # Qt
-    try:
-        import PyQt5.QtCore
-    except ImportError:
-        qt = '(Can not import PyQt5)'
-    else:
-        qt = 'PyQt {} / Qt {}'.format(PyQt5.QtCore.PYQT_VERSION_STR,
-                                      PyQt5.QtCore.QT_VERSION_STR)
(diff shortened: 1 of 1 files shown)


aryoda:
Very helpful, THX!

> Question by the way: Does someone know what `:/icons` mean a list of paths?

The leading colon does indicate an absolute path in Qt:

"Paths starting with a colon (:) are always considered absolute, as they denote a [QResource]"

https://doc.qt.io/qt-6/qdir.html

common/diagnostics.py
```diff
@@ -195,6 +186,38 @@ def collect_diagnostics():
...
+        import PyQt5.QtCore
+        import PyQt5.QtGui
+        import PyQt5.QtWidgets
+    except ImportError:
+        return '(Can not import PyQt5)'
+
+    # Themes
+    theme_info = {}
+    if 'DISPLAY' in os.environ:
+        qapp = PyQt5.QtWidgets.QApplication([])
```
aryoda, line 205: I am not sure if using the DISPLAY env var is working in case of Wayland. How about simply using a try-catch clause (which is missing here anyhow I think since establishing a `QApplication` instance may fail somehow (we had some related issues with that)...

buhtz: `try...except` was my first try. But Qt doesn't raise anything here but just crash the whole Python instance when `QApplication()` without running X. Any idea how I can take Wayland into account here?

What about `XDG_SESSION_TYPE`. Might that help? Its value should be `wayland` or `x11`, right?

buhtz: Or much more "dirty" just using an external python call and checking its result.

    python3 -c "from PyQt5.QtWidgets import QApplication; QApplication([])"

aryoda: > try...except was my first try. But Qt doesn't raise anything here but just crash the whole Python instance when QApplication() without running X

uh, very nasty :-(   I can remember now why I am stuck with my systrayicon plugin PR where I have the same challenge: To find out if Qt5 is installed **and** working.

This is currently done by spawning a `xdpyinfo` call and checking the return value:

https://github.com/bit-team/backintime/blob/e44e9d64b972be2dab3aa0f94af34d44e80eabf9/qt/plugins/qt4plugin.py#L45-L47

Perhaps we should stick to reuse this `tools.checkXServer()` call in your diagnostics code here too and fully check the Wayland compliance later. I think XWayland kicks in in case of Wayland and returns "X is available":

https://github.com/bit-team/backintime/blob/e44e9d64b972be2dab3aa0f94af34d44e80eabf9/common/tools.py#L474-L481

I think the final solution (in the future) must be independent of checking for X or Wayland or any other technology but "simply" ask Qt5/6 if a GUI can be shown... (by somehow catching `QApplication` exceptions without immediately crashing python)

aryoda (approved):

PR #1470, opened 2023-07-06
fix: Move MountControl template ("Dummy") into docstrings

### LINKED ISSUE #1451: dummytools.py: F821 Undefined name "parent"
While reviewing my own PR #1448 with linters (ruff & pycodestyle) I found a possible bug. Further investigation needed.
https://github.com/bit-team/backintime/blob/8fbec2d5e2d3c40af7cad5f5c6945ac1adf15f9e/common/dummytools.py#L51
Ruff says `common/dummytools.py:51:61: F821 Undefined name "parent"`.

 - What happens there?

buhtz:
The code from dummytools.py is moved into the module docstring of "mount.py". As discussed in #1451 the purpose of this code was to be a template for developers deriving from "mount.MountControl".

Resolves #1451

The HTML docu
image


buhtz:
Reopen because `common/dummytools.py` still exists but should be deleted.

PR #1472, opened 2023-07-13
feat!: Remove "fuse" user group handling and checking

### LINKED ISSUE #1450: mount.py: F821 Undefined name "grp" / Check fuse-group membership
While reviewing my own PR #1448 with linters (ruff & pycodestyle) I found a possible bug.
Ruff says `common/mount.py:608:36: F821 Undefined name "grp"` .
https://github.com/bit-team/backintime/blob/8fbec2d5e2d3c40af7cad5f5c6945ac1adf15f9e/common/mount.py#L591
It should IMHO throw a `NameError`. Further investigation needed.

 - What happens there?

buhtz:
The following behavior, introduced in 2013 via 26474f8, was removed: While installing (in configure script) and when mounting (see mount.py::MountControl.checkFuse()) the existence of the user group "fuse" was checked. The reasons and arguments described in #1450.

The exclamation mark (in "feat!:") is used even if it isn't a breaking change for real. The behavior shouldn't change because of the assumption that the code is not reached in any situation. But if the assumption is wrong it might be easier to find the relevant commit message.

Close #1450


aryoda (approved):
Looks reasonable and good, let's remove it and wait for new issues (if any!)

PR #1474, opened 2023-07-13
fix: Remove use of shell=True in update-language helper script

buhtz:
The helper script updating language files had insecure use patterns of `subprocess.run()`.

The use of `shell=True` should be avoided because of [security concerns](https://docs.python.org/3/library/subprocess.html#security-considerations).

I also added `check=True` to make all `run()` calls raise an exception if something goes wrong.

### DIFF
--- a/update_language_files.py
+++ b/update_language_files.py
@@ -3,7 +3,7 @@ import sys
 from pathlib import Path
 from subprocess import run, check_output
 
-"""This helper script do manage transfering translations to and from the
+"""This helper script do manage transferring translations to and from the
 translation platform (currently Weblate).
 """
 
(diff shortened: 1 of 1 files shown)


update_language_files.py
```diff
@@ -131,19 +105,24 @@ def check_existance():
...
 def update_from_weblate():
-    """
-        git clone --no-checkout https://translate.codeberg.org/git/backintime/common
-        cd common
-        git checkout dev -- common/po/*.po
+    """Translations done on Weblate platform are integrated back into the
+    repository.
 
+    The Weblate translations live on https://translate.codeberg.org and has
+    its own internal git reposiotry. This repository is cloned and the
```
aryoda: Typo: "repository"

aryoda (approved):
All fine (just a little typo -> could be fixed after merging)

aryoda (approved):

PR #1478, opened 2023-07-17
refactor: Menus, Toolbars and Actions in MainWindow

buhtz:
Partly reorganized code in MainWindow related to generation of menus, toolbars and the actions behind it. Also fixed some minor issues. The issues came up on the fly. I saw no way to separate them in a different Issue. But I keep to declare this PR/commit as "refactor" and not as "fix" because the Issues are tiny and didn't cause harm. See detailed list below.

I checked aryodas branch [`1306_tray_icon_master_issue`](https://github.com/aryoda/backintime/tree/issue/1306_tray_icon_master_issue) and can't find potential merge conflicts. But I'll wait for your feedback on that.

My first intention was to add a new menu to give the user the opportunity to modify the UI language. That is why the branch name is that way. But I realized I have to reorganize the code before doing this.

- The code try to follow the Qt concept of separate "actions" (e.g. start snapshot) from visual UI elements (buttons, menus and their entries). An action is not equal to a slot or signal. Good example to see its advantage is how the menu and toolbar elements related to restoring a snapshot are enabled/disabled.
- Generation of actions, menus and buttons are separated into multiple methods to reduce the amount of code in `__init__()`.
- Used underscores (`_`) for variable names (e.g. actions) because it is PEP8 convention. Camel case is suited for class names only.
- Removed some instance attributes `self.*` because there is no need of them. Qt elements are managed by QMainWindow (or components of it) in the back. No need to keep references of them just to keep them alive.
- Removed `qttools.Menu` class, which was a workaround to offer tooltips for menu entries. Qt 5.1 is able to do it itself.
- The buttons "Take snapshot" (main toolbar) and "Restore" (file view toolbar) now have drop-down menus. It was not my idea. As I could see in the "old" code it was intended to be a dropdown button in the past. But it never (?) worked. I just fixed it and made it work. See my comparing screenshots below.
- Main menu -> Snapshots: Removed the "Take Snapshots" submenu and integrated its only entry (Take snapshot with checksum) into the Snapshots menu.
- Removed the Help button from main toolbar. All its content is present in the main menus "Help" menu. It was also drop-down tool button which dropdown menu doesn't work. I wasn't motivated to fix it. I just removed it. I can re-add it without problem if someone like.

This screencast do illustrate the drop-down toolbar buttons and the new "Snapshots" menu.

Peek 2023-07-17 14-09

Let's compare this with this screenshot of an older version of BIT. Take a look at the three red circles. There are "down arrows" on each of that buttons. But they didn't work in the current BIT version because the drop down menu setup code wasn't finished. It might be that it worked in an older version with Qt4 or something like that.

image

### DIFF
--- a/qt/app.py
+++ b/qt/app.py
@@ -125,19 +125,6 @@ class MainWindow(QMainWindow):
         self.firstUpdateAll = True
         self.disableProfileChanged = False
 
-        # Sub-Menu: Take snapshot
-        takeSnapshotMenu = QMenu(self)
-        takeSnapshotMenu.setToolTipsVisible(True)
-        takeSnapshotMenu.addAction(self.act_take_snapshot)
-        takeSnapshotMenu.addAction(self.act_take_snapshot_checksum)
-
-        # Fix it. See original:
-        # https://github.com/bit-team/backintime/blob/1003d360d758bce78c6cd528ea41df3734fec95b/qt/app.py#L94-L101
-        for action in takeSnapshotMenu.actions():
-            # What is this?
-            # Isn't this by default?
-            action.setIconVisibleInMenu(True)
-
         # Menu: Snapshot
         menuSnapshot = self.menuBar().addMenu(_('&Snapshot'))
         menuSnapshot.addAction(self.act_quit)
@@ -719,13 +706,14 @@ class MainWindow(QMainWindow):
         for act in actions_for_toolbar:
             toolbar.addAction(act)
 
-        # sub menu: take snapshot
+        # toolbar sub menu: take snapshot
         submenu_take_snapshot = QMenu(self)
-        submenu_take_snapshot.setToolTipsVisible(True)
         submenu_take_snapshot.addAction(self.act_take_snapshot)
         submenu_take_snapshot.addAction(self.act_take_snapshot_checksum)
-        # get the toolbar buttons widget and add the menu to it
+        submenu_take_snapshot.setToolTipsVisible(True)
+        # get the toolbar buttons widget...
         button_take_snapshot = toolbar.widgetForAction(self.act_take_snapshot)
+        # ...and add the menu to it
         button_take_snapshot.setMenu(submenu_take_snapshot)
         button_take_snapshot.setPopupMode(QToolButton.MenuButtonPopup)
 
--- a/qt/app.py
+++ b/qt/app.py
@@ -47,7 +47,7 @@ import mount
 import progress
 from exceptions import MountException
 
-from PyQt5.QtGui import QKeySequence, QDesktopServices, QColor, QIcon
+from PyQt5.QtGui import QDesktopServices, QColor, QIcon
 from PyQt5.QtWidgets import (QWidget,
                              QAction,
                              QFrame,
@@ -71,15 +71,16 @@ from PyQt5.QtWidgets import (QWidget,
                              QToolBar,
                              QProgressBar,
                              QMessageBox,
-                             QSizePolicy,
                              QInputDialog,
                              QDialog,
                              QDialogButtonBox,
                              QShortcut,
                              QFileSystemModel,
+                             QSizePolicy
                              )
 from PyQt5.QtCore import (Qt,
                           QObject,
+                          QPoint,
                           pyqtSlot,
                           pyqtSignal,
                           QTimer,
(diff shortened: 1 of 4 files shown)


qt/app.py
```diff
@@ -69,267 +105,61 @@ def __init__(self, config, appInstance, qapp):
...
-        self.mainToolbar = self.addToolBar('main')
-        self.mainToolbar.setFloatable(False)
+        # "Magic" object handling shutdown procedure in different desktop
+        # environments.
+        self.shutdown = tools.ShutDown()
 
-        #window icon
+        # Import on module level not possible because of Qt restrictions.
         import icon
+        globals()['icon'] = icon
```
buhtz, line 120: That is a workaround until we implement a solution to our icon handling problem. I'll wait on aryodas trayicon-fix until that.

qt/app.py
```diff
@@ -1078,7 +1250,7 @@ def backupOnRestore(self):
         cb = QCheckBox(_(
             'Create backup copies with trailing {suffix} before\n'
             'overwriting or removing local files.').format(
-                suffx=self.snapshots.backupSuffix()))
+                suffix=self.snapshots.backupSuffix()))
```
buhtz, line 1258: A bugfix. Hitting restore button would raise an Exception. Bug was introduce by myself in the context of improving translations.

aryoda:
@buhtz My sys-tray icon PR #1480 is merged now int dev. You could continue with this PR.

Very thoughtful UX improvement I think!

I am still trying to review the (huge) diff, please give me a few days (git diffs are not the best if much code has been moved around)

buhtz:
> I am still trying to review the (huge) diff, please give me a few days (git diffs are not the best if much code has been moved around)

No problem and no hurry. Thanks.

Last time I used Qt was with C++ and Qt3 or 4 a 100 years ago. With this PR I learned some new things about Qt and how to handle it.

**EDIT**: Maybe viewing the diff in your terminal using `git diff` with `--ignore-all-space` and/or `--ignore-blank-lines` can help to improve the clearness of diff output.

qt/app.py
aryoda: THX for waiting, my change is merged, your way is clear now :-)

qt/app.py
```diff
@@ -555,6 +399,337 @@ def __init__(self, config, appInstance, qapp):
...
+        All actions are stored as instance attributes to ``self`` and their
+        names begin with ``act_``. The actions can be added to GUI elements
+        (menu entries, buttons) in later steps.
+
+        Note:
+            All actions used in the main window and its child widgets should
+            be created in this function.
+
+        Note:
+            Shortcuts need be strings in a list even if it is only one entry.
```
aryoda: typo: "need **to* be...

qt/app.py
```diff
@@ -665,7 +840,7 @@ def comboProfileChanged(self, index):
             if not path == self.path:
                 self.path = path
                 self.path_history.reset(self.path)
-                self.editCurrentPath.setText(self.path)
+                self.wdg_current_path.setText(self.path)
```
aryoda: wdg_ means what?

aryoda:
@buhtz Incredible work to introduce such a clear `QAction` abstraction layer into the GUI!

It is very good like it is now and is ready to be merged.

I have just two questions to learn more about the reasons to implement the actions using an `action_dict` which creates action "on the fly" (meta programming).

1. What is the reason and advantages to prefer this meta programming over using an action factory function basically containing the same code (+ using python code to create the actions then instead of using the `action_dict`) :

   https://github.com/bit-team/backintime/blob/8d418c537ac12443e395b08baf0a7691c57b1a22/qt/app.py#L542-L562

    ```
    # Let's assume the logic of above is implemented in the body of a factory function:
    # def action_factory(ico, txt, slot, keys, tip): ...

    self.act_take_snapshot = action_factory(icon.TAKE_SNAPSHOT, _('Take snapshot'),
                      self.btnTakeSnapshotClicked, ['Ctrl+S'], None))
    ```

2. The action handler code could be implemented via a lamda/anonymous function instead of an explicit function since it most of the time is only short (passes arguments on the the back end function that knows nothing about the GUI and perhaps changes the GUI state). See e.g. this code:
    ```
    action_dict = {
                'act_take_snapshot': (
                    icon.TAKE_SNAPSHOT, _('Take snapshot'),
                    self.btnTakeSnapshotClicked, ['Ctrl+S'], None),
    ...
        def btnTakeSnapshotClicked(self):
            backintime.takeSnapshotAsync(self.config)
            self.updateTakeSnapshot(True)
    ```

aryoda (approved):

buhtz:
> I have just two questions to learn more about the reasons to implement the actions using an `action_dict` which creates action "on the fly" (meta programming).
> 
> 1. What is the reason and advantages to prefer this meta programming over using an action factory function basically containing the same code

Good question. I don't have a computer science like answer to this. It is just a habit. I saw something like this before I kind a liked it. I see no disadvantage of a factory function. Maybe it is because I see actions and gui elements as data. So I put this data elements in a data structure like a dict. It is more a feeling than a strong professional opinion. 😃 

> 2. The action handler code could be implemented via a lamda/anonymous function

Yes I thought the same. But I just stopped at this point to prevent the blowing up the PR.

aryoda:
>> 1. What is the reason and advantages to prefer this meta programming over using an action factory function basically containing the same code

> Maybe it is because I see actions and gui elements as data. So I put this data elements in a data structure like a dict.

Absolutely OK so. One possible impact could be the missing or reduced intellisense and code navigation functionality in the IDE, let's try it and see how clever PyCharm is to handle this :-)

aryoda (approved):

Issue #1479, opened 2023-07-17
Scheduling for secondary profile is ignored: set to run every 4 days; runs every 15 minutes.


danielaixer:
Ubuntu MATE 22.04
BackInTime 1.3.3-3 (same issue on BIT 1.2.1-3)

I've created a second profile that appears as "profile2.name=Docs" in the config file.

This is the secondary profile scheduling settings:
![imagen](https://github.com/bit-team/backintime/assets/5376720/26fa491f-f8e6-46d6-9f62-648b36fa179b)

These are the schedule related entries that are generated on "~/.config/backintime/config":
```
profile2.schedule.custom_time=8,12,18,23
profile2.schedule.day=1
profile2.schedule.mode=25
profile2.schedule.repeatedly.period=4
profile2.schedule.repeatedly.unit=20
profile2.schedule.time=0
profile2.schedule.weekday=7
```

These are the relevant lines in the crontab file when running "crontab -e":
```
#Back In Time system entry, this will be edited by the gui:
*/15 * * * * /usr/bin/nice -n19 /usr/bin/ionice -c2 -n7 /usr/bin/backintime --profile-id 2 backup-job >/dev/null
```

I've even deleted the config file, created it from scratch and the issue persists.

To help us diagnose the problem quickly, please provide the output of the console command `backintime --diagnostics`.
```
backintime --diagnostics
Traceback (most recent call last):
  File "/usr/share/backintime/common/backintime.py", line 1190, in <module>
    startApp()
  File "/usr/share/backintime/common/backintime.py", line 507, in startApp
    args = argParse(None)
  File "/usr/share/backintime/common/backintime.py", line 568, in argParse
    args, unknownArgs = mainParser.parse_known_args(args)
  File "/usr/lib/python3.10/argparse.py", line 1871, in parse_known_args
    namespace, args = self._parse_known_args(args, namespace)
  File "/usr/lib/python3.10/argparse.py", line 2080, in _parse_known_args
    start_index = consume_optional(start_index)
  File "/usr/lib/python3.10/argparse.py", line 2020, in consume_optional
    take_action(action, args, option_string)
  File "/usr/lib/python3.10/argparse.py", line 1948, in take_action
    action(self, namespace, argument_values, option_string)
  File "/usr/share/backintime/common/backintime.py", line 742, in __call__
    diagnostics = collect_diagnostics()
  File "/usr/share/backintime/common/diagnostics.py", line 177, in collect_diagnostics
[...]

PR #1481, opened 2023-07-18
fix: Use C locale in diagnostics

buhtz:
C locale is always used when extracting version strings from external tools in diagnostics output.

In addition: Renamed "match" variable because it is a reserved keyword in newer Python versions.

Related #1479

### DIFF
--- a/common/diagnostics.py
+++ b/common/diagnostics.py
@@ -245,6 +245,7 @@ def _get_extern_versions(cmd,
     try:
         # as context manager to prevent ResourceWarning's
         with subprocess.Popen(cmd,
+                              env={'LC_ALL': 'C'},
                               stdout=subprocess.PIPE,
                               stderr=subprocess.PIPE,
                               universal_newlines=True) as proc:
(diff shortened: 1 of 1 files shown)


aryoda:
I am not sure if setting LC_ALL will really work (at least I had a lot of problems with that with R).

If do a quick test the other way around it does not work but still print English output on my machine (U22.04):

`env LC_ALL=de_DE.UTF-8 rsync --version`

Perhaps it works only with `C` as value...

buhtz:
On a bash I have to do `export LC_ALL=` instead of `env LC_ALL=`. ~Don't know why "env" does not work.~ I was using `env` wrong in the past using `&&` to connect it with a command: `env LC_ALL=C && encfs` is of course wrong.

I installed italian local on my machine (via `sudo dpkg-reconfigure locales` on Debian 11). And then I checked it inside a Python REPL using the `Popen(cmd, env=...)`. This worked fine.

aryoda:
> On a bash I have to do `export LC_ALL=` instead of `env LC_ALL=`.

Good to know, THX for testing!

I have justs asked the user of the underlying issue to to a test with `env`, he could also do test the without `env` to be sure that at least one variant is working on his computer...

aryoda:
Good work, ready to merge!

OP did successfully test that LC_ALL=C works (with or without `env` doesn't matter here, since in Python we use `popen`: https://github.com/bit-team/backintime/issues/1479#issuecomment-1640308227

PR #1486, opened 2023-07-23
fix!: Correct SYSLOG_IDENTIFIER

### LINKED ISSUE #1451: dummytools.py: F821 Undefined name "parent"

### LINKED ISSUE #1483: Missing identifier in syslog entries
# Introduction
Back In Time do use pythons `syslog` module to do its application specific logging (not snapshots).
On `systemd` using GNU Linux systems this syslog entries to appear in the `journalctl` output. Syslog do not log just a string and timestamp but several fields.

# Problem
The identifier BIT do use for syslog entries is not recognized (by syslog deamon or systemd). It is because there is a blank character in the identifer. It took me some time and tests to reproduce this and find out what the problem was. It might be [introduced](https://github.com/bit-team/backintime/commit/db7ca8687ffff7d4243e2f308b6c5f620bdb4314) in 2008 and it could be that it worked these days and the syslog behavior did changed since then.

BIT do set its identifier using `syslog.openlog()`.
https://github.com/bit-team/backintime/blob/71c9657324c0e08dd3288dbff830dd90c1cc15d2/common/logger.py#L29-L36

This is the result in `journalctl`.

    Sep 11 02:00:01 temple python3[183882]: backintime (zero/1): INFO: Take a new snapshot. Profile: 1 Main profile

The identifier string (`backintime (zero/1)`) setup by BIT in its `syslog.openlog()` call become part of the message string itself (`MESSAGE` field). Journalctl do offer its output also in JSON format which made it easier to identify the fields it uses. There is no `SYSLOG_IDENTIFIER` field and no `SYSTEMD_UNIT`. That makes it impossible to identify backintime related log entries without hacking around with `grep` and shell pipes/redirection.

# Simple fix
    syslog.openlog("%s (%s/1)" %(APP_NAME, name))
                      ^

This blank need to be removed. Then it works. I checked that.

# Fix with improvement
My intention was to identify backintime log entries. It is not possible with that fix because the identifier is not only `backintime` but mixed up with additional information. The identifier is specific to a user and the used snapshot profile id. It is official (because of systemd bug reports) that systemd won't support pattern matching (like grep, regex) for specific fields (but only for unit names).

My proposal
 - Make the identifier `backintime` in all situations, all instances including the GUI (currently using `backintime-qt (x/y)`).
 - Remove the user name without replacement. Because the users name is part of the syslog fields and easy to find via journalctl.
 - Move the profile-id to the beginning of the message and include the profile name. In that case redirecting and grepping is possible. Example: `backintime: ProfileName(3): INFO: Message`

buhtz:
Always use "backintime" as identifier for syslog entries.

It fix a long year bug where syslog used the python interpreter as identifier instead of backintime. Beside of that the snapshot profile now is part of message string. The problem was caused because the identifier used by BIT in `syslog.openlog()` did contain a blank character. In that case it seems that syslog do silently ignore that identifier.

BREAKING CHANGE: This does modify the syntax of syslog entries.

Fix #1483
Close #1451

Syslog entries now looking like this (using `journalctrl`):

```
Jul 23 14:45:01 TONNE backintime[10655]: playground(3) :: INFO: Profile "playground" is not scheduled to run now.

# While running unittests
Jul 23 14:38:11 TONNE backintime[7638]: Main profile(1) :: INFO: unmount ssh: user@localhost:/tmp/tmpd171t87j/foo from /tmp/tmp8ruhibed/.local/share/backintime/mnt/D52CD276/mountpoint
Jul 23 14:38:11 TONNE backintime[10194]: Main profile(1) :: ERROR: pidfile /home/user/.local/share/backintime/password_cache/PID already exists. Daemon already running?
Jul 23 14:38:11 TONNE backintime[10201]: Main profile(1) :: ERROR: pidfile /home/user/.local/share/backintime/password_cache/PID already exists. Daemon already running?
Jul 23 14:38:12 TONNE backintime[7638]: Main profile(1) :: INFO: mount ssh: user@localhost:/tmp/tmpuqy_g6sh/foo on /tmp/tmpdyklkh4z/.local/share/backintime/mnt/8CC4CED7/mountpoint
Jul 23 14:38:12 TONNE backintime[7638]: Main profile(1) :: INFO: Call rsync to take the snapshot
Jul 23 14:38:12 TONNE backintime[7638]: Main profile(1) :: INFO: Save config file

# Using --debug. In the first two entries you see that the profile isn't set yet (config not loaded at this time).
ul 23 14:47:30 TONNE backintime[10707]: DEBUG: [common/backintime.py:589 argParse] Arguments: {'debug': True} | unknownArgs: []
Jul 23 14:47:30 TONNE backintime[10707]: DEBUG: [common/configfile.py:590 Config.setCurrentProfile] Change current profile: 1=Hauptprofil
Jul 23 14:47:30 TONNE backintime[10707]: Hauptprofil(1) :: DEBUG: [common/backintime.py:677 getConfig] config file: /home/user/.config/backintime/config
Jul 23 14:47:30 TONNE backintime[10707]: Hauptprofil(1) :: DEBUG: [common/backintime.py:678 getConfig] share path: /home/user/.local/share/backintime
```

Now you can use `journalctl -t backintime` to look for all entries with syslog identifiers "backintime".


aryoda:
FYI: I need more time to "review" (= learn new things) the code changes, perhaps a week

PR #1487, opened 2023-07-26
feat: Rearrange menu bar in main window

### LINKED ISSUE #1482: Discussion: re-arrange the menubar and its entries
I would like to discuss if and how we could rearrange the mainwindows menubar and its entries. Additionally I see the need to separate the multiple meanings of "snapshot" and find consensus about alternative terms.

# Current situation

image

# My intention and goals
My intention in the back is to place a "Setup language" menu entry somewhere where the users can select the UI language. I couldn't find a good place. Add a new menu entry on the first level seems a bit to much. We also don't have an application specific settings dialog where this usually would be placed.

I don't want to make it perfect. This means I don't want to to modify and fix everything that could be discusses. For example the help menu itself needs to be discussed; but not here. I want to keep it as simple as possible.

# What is a "Snapshot"?
One other thing I want to take into account is the terminology about "Snapshot". In my opinion this term does have multiple meanings in the way it is currently used in BIT's UI (and also in the code base). We need to separate the meaning between:

1. A "snapshot" as the result of an rsync run. It is in the backup destination a folder named with a specific timestamp.
2. The "profile" (or alternative term; e.g. "backup job", "backup task") which is represented by all `profile*` lines in the config file, in the settings dialog and the whole mainwindow itself.

I think the meaning of "snapshot" is clear. But we should find a term for what I would call a "profile" and using it consequently. I choose "profile" because this term is used in "Main profile", in the settings dialog and sometimes in the terminal output.

# My rough proposal about the menubar
- New menu "Back In Time" as first and application specific menu.
- New menu "Profile" was added.
- I removed the "View"
  - Entries "Up" and "Show hidden" are removed without replacements. They do exists in the fileview toolbar.
  - Other entries are moved to snapshot or profile menu.

```
- Back In Time
    - Setup language
    - Shutdown
    - Exit
- Profile
    - Settings
    - Take a snapshot
    - Take a snapshot with checksum
    - View last log
    - Refresh snapshot list
    - Snapshots
- Snapshot
    - Remove
    - View log
- Restore
    - Restore
    - Restore to ...
    - Restore {current_folder}
    - Restore {current_folder} to ...
- Help
    - Help
    - Config
    - Website
    - Changelog
    - FAQ
    - Ask a question
    - Report a bug
    - About
```

The "Snapshot" and "Restore" menus are only enabled if a snapshot is selected in the timeline.

I could be said that "Restore" is related to a specific "Snapshot" and should belong to this menu. I agree with this logic. But "Restore" is one of the main features. It might not be consistent but it is more "shiny" to have this as a separate menu.
On the other hand the "Snapshot" only has two entries. It wouldn't hurt to move the entries from "Restore" into it. What do you think?

"Refresh snapshot list" could be renamed into "Refresh snapshot timeline" or "Refresh timeline".

"Profile" -> "Snapshots" opens the `SnapshotsDialog`. As I discussed in #1485 that dialog can be removed in the future. But until then we need to place its entry point somewhere in the menus and find a label for it. The current label is IMHO to broad but I can't find a better solution.

As I said there is much more room for improvement. But would keep it simple only would modify it that way that I'm able to have a good place for the language menu. Everything else should be done in a separate Issue/PR and is not urgent.

buhtz:
Rearrange the menu bar in main window and its entries. Some labels and tooltips are modified.

I would say the entry for a Language Selection and maybe for Diagnostics can be handled in a separate PR.

Different from our discussion in #1482 I modified the order of the entries in the sub menus a bit. It is just a proposal and easy to modify.
Peek 2023-07-26 11-38

Close #1482

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,2 @@
+Don't forget the changelog entry! ;)
+
--- a/CHANGES
+++ b/CHANGES
@@ -11,15 +11,17 @@ Version 1.3.4-dev (development of upcoming release)
 * Fix bug: Generation of config-manpage caused an error with Debian's Lintian (#1398).
 * Fix bug: Return empty list in smartRemove (#1392, Debian Bug Report 973760)
 * Breaking change: Minimal Python version 3.8 required (#1358).
+* Removed: Handling and checking of user group "fuse" (#1472).
 * Feature: Exclude /swapfile by default (#1053)
+# Feature: Rearranged menu bar and its entries in the main window (#1487, #1478).
 * Documentation: Removed outdated docbook (#1345).
 * Build: Introduced .readthedocs.yaml as asked by ReadTheDocs.org (#1443).
 * Dependency: The oxygen icons should be installed with the BiT Qt GUI since they are used as fallback in case of missing icons
 * Fix bug: Add support for ChainerBackend class as keyring which iterates over all supported keyring backends (#1410)
 * Translation: Strings to translate now easier to understand for translators (#1448, #1457, #1462).
 * Translation: Updated and completed "German" (#1454).
-* Translation: Remove language Canadian English, British English and Javanese (#1455).
-* Translation: Persian and Vietnamese added (#1460).
+* Removed: Translation in Canadian English, British English and Javanese (#1455).
+* Added: Tranlsation in Persian and Vietnamese (#1460).
 
 Version 1.3.3 (2023-01-04)
 * New feature: Command line argument "--diagnostics" to show helpful info for better issue support (#1100)
--- a/CHANGES
+++ b/CHANGES
@@ -21,7 +21,7 @@ Version 1.3.4-dev (development of upcoming release)
 * Translation: Strings to translate now easier to understand for translators (#1448, #1457, #1462).
 * Translation: Updated and completed "German" (#1454).
 * Removed: Translation in Canadian English, British English and Javanese (#1455).
-* Added: Tranlsation in Persian and Vietnamese (#1460).
+* Added: Translation in Persian and Vietnamese (#1460).
 
 Version 1.3.3 (2023-01-04)
 * New feature: Command line argument "--diagnostics" to show helpful info for better issue support (#1100)
--- a/qt/app.py
+++ b/qt/app.py
@@ -429,13 +429,14 @@ class MainWindow(QMainWindow):
             #     tooltip
             #),
             'act_take_snapshot': (
-                icon.TAKE_SNAPSHOT, _('Take snapshot'),
-                self.btnTakeSnapshotClicked, ['Ctrl+S'], None),
+                icon.TAKE_SNAPSHOT, _('Take a snapshot'),
+                self.btnTakeSnapshotClicked, ['Ctrl+S'],
+                _('Use modification time & size for file change detection.')),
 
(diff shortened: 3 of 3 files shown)


aryoda (approved):

aryoda:
All good!

Perhaps the shortcut for "Manage profiles" could be changed now (Ctrl + Shift + , is clumsy and neccessary because Ctrl + Shift + S was already used for "Take snapshots with checksum").

I think Ctrl + Shift + M (or P) could still be unused and more intuitive now...

BTW: Don't forget the changelog entry (could be done later when the GUI refactoring is change-stable)

buhtz:
> BTW: Don't forget the changelog entry (could be done later when the GUI refactoring is change-stable)

Always... 🍌 I also added some "lost" changelog entries from previous PRs.

btw: Do you know about https://common-changelog.org/ as a standard how to write a changelog and its entries? I'm not experienced with it but try it in one of [my own one-person-projects](https://codeberg.org/buhtz/buhtzology/src/branch/develop/CHANGELOG.md).

CHANGES
```diff
@@ -11,15 +11,17 @@ Version 1.3.4-dev (development of upcoming release)
...
 * Documentation: Removed outdated docbook (#1345).
 * Build: Introduced .readthedocs.yaml as asked by ReadTheDocs.org (#1443).
 * Dependency: The oxygen icons should be installed with the BiT Qt GUI since they are used as fallback in case of missing icons
 * Fix bug: Add support for ChainerBackend class as keyring which iterates over all supported keyring backends (#1410)
 * Translation: Strings to translate now easier to understand for translators (#1448, #1457, #1462).
 * Translation: Updated and completed "German" (#1454).
-* Translation: Remove language Canadian English, British English and Javanese (#1455).
-* Translation: Persian and Vietnamese added (#1460).
+* Removed: Translation in Canadian English, British English and Javanese (#1455).
+* Added: Tranlsation in Persian and Vietnamese (#1460).
```
aryoda: Typo in "translation"

aryoda (approved):

aryoda:
> btw: Do you know about https://common-changelog.org/ as a standard how to write a changelog and its entries? 

I can't remember having read this one and I am little bit in-consequent here ("as it fits")

What I always hate is the lurking merge conflict in the changelog file if multiple devs work in different feature branches and I have already seen a fancy script in a project where the changelog entry per feature is contained in a separate folder and merge into the changelog when the feature is merged into the dev branch (all this to avoid merge conflicts)

buhtz:
> I can't remember having read this one and I am little bit in-consequent here ("as it fits")

I'm also in-consequent. That's why I need a standard because I can't do it by my own. 😃 

> I have already seen a fancy script in a project where the changelog entry per feature is contained in a separate folder and merge into the changelog when the feature is merged into the dev branch (all this to avoid merge conflicts)

Yes, I've seen things like this, too. Hard to track for external contributors. Make sense in projects with more complex branching models then ours. We don't collect multiple features in separate public branches. We merge into "dev" all the time.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,2 @@
+Don't forget the changelog entry! ;)
```
aryoda, line 1: :-)

aryoda (approved):

aryoda (approved):

aryoda:
@buhtz Do you want to merge this, it looks fine!?

PR #1492, opened 2023-08-07
fix: Naming of actions/buttons variables in MainWindow

buhtz:
Fix naming of action variables in main window.

This was introduced via refactoring buttons/actions in the main window.


FAQ.md
```diff
@@ -143,6 +144,13 @@ ps aux | grep -i backintime
 If the output shows a running instance of _Back In Time_ it must be waited until it finishes
 or killed via `kill <process id>`.
 
+### The application is already running! (pid: 1234567)
+This message occurs when _Back In Time_ did not finish regularly and wasn't able
+to delete its application lock file. Before deleting that file make sure
+no backintime process is running via `ps aux | grep -i backintime`. Otherwise
+kill the process. After that lock into the folder
```
aryoda: "lock" = look ?

BTW: I think the lock file contains the process ID (PID) of BiT so I am wondering why BiT is not able to detect at startup that the existing lock file references a no longer existing PID and resets the lock file itself...

aryoda (changes requested):

aryoda:
@buhtz I have merged this right away since I need this fix too :-)

Issue #1371, opened 2022-11-22
Can't remove defunct main profile


jwbrase:
I recently moved my backups from a locally attached external drive to a TrueNAS box on the LAN. I ran both for a while, and once I was sure that the remote backups were working, I removed the external drive. Unfortunately, it seems that BackInTime does not allow the first profile created to be removed (I guess to prevent people from removing all of their profiles by accident?). I've disabled it, and set the backup directory to a dummy directory (otherwise any time I try to change the new profile, the "main" profile generates an error dialog about a missing backup directory and doesn't allow the profile settings dialog to be closed), but I'd like to be able to get rid of it completely.

buhtz:
Dear Jon,
thank you for your report.

This is a known shortcoming of BIT that the "main profile" is undeletable. And I'm not aware of a good workaround for that.

What do you mean by "disabled" the profile? IMHO there shouldn't be an error message when you
 - set the profile to "Local"
 - deactivate the time scheduling
 - and give it an existing folder

Did you tried that?

Maybe related to #1032

jwbrase:
"Disabled" means I turned off scheduling. I've already set the backup directory to an existing directory that's excluded from the backup scope. The main profile isn't doing anything at the moment and I've resolved the issue with not having an existing backup folder. Mostly at this point I guess I'm just asking for a workaround and making a feature request for some better way of handling this.

As for the workaround, you say there's no good workaround; Could one in principle hand edit config files or something to change what's considered the "main" profile? If not, I guess I'll keep the profile around, but it would be nice to be able to clean it up.

As for ways that BIT could handle this better, I guess it depends on exactly why the main profile is unremovable. If it's a safety feature to prevent the user from doing stupid things, it doesn't prevent the user's most important profile from being removed if their most important is not their first, so I'm not sure it actually makes things all that much safer. If the application needs at least one profile to exist at all times for some reason, maybe have some way of designating a new main profile?

buhtz:
I agree with you. The current situation is not a good solution and there is IMHO no good reason to keep the main profile.

We will keep this as a feature request. But as far as I can see this won't get fixed in the near future because of Issues with higher priority.

My suggestions for fix would be
 - BIT should allow to delete each profile including the last one.
 - Starting BIT again it should act like it does when starting the first time: Forcing to setup a profile.
 - By default BIT should select the last opened profile in the list.
   - This will make problems (error messages) when there is a SSH profile with unavailable remote.
[...]

PR #1493, opened 2023-08-08
feat: Language in config file

buhtz:
New key "global.language" in config file to customize the UI language.

Language codes are used as values. If empty (or key not present) the systems current locale is used. If "en" or unknown the sources strings (in English) are used as fallback.

No matter that it hurt me I decided against creating a unit test for "global.language" config value. I have a refactoring about the config-file handling in my mind which will ease up a lot of things including unit testing.

### DIFF
--- a/common/config.py
+++ b/common/config.py
@@ -560,7 +560,7 @@ class Config(configfile.ConfigFileWithProfiles):
         return self.strValue('global.language', '')
 
     def setLanguage(self, language: str):
-        self.setStrValue('globl.language', language)
+        self.setStrValue('global.language', language)
 
     # SSH
     def sshSnapshotsPath(self, profile_id = None):
--- a/common/config.py
+++ b/common/config.py
@@ -192,6 +192,9 @@ class Config(configfile.ConfigFileWithProfiles):
     PLUGIN_MANAGER = pluginmanager.PluginManager()
 
     def __init__(self, config_path=None, data_path=None):
+        # Note: The main profiles name here is translated using the systems
+        # current locale because the language code in the config file wasn't
+        # read yet.
         configfile.ConfigFileWithProfiles.__init__(self, _('Main profile'))
 
         self._APP_PATH = tools.backintimePath()
@@ -326,7 +329,10 @@ class Config(configfile.ConfigFileWithProfiles):
         self.inhibitCookie = None
         self.setupUdev = tools.SetupUdev()
 
-        tools.initiate_translation()
+        tools.initiate_translation(self.language())
+
+        # Workaround
+        self.default_profile_name = _('Main profile')
 
     def save(self):
         self.setIntValue('config.version', self.CONFIG_VERSION)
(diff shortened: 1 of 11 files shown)


common/config.py
```diff
@@ -192,6 +192,9 @@ class Config(configfile.ConfigFileWithProfiles):
     PLUGIN_MANAGER = pluginmanager.PluginManager()
 
     def __init__(self, config_path=None, data_path=None):
+        # Note: The main profiles name here is translated using the systems
+        # current locale because the language code in the config file wasn't
+        # read yet.
         configfile.ConfigFileWithProfiles.__init__(self, _('Main profile'))
```
buhtz, line 198: In the future we should remove the ability to translate the first profiles name. A name as an identifier shouldn't be translated at all. Related to #1371

common/config.py
```diff
@@ -326,7 +329,10 @@ def __init__(self, config_path=None, data_path=None):
         self.inhibitCookie = None
         self.setupUdev = tools.SetupUdev()
 
-        tools.initiate_translation()
+        tools.initiate_translation(self.language())
+
+        # Workaround
+        self.default_profile_name = _('Main profile')
```
buhtz, line 335: Here the main profiles name is translated again now using the correct configured language. A bit dump I know. 😄

common/config.py
aryoda: or even better: Why do we need a default profile at all?

aryoda:
I did not yet deep dive into the change set but AFAIR changing the config file format (here: by adding a new entry) requires to increment the config file version in `common/config.py` (CONFIG_VERSION) so that BiT automatically updates older config files to the new format.

I couldn't see this in the change set (and this also possibly requires to update the config file format of unit test configs to avoid an unexpected output that fails some unit tests)...

buhtz:
> but AFAIR changing the config file format (here: by adding a new entry) requires to increment the config file version

I'm aware of that version variable but I I don't see a need to increment it. Adding a new variable do not "change the format". It do not break existing behavior. And the variable do not need to be present in the config file because there is a default value for it.

Technically there is no need to update old config files if you do not modify the name or format of existing variables. If the line `global.language` is not present the default value is used. If the value is stored the first time it is just added to the config file. This is quit flexible.

aryoda:
> Adding a new variable do not "change the format"

I think reusing the new config format with an older BiT version will no longer work (though I am not sure if this is a use case).

Anyhow I think it would be more clean to increment the config file version every time we change the format (even though it is non-destructive/additive).

I am not sure how Germar handled this in the past, perhaps browsing the change history of the version could help.

buhtz:
> I think reusing the new config format with an older BiT version will no longer work (though I am not sure if this is a use case).

It is an unusual case but possible. I'll test that. But I'm quit deep into the code of `config.py` and `configfile.py`. In my understanding this won't be a problem. The class read only what it needs and ignores the rest.

> Anyhow I think it would be more clean to increment the config file version every time we change the format (even though it is non-destructive/additive).

I don't think it is clean to even have a version number for config formats. That is my problem with the solution. I don't see a need for such a version number. There are other ways to handle modifications of variable names and similar things in a config file.

buhtz:
> I'll test that.

I added two lines to my BIT 1.3.3 config file.

```
foo.bar=8
global.bähm=nein
```

No problem. I can read all snapshot profiles and also modify them. That two lines are still in the config file.

aryoda (approved):

PR #1494, opened 2023-08-10
feat: Customize UI language via GUI

buhtz:
Customize UI language with GUI dialog.

- I try not to explain to much in this PR comment because the code and its docu should be understandable by itself. I assume it is not yet the case. 😄  So please ask back and I'll improve the docstrings and comments.
- The dialog with its radiobuttons is big. That was my intention to make it clear how cool BIT is and how many languages we do support. Of course I can change that and make it a simple list box or something like this. No big deal.
- I tried to design all this in a way that it is resistent against errors caused by missing language files, unknown languages, etc pp
- Keep the following in mind
    - The "system default language" (aka the "locale") can have multiple language codes at once when users do live in multilingual environments. So the language really used by BIT might be different from the one that is configured in BIT because of several reasons.
    - The 2-letter language codes (e.g. `de`) are not the same used in GNU Linux `locale` (e.g. `de_DE.UTF-8`)
- One aspect that makes this PR complex is the feature that the language dialog do display the name of each language in three different flavours:
    1. Native name using native letters. (e.g. `日本語` for Japanese)
    2. Current UI language (e.g. `Japanisch` when the UI language is German)
    3. In English (the source language) (e.g. `Japanese`)
- To realize this flavors-feature we need to have a list of language names in all this flavors. To avoid to much dependencies I decided to generate that list using the package `babel` via the `update_language_files.py` script when downloading the translations (po-files) from Weblate. `babel` is used only in that helper script and don't become a dependency to the project because of that.

Peek 2023-08-10 12-51

Some details to that screencast:
 - It is created on a system using `de_DE.UTF-8` (German) as current locale.
 - See "Japanisch" as a full example.
 - "Koreanisch" (eng: Korean): The font used by my Qt/Desktop do not support korean letters. BIT does check this (`qttools.can_render()`) and don't use the native language name in that case to prevent errors.
 - "xy" Also languages not listed in the language names list are used and don't cause errors. This for example happens when you use BIT 1.3.4 on a system where BIT 1.3.3 is installed which comes with languages that where removed in 1.3.4 (e.g. Canadian and British English).


qt/app.py
```diff
@@ -1727,6 +1736,23 @@ def suspendMouseButtonNavigation(self):
...
+        dlg = LanguageDialog(
+            used_language_code=self.config.language_used,
+            configured_language_code=self.config.language())
+
+        dlg.exec()
+
+        if dlg.result() == 1:
+            self.config.setLanguage(dlg.language_code)
+
+            # How to send notify message (systray bubble)?
```
buhtz: It is possible to modify the language used by gettext and Qt on the fly without restarting the whole application. But BITs current code structure do not allow it. So users have to restart BIT as most other GUI applications to my knowledge realizing it.

Here I would like to inform the users about that fact. I simple systray bubble I would say. How can we do this with BIT?

buhtz: I was looking in #1156 and found about `NotifyPlugin` class. But I can not see where it is used. It seems to be unused.
I also tried something like `self.config.PLUGIN_MANAGER.error("foo")` but without effect.

I don't understand how BIT produce the systray bubbles/balloons.

buhtz: I could find `common/pluginmanager.py` and the class `PluginManager` in there.
I don't understand that approach. Why do I need a "plugin" to send notify messages? And all this happens without `logger.py`.
If I would understand the intention behind that approach I could come up with a solution/improvement.

This smells like a lot of refactoring and redesign work. So I'll stop at this point.

The question is how should we inform the user about the fact that the language settings taking effect only after restarting BIT? IMHO a message box would be a pragmatic solution.

buhtz: Peek 2023-08-11 12-26

This is how it looks like with a info box that shows up only when the language was modfied.

buhtz:
Here is an idea how to solve the problem with the width. The radio buttons label now has two lines.
image

Another question: Currently the radio buttons are ordered by language code column wise. Might it be better do make it row wise? In that case the user would read from left to right.

Question: It is also no problem to order the radio buttons lexicographical by the displayed language name instead of the language code.

aryoda:
> The radio buttons label now has two lines.

This is much better and faster to read now.

Still I think the current 3-column radio button + scroll bar layout will not fit into any existing UI guidelines - but does work of course very well.

A simple combo box with auto-complete + allowing only values from the list is a commonly seen way of choosing a language without wasting too much space in a settings windows.

Since we have now common "BiT" settings dialog (or menu entry) a combo box would look a little bit "alone" in a window though :wink:

buhtz:
Thanks for your feedback.

> Still I think the current 3-column radio button + scroll bar layout will not fit into any existing UI guidelines - but does work of course very well.

Yes, it is very unconventional. 🤣

> A simple combo box with auto-complete + allowing only values from the list is a commonly seen way of choosing a language without wasting too much space in a settings windows.
> 
> Since we have now common "BiT" settings dialog (or menu entry) a combo box would look a little bit "alone" in a window though 😉

Replacing the dialog with another GUI element is no big deal. Most of the work was done somewhere else in the code.

If no one else comes up with an alternative approach I would say we keep the dialog that way until we implement an application settings dialog where we can integrate it then.

aryoda:
> If no one else comes up with an alternative approach I would say we keep the dialog

Fully agree (it does work)

aryoda (approved):

Issue #1491, opened 2023-08-07
user-callback is not called properly when an error occurs


jean-christophe-manciot:
backintime-common/backintime-qt: 1.3.3-4 https://fr.archive.ubuntu.com/ubuntu lunar/universe amd64 Packages

When the "[Invalid cross-device link](https://github.com/bit-team/backintime/issues/851)" issue happens, the snapshot fails but the user-callback is not called properly. It should be called as:
```
user-callback <profile_id> <profile_name> 4
```
I discovered this issue because my user-callback is supposed to send an email when an error (4) happens, but I received nothing.
However, If I run it manually with the right parameters, for instance ```./user-callback 1 Main 4```, the email is sent & received. 

Attached files:
1) bit diagnotics: [bit-diags.txt](https://github.com/bit-team/backintime/files/12274947/bit-diags.txt)
2) [user-callback](https://github.com/bit-team/backintime/files/12274967/user-callback.txt)

aryoda:
I think this bug may occur whenever `rsync` exits with an error.

I will look into our code and try to fix this when I find the time for this (holidays at the moment).

The underlying error reason (documented in #851) I cannot fix unfortunately, this needs the eyes of an `rsync` specialist (cmd arg tweaking required I guess)...

aryoda:
Possibly related to #1053

aryoda:
## First bug analysis results

I could reproduce the underlying error ("invalid cross-device link", see https://github.com/bit-team/backintime/issues/851#issuecomment-1670423273) and have traced which user callback functions are called in case of a successful vs. erroneous backup (`takeSnapshot`):

```
------- user callback sequence for backup w/o errors ---------------------------
2023-08-09 07:32:35 (<user>/168194) profile_id=1: 7 - mount drive requested
2023-08-09 07:32:36 (<user>/168199) profile_id=1: 1 - backup process begins
2023-08-09 07:32:38 (<user>/168310) profile_id=1: 3 - new snapshot taken (snapshot_id=20230809-093235-919 - snapshot_name=/home/<user>/temp/backintime/AMD5600x/<user>/1/20230809-093235-919)
2023-08-09 07:32:38 (<user>/168313) profile_id=1: 2 - backup process ends
2023-08-09 07:32:38 (<user>/168317) profile_id=1: 8 - unmount drive requested

---- user callback sequence with an rsync error 23 in takeSnapshot -------------
[...]

Issue #489, opened 2015-11-08
Killing rsync not handled gracefully / rsync exit code not evaluated


kerner1000:
When I abort a running backup by
<code>sudo killall rsync</code> BIT does not flag this snapshot with <code>WITH ERRORS!</code>

Germar:
BIT doesn't handle return codes from rsync yet. It's on my Todo list. But before I need to switch all processes to be called by `subprocess.Popen`

protist:
For the record, I ran into this issue too. I had a network connection go down, and my backup was aborted. Despite setting backintime to delete the backup on errors, this meant I had an empty backup directory for that particular timepoint. The next backup was thus not incremental, and took a massive amount of space. Unfortunately this happened several times several months ago, and I only noticed it now.

To fix it, I moved the bad backup(s) to a temporary directory, created a hard-link copy of the last good backup, replaced the changed files, then deleted the old directory.

For example, where `20200719-160001-721` is a bad backup and `20200430-061501-472` is the last good backup:
```
$ mv 20200719-160001-721 tmp/
$ sudo rsync -aPHAX --link-dest=/path/to/backintime/hostname/root/1/20200430-061501-472 20200430-061501-472/ 20200719-160001-721
$ sudo rsync -aPHAX tmp/20200719-160001-721/ 20200719-160001-721 --delete
$ sudo rm -r tmp/20200719-160001-721
```

I imagine there are faster ways to do this. For my ~430G backup, the hard-link copy took 15 minutes, and the replacement rsync took 1h 36m (!).

Moving forward, I have a cron job on my server to check the log file for errors (then delete the last backup if necessary). Something like (untested):
```
#!/bin/sh

set -e

cd /path/to/backintime/hostname/root/1

errors=$(bzcat last_snapshot/takesnapshot.log.bz2 | grep '^\[I\] Take snapshot (rsync: rsync error' -B3)

if [ -n $errors ]; then
  echo 'Errors found! Log shows:'
  printf '%s\n\n' "$errors"

  printf '%s' "Removing $(realpath last_snapshot) . . ."
  rm "$(realpath last_snapshot)"
  echo removed!
  ln -s "$(ls 20* -d | tail -1)" last_snapshot
fi
```

aryoda:
[...]

PR #1502, opened 2023-08-16
Fix for #1491 (error user callback not called sometimes) and #489 (killing rsync not handled gracefully / rsync exit code not evaluated)

aryoda:
- Fix bug: Taking a snapshot reports `rsync` errors now even if no snapshot was taken
   (#1491)
- Fix bug: takeSnapshot() recognizes errors now by also evaluating the rsync exit code
  (#489)
         Fixes related problem: Killing `rsync` was not handled gracefully (by ignoring the rsync exit code)
- Fix bug: The error user-callback is now always called if an error happened during taking a snapshot
   (#1491)
- Improvement: Introduce new error codes for the "error" user callback
   (as part of #1491):
             5: Error while taking a snapshot.
             6: New snapshot taken but with errors.
- Feature: The `rsync` exit code is now contained in the snapshot log. Example:
             [E] Error: 'rsync' ended with exit code -9 (negative values are signal numbers, see 'kill -l')

## Notes

I did only apply required changes to fix the issues since the `takeSnapshot()` function is the "heart" of BiT
and has a quite complicated logic (many `if`s, legacy parameter passing with tuples instead of named classes etc.).

Still I have added a lot of TODOs, comments and docstrings to make the logic better understandable (I hope).

## Testing

I have done many manual tests & debugging with different signals, `rsync` errors and with and without the `Take a new snapshot whether there were changes or not` option.

The unit tests do succeed (`make test`) on my local machine.


CHANGES
```diff
@@ -10,6 +10,15 @@ Version 1.3.4-dev (development of upcoming release)
 * Fix bug: Unit test fails on some machines due to warning "Ignoring XDG_SESSION_TYPE=wayland on Gnome..." (#1429)
 * Fix bug: Generation of config-manpage caused an error with Debian's Lintian (#1398).
 * Fix bug: Return empty list in smartRemove (#1392, Debian Bug Report 973760)
+* Fix bug: Taking a snapshot reports `rsync` errors now even if no snapshot was taken (#1491)
+* Fix bug: takeSnapshot() recognizes errors now by also evaluating the rsync exit code (#489)
+           Fixes related problem: Killing `rsync` was not handled gracefully (by ignoring the rsync exit code)
+* Fix bug: The error user-callback is now always called if an error happened during taking a snapshot (#1491)
+* Improvement: Introduce new error codes for the "error" user callback (as part of #1491):
```
buhtz: Based on changelog standards I would name this a "Feature".

common/pluginmanager.py
```diff
@@ -91,14 +91,19 @@ def error(self, code, message):
 
                 Known error codes:
 
-                1. No or no valid configuration
+                1: No or no valid configuration
```
buhtz, line 94: I assume that new codes should be mentioned in the [user-callback](https://github.com/bit-team/user-callback) repo also.

common/snapshots.py
```diff
@@ -129,6 +129,24 @@ def takeSnapshotMessage(self):
 
     #TODO: make own class for takeSnapshotMessage
     def setTakeSnapshotMessage(self, type_id, message, timeout = -1):
+        """Update the status message of the active snapshot creation job
```
buhtz, line 132: Awesome. Never really get what this function does! :rofl:

common/snapshots.py
```diff
@@ -783,12 +801,20 @@ def backup(self, force = False):
...
                                 time.sleep(2)
                             else:
                                 logger.warning("No new snapshot", self)
-                        else:
-                            ret_error = False
+                        else:  # new snapshot taken...
+                            if ret_error:
+                                logger.error('New snapshot taken but errors detected', self)
+                                self.config.PLUGIN_MANAGER.error(6, sid.displayID)  # Fixes #1491
+                            ret_error = False  # Why ignore errors now?
```
buhtz, line 815: Not sure about this but your arguments sounds fine to me.

buhtz:
Thanks a lot for that work and all the comments and improved docu. This will help a lot in the future.

buhtz (approved):

CHANGES
aryoda: Do you still found the link to the changelog standards (I have lost oversight here).

I would like to add it at the top of the changelog file keep it "in sight" whenever I change the changelog...

common/pluginmanager.py
aryoda: Right, I forgot to update the README.md there!

I will update this too today

CHANGES
buhtz: In my [own projects](https://codeberg.org/buhtz/buhtzology) where I also decided to write [markdown changelog](https://codeberg.org/buhtz/buhtzology/src/branch/develop/CHANGELOG.md) I do use
 - https://common-changelog.org
 - https://keepachangelog.com
 - and slightly related https://www.conventionalcommits.org
 - beside of that I summarize the important aspects and examples in my own Zettelkasten (which is not public available)

I didn't want to bother you folks with that details. And myself I'm also quit new to handling things like this. So I'm still in experimenting period and try out what fits best to me.

While getting in touch with that standards I also read about auto-generating changelogs from git commit history. But I don't like that approach because the target readers for commit messages and changelog messages are IMHO different. The first is more technical based and the latter should focus on behavior of the application.

aryoda: Fixed

common/pluginmanager.py
aryoda: Done

PR #1503, opened 2023-08-17
Fix broken translation in systrayicon.py (NameError: name '_' is not defined)

aryoda:
## Symptoms

Traceback when running a backup (taking a snapshot) via GUI or CLI:

```
  File "/home/<user>/dev/backintime/qt/qtsystrayicon.py", line 38, in <module>
    import logviewdialog
  File "/home/<user>/dev/backintime/qt/logviewdialog.py", line 28, in <module>
    import messagebox
  File "/home/<user>/dev/backintime/qt/messagebox.py", line 55, in <module>
    def info(text, title=_('Information'), widget_to_center_on=None):
NameError: name '_' is not defined
```

## Notes

There is no changelog entry since the bug was introduced on the dev branch only (not in a release) in a commit four days ago...

The original idea to init the translations in `QtSystrayIcon.__init__()`  (where it would belong to)
did not work because a previous `import logviewdialog` already causes this traceback
**before** the ``__init__()`` constructor is called.


qt/qtsystrayicon.py
```diff
@@ -31,7 +31,10 @@
 import qttools
 qttools.registerBackintimePath('common')
 
+# Workaround until the codebase allows a single place to init all translations
 import tools
```
buhtz, line 35: I would move that import before `import qttools` in line 31.

qt/qtsystrayicon.py
```diff
@@ -31,7 +31,10 @@
...
+tools.initiate_translation(None)
```
buhtz, line 36: In this case (because of `None`) the systems local will be used to determine the language. The configured language in BIT is ignored.

See this lines 
https://github.com/aryoda/backintime/blob/25c734393d26f77d9ee9c04fa19426d144bd1f5c/qt/qtsystrayicon.py#L60-L63

```
        tools.initiate_translation(self.config.language())
        self.qapp = qttools.createQApplication(self.config.APP_NAME)
        translator = qttools.initiate_translator(self.config.language())
        self.qapp.installTranslator(translator)
        self.qapp.setQuitOnLastWindowClosed(False)
```

Here I would insert the gettext-initialization with the configured language.

```
        self.qapp = qttools.createQApplication(self.config.APP_NAME)
        translator = qttools.initiate_translator(self.config.language())
        self.qapp.installTranslator(translator)
        self.qapp.setQuitOnLastWindowClosed(False)
```

I'm not sure if I do get all details here. That is why I just do review instead of commiting to your branch.

qt/qtsystrayicon.py
aryoda: This does not work because `import tools` does not find `tools` then because the search path for BiT is registered later (in the line 32 above)

qt/qtsystrayicon.py
aryoda: The current code in `app.py` has the same structure (and is using the system language for all imports until `__main__` is executed), is this a bug then?

https://github.com/bit-team/backintime/blob/976081bbd3e38cee0e7f5efcffe0f7392777e76f/qt/app.py#L31-L37

and in `__main__`:

https://github.com/bit-team/backintime/blob/976081bbd3e38cee0e7f5efcffe0f7392777e76f/qt/app.py#L1945-L1948

aryoda: In other words my dilemma is:

- I cannot use the language from the config file directly after import because the config is not loaded (and cannot because of missing imports):
  `tools.initiate_translation(self.config.language())`

- I cannot move the init code from the `import tools...` to the `__init__` constructor because another import in the same file contains a function signature which already uses `_()` as default value and causes an "compilation" error:

https://github.com/bit-team/backintime/blob/976081bbd3e38cee0e7f5efcffe0f7392777e76f/qt/qtsystrayicon.py#L38
loads...
https://github.com/bit-team/backintime/blob/976081bbd3e38cee0e7f5efcffe0f7392777e76f/qt/logviewdialog.py#L28
which tries to validate
https://github.com/bit-team/backintime/blob/976081bbd3e38cee0e7f5efcffe0f7392777e76f/qt/messagebox.py#L55

- I cannot change the signature of `def info(text, title=_('Information'), widget_to_center_on=None):` because the default value needs to be there and translated

**So I suspect the structure of `app.py` is like my PR (causing the info message box to use the system language instead of the configured one) as best-possible workaround with the smallest impact (= wrong translations).**

I see no easy way to fix this :-(

aryoda: So if I get you right the best work-around is **to add** also this line into my `` constructor:

`tools.initiate_translation(self.config.language())`

I will send another commit ASAP...

PS: I think the fundamental problem is that the translations should be initialized **before** all imports happen to support correct translation of string default values in functions. Since our translation is loaded from a config file which requires heavy imports before this is not easily possible. And I see no simple solution for that and would say, above added line is the best I can do for now...

PS2: Thanks for pointing out the problem, no I understand the translation boot-strapping :-)

aryoda: Sorry for the noise but I think now that nothing must be changed because the language of the config file is already used to enable translation above (line 48 - 49) by loading the config (remember: `Config.__init__()` also calls `language_used = tools.initiate_translation(self.language())`):

https://github.com/bit-team/backintime/blob/976081bbd3e38cee0e7f5efcffe0f7392777e76f/qt/qtsystrayicon.py#L46-L60

So adding 

tools.initiate_translation(self.config.language())

into the line 56 would do the same again (unnecessarily) IMHO.

**So nothing todo but merge as-is?**

buhtz: I might have confused myself... :rofl: 
I'll take a look into it in the next hours to be sure. :face_with_spiral_eyes:

buhtz:
Awesome! I forgot my own solution. Thanks for recover it again.
You are right. Merge it. Forget my review comments.

Some thoughts about `def info(text, title=_('Information'), `. This happens because the systray do import `logviewdialog`. But I see no problem because the translation will take effect only when `info()` is called and not when it is defined (or declared?). The time when `messagebox.info()` is called is far away after `config.Config.__init__()` happens.

Merge it.

buhtz (approved):

PR #1504, opened 2023-08-20
feat: Approach translators dialog message after 10 manual starts of BIT GUI

### LINKED ISSUE #1473: Dialog to approach translators
This is a quick shot idea. 😄 
I would like to present our users this dialog.
translate_dialog drawio (1)

- This is a modal dialog.
- The two buttons "German" and "English" are tab buttons. The current set language (e.g. "German") is presented by default. But the user is also able to view the "source language" English if needed.
- The dialog do not appear on the first start of BIT but after some starts (e.g. 5 times or 10 or more). The reason is that a new user need to get comfortable with BIT and its GUI before doing anything else. Such a dialog on the first start will be experienced as an annoying disturbance. And the "bonding" between BIT and the user is not established on the first start.
- I do have a solution to count the number of BIT starts and differentiate them from scheduled starts.
- I do have a solution to calculate the completeness of the current used language on the local machine without asking the translation platform via a remote connect.
- We could restrict this dialog to languages with a completeness threshold (less then 50%?).

### Alternatives
There are also ideas about alternative ways to present such a message to the user. But I'm not convinced with them. 

For example Thunderbird do use a non modal landing page integrated in its main window.
image
I do use Thunderbird for many years but never read that page. It is IMHO to "silent". And it appears on the first start where I don't want a user to be disturbed with something like this.

Another approach is to have this translation message hidden somewhere in the menus. I will implement a separate "Language selection" menu where this could be place somehow. But I would treat this alternative approach as an addition and not as an replacement to my initial approach. By the way: In such a language selection menu I would also present the completeness of each language in % and have an extra entry like "Help with translations".

### Plain text in German and English (typos included)
```
Hallo.

Sie haben Back In Time nun schon einige Mal genutzt und die Sprache XYZ
eingestellt.

Die Übersetzung von Back In Time in diese Sprache ist nur zu N % vollständig.
Auch ohne besondere technische Kenntnisse, können Sie die Übersetzung
vervollständigen und somit dem Projekt helfen.

Wenn Sie helfen möchten, besuchen Sie bitte die Übersetzungsplattform.
Für weitere Hilfestellung und Fragen besuchen Sie bitte die Website von
Back In Time.

Bitte entschuldigen Sie die Störung. Sie werden diese Nachricht nicht noch
einmal sehen. Ihr Back In Time Team.

--------------------------------------------------------------------------------

Hello.

You have used Back In Time a few times now and set the language to XYZ.

The translation of Back In Time into this language is only N% complete. Even
without special technical knowledge, you can complete the translation and
thus help the project.

If you want to help, please visit the translation platform.
For further assistance and questions, please visit the Back In Time website.

Please accept our apologies for the disruption. You will not see this message
again. Your Back In Time Team.
```

The following terms would appear as hyperlinks
 - translation platform -> https://translate.codeberg.org/engage/backintime
 - Back In Time website -> https://github.com/bit-team/backintime

buhtz:
- The dialog can be opened via the "Translation" menu entry in the "Help" menu.
- It comes up by itself after 10 manual starts of the BIT GUI.
Peek 2023-08-20 22-00

Will close #1473


aryoda:
How can I enable more than the system language and English when running my DEV version of BiT?

image

I would like to do some tests (and admire your new "x % translated" tool tips with my own eyes :smile: )

buhtz:
> How can I enable more than the system language and English when running my DEV version of BiT?

Interesting. How does this happens?

I assume you don't have BIT installed (via configure-make or via apt) on your system. Is this true?

BIT do look into the systems local-folder (/urs/local) using the domain "backintime". The pattern is

    /usr/share/locale/*/LC_MESSAGES/backintime.mo

The wildcard `*` is the language code. The language selection dialog do use all codes it can find via this pattern. This happens [here](https://github.com/bit-team/backintime/blob/ecf9312cdfc87f3417b30e883ee7cc89f0d285d1/common/tools.py#L187).

That is the plan. But maybe something went wrong here.

aryoda:
I could not directly see a reason why only EN is returned as code in my case despite the installation seems to be OK:

```
/usr/share/locale  > find . -name backintime.mo
./nb/LC_MESSAGES/backintime.mo
./en_CA/LC_MESSAGES/backintime.mo
./da/LC_MESSAGES/backintime.mo
./gl/LC_MESSAGES/backintime.mo
./ru/LC_MESSAGES/backintime.mo
./lt/LC_MESSAGES/backintime.mo
./eo/LC_MESSAGES/backintime.mo
./fa/LC_MESSAGES/backintime.mo
./bs/LC_MESSAGES/backintime.mo
./it/LC_MESSAGES/backintime.mo
./hr/LC_MESSAGES/backintime.mo
./jv/LC_MESSAGES/backintime.mo
./sr/LC_MESSAGES/backintime.mo
./vi/LC_MESSAGES/backintime.mo
./th/LC_MESSAGES/backintime.mo
./pt/LC_MESSAGES/backintime.mo
./sv/LC_MESSAGES/backintime.mo
./pt_BR/LC_MESSAGES/backintime.mo
./zh_TW/LC_MESSAGES/backintime.mo
./et/LC_MESSAGES/backintime.mo
./nn/LC_MESSAGES/backintime.mo
./ko/LC_MESSAGES/backintime.mo
./sk/LC_MESSAGES/backintime.mo
./ar/LC_MESSAGES/backintime.mo
./de/LC_MESSAGES/backintime.mo
./tr/LC_MESSAGES/backintime.mo
./fo/LC_MESSAGES/backintime.mo
./hu/LC_MESSAGES/backintime.mo
./is/LC_MESSAGES/backintime.mo
./sl/LC_MESSAGES/backintime.mo
./ja/LC_MESSAGES/backintime.mo
./id/LC_MESSAGES/backintime.mo
./cs/LC_MESSAGES/backintime.mo
./pl/LC_MESSAGES/backintime.mo
./es/LC_MESSAGES/backintime.mo
./he/LC_MESSAGES/backintime.mo
./zh_CN/LC_MESSAGES/backintime.mo
./bg/LC_MESSAGES/backintime.mo
./ca/LC_MESSAGES/backintime.mo
./fr/LC_MESSAGES/backintime.mo
./nl/LC_MESSAGES/backintime.mo
./ro/LC_MESSAGES/backintime.mo
./uk/LC_MESSAGES/backintime.mo
./el/LC_MESSAGES/backintime.mo
./eu/LC_MESSAGES/backintime.mo
./fi/LC_MESSAGES/backintime.mo
./en_GB/LC_MESSAGES/backintime.mo
/usr/share/locale  > echo $LANG
en_US.UTF-8
```

The env vars `LANGUAGE`, `LC_ALL` and `LC_MESSAGES` are NOT set.

I will investigate this next weekend (no prio ATM)...

buhtz:
Interesting. Please `cd` into `common`, start a Python interpreter. Import `tools` and call `tools.get_available_language_codes()`.

grafik

I added some debug output to the function here in this branch. Maybe this will help.

emtiu:
> Interesting. Please `cd` into `common`, start a Python interpreter. Import `tools` and call `tools.get_available_language_codes()`.

I get the same almost-empty dialog as @aryoda on a fresh VM with the dev branch installed.

I checked out this PR's branch and ran the commands you suggested. The output is the following:
```
>>> import tools
>>> tools.get_available_language_codes()
gettext.find() :: po=None
as Path: po=PosixPath('/usr/share/locale/xy/LC_MESSAGES/backintime.po')
rglob() ::
        _GETTEXT_LOCALE_DIR=PosixPath('/usr/share/locale')
        po=PosixPath('*/LC_MESSAGES/backintime.po')
pofiles=<generator object Path.rglob at 0x7fe6baffc9d0>
[]
>>> 
```

buhtz:
OK, I was able to reproduce it and found two problems.

The situation aryoda described might come from the fact that the function searched fro `*.po` instead of `*.mo` files. Fixed it.

The situation from emtiu might occur because you did not installed BIT on your system. You have to do `./configure && make && sudo make install` in `common` (and in `qt`) folder. This will "compile" the `*.po` files into `*.mo` files and copy them into the systems designated folder.

I left the debug-print-statements in the code for further diagnosis.

emtiu:
Whatever the problem was, it works now after your latest commit :)

aryoda:
@buhtz With your current commits (but without updating it with head of our dev) this PR does now show all languages in my machine:

image

BTW: The "cancel" button is missing ;-)

qt/languagedialog.py
```diff
@@ -39,12 +39,16 @@ def __init__(self, used_language_code: str, configured_language_code: str):
         new_width = self._calculate_scroll_area_width()
         self._scroll.setMinimumWidth(new_width)
 
-        button = QDialogButtonBox(QDialogButtonBox.Apply, self)
-        button.clicked.connect(self.accept)
+        buttonbox = QDialogButtonBox(
+            QDialogButtonBox.Cancel | QDialogButtonBox.Apply, self)
+
+        # ApllyRole emit "clicked" instead of "accpted"
```
aryoda: little typos ("ApplyRole" + "accepted")

emtiu:
```suggestion
        # ApplyRole emits "clicked" instead of "accepted"
```

buhtz:
- OK the language selection dialog now use an Apply and Cancel button.
- I finalized the text with removing the linebreaks and let Python's `textwrap` package do the job.

After this is merged I will update the translation strings (#1505).

buhtz:
btw: Here you see only "Website" and "German" translated because this BIT runs in a German locale but there is no translation for the main text yet.

image

Might we add some bold or italic formats?

image

Italic looks IMHO boring.

image

And another flavour trying to imitate the URL formats with using underline tag.
image

emtiu:
I think the bold-only formatting is the most helpful :+1:

However, seeing this message … maybe we should implement a "cutoff" for translation completeness? Only show this dialog if the translation is <97% complete? (any sensible value from 95% to 99%). … Otherwise, it's an annoyance that doesn't really address a pressing problem.

Or, still more sensible: Never show the dialog for almost-complete languages (ca. 95–99%), and never show it for German (because we can always do that ourselves ;)).

buhtz:
> However, seeing this message … maybe we should implement a "cutoff" for translation completeness? Only show this dialog if the translation is <97% complete?

Ah, yes thought of that, too. But forget. I thought more about 90 %. But I did some calculations.
We do have [326 translatable strings](https://translate.codeberg.org/engage/backintime/) currently on Weblate. 90 % would be 33 missing strings. That is IMHO to much.  With 95 % we would allow 17 missing strings; 97 % is 10 missing.
Yes, I also think 97% is a good cutoff.

buhtz:
Currently using the bold-only test.
And added a cutoff 97. But the dialog can be opened independent from the cutoff when using the Help menu entry.

emtiu:
Okay. With a 97% cutoff, the dialog will probably never be shown for German, because we'll always write those strings ourselves before a release. Very good :)

aryoda:
Just a few usability observations:

1. The menu item is called "**Setup language**" while the dialog has the title "**Language selection**".
     Perhaps the same imperative could be used, e.g. "Select language..."

2. The "Language selection" dialog has a different button order for "Cancel" than e.g. the "manage profiles" dialog:

image

image

Also we use "OK" and "Apply" for almost the same semantics here (confusing: "Apply" does not immediately apply the change but requires a restart of the GUI ;-)

3. _Off-topic:_ The "Manage profiles" dialog has another window title than its menu item ("**Settings**" instead of "**Manage profiles**")

buhtz:
>     1. The menu item is called "**Setup language**" while the dialog has the title "**Language selection**".
>        Perhaps the same imperative could be used, e.g. "Select language..."

OK, I'll fix it.

>     2. The "Language selection" dialog has a different button order for "Cancel" than e.g. the "manage profiles" dialog:

I'll check that. Both are create with a QDialogButtonBox. It seems that the [button order is platform specific](https://stackoverflow.com/a/35907926/4865723) and different because of Apply and OK.

> Also we use "OK" and "Apply" for almost the same semantics here (confusing: "Apply" does not immediately apply the change but requires a restart of the GUI ;-)

Ah! :guitar: The first time I really get the difference between these two buttons. :rofl: Yes, OK makes more sense.

>     3. _Off-topic:_ The "Manage profiles" dialog has another window title than its menu item ("**Settings**" instead of "**Manage profiles**")

I'll add this to this PR, too.

buhtz:
Ähm... I messed up with git(hub) having my forked "dev" branch 1 commit ahead of "upstream". I played arround with another "format" of issue templates and needed to commit to (my forked) "dev" to the the results. Don't see an elegant way to fix it. Lot of error messages on my local machine. Currently I'm not able to commit to this branch or "dev" or do something else.

The most clean way for me would be to delete my fork on github and re-fork it again. Then it should be clean.

Not sure if this affects PRs. Because of that I would merge this PR (no matter that there are minor problems) before.
When it is merged I can do my thing and add these minor fixed in a new PR.

```
index 2afa7eb1..faae296c 100644
--- a/common/tools.py
+++ b/common/tools.py
@@ -200,7 +200,6 @@ def get_available_language_codes():

     if mo:
         mo = pathlib.Path(mo)
-        print(f'{mo=} from Path')
     else:
         # Workaround. This happens if LC_ALL=C and BIT don't use an explicite
         # language. Should be re-design.
diff --git a/qt/app.py b/qt/app.py
index 51af52db..9e49b600 100644
--- a/qt/app.py
+++ b/qt/app.py
@@ -1771,8 +1771,9 @@ files that the receiver requests to be transferred.""")

         dlg.exec()

+        print(f'{self.config.language()=} {dlg.language_code=}')
         # Apply/OK pressed & the language value modified
-        if dlg.result() == 1 and self.config.language != dlg.language_code:
+        if dlg.result() == 1 and self.config.language() != dlg.language_code:

             self.config.setLanguage(dlg.language_code)

diff --git a/qt/languagedialog.py b/qt/languagedialog.py
index 780027f2..c25b19c0 100644
--- a/qt/languagedialog.py
+++ b/qt/languagedialog.py
@@ -41,7 +41,7 @@ class LanguageDialog(QDialog):
         self._scroll.setMinimumWidth(new_width)

         buttonbox = QDialogButtonBox(
-            QDialogButtonBox.Cancel | QDialogButtonBox.OK, self)
+            QDialogButtonBox.Cancel | QDialogButtonBox.Ok, self)

         buttonbox.accepted.connect(self.accept)
         buttonbox.rejected.connect(self.reject)  
```

aryoda:
> I messed up with git(hub) having my forked "dev" branch 1 commit ahead of "upstream". I played arround with another "format" of issue templates and needed to commit to (my forked) "dev" to the the results.

Stash your changes and try this on `dev`:

```
# undo last (one!) commit and put changes back into the working directory as unstaged changes
git reset --mixed HEAD~1 
```

> Because of that I would merge this PR...

Totally OK, easiest way :-)

buhtz:
Even stashing is not possible anymore. The "tree" is corrupt. Not sure what this means.

But I got another idea. I will merge this PR not into "dev" but into an extra (temporary) branch where I can add me minor fixes.

Issue #1506, opened 2023-08-23
The rsync 24 return code should not trigger a BIT failure 


jean-christophe-manciot:
Using BIT ecf9312, I noticed that some snapshots fail and are consequently cancelled due to an ```rsync warning```:
```
[I] Take snapshot (rsync: total size is 46.24G  speedup is 13.62)
[I] Take snapshot (rsync: rsync warning: some files vanished before they could be transferred (code 24) at main.c(1338) [sender=3.2.7])
[E] Error: 'rsync' ended with exit code 24 (negative values are signal numbers, see 'kill -l')
[E] Failed to take snapshot 2023-08-20 00:31:55.
```
It can be part of a normal live linux system to create some files temporarily before the backup starts and discard them before the end of the snapshot.
This warning should not cancel the entire backup.

I propose 2 alternatives:
- dismiss this specific rsync return code and let the snapshot complete normally (preferred solution)
- or add an option in the "expert options" of the settings to dismiss this 24 return code

[bit-diags.txt](https://github.com/bit-team/backintime/files/12418000/bit-diags.txt)

aryoda:
THX for testing and reporting this issue!

I think I will implement alternative 1 by logging the exit code 24 as `[I]` (info) only in the snapshot log instead of treating it as error.

I have also just checked the list of `rsync`'s exit codes to find other candidates (beyond `24`) which should NOT be treated as an error but just as a warning or info in BiT and have found no more additional exit codes:

```
EXIT VALUES
 0 - Success
 1 - Syntax or usage error
 2 - Protocol incompatibility
 3 - Errors selecting input/output files, dirs
 4 - Requested action not supported. Either:
     - an attempt was made to manipulate 64-bit files on a platform that cannot support them
     - an option was specified that is supported by the client and not by the server
 5 - Error starting client-server protocol
 6 - Daemon unable to append to log-file
10 - Error in socket I/O
11 - Error in file I/O
12 - Error in rsync protocol data stream
13 - Errors with program diagnostics
14 - Error in IPC code
20 - Received SIGUSR1 or SIGINT
21 - Some error returned by waitpid()
22 - Error allocating core memory buffers
23 - Partial transfer due to error
24 - Partial transfer due to vanished source files
25 - The --max-delete limit stopped deletions
30 - Timeout in data send/receive
35 - Timeout waiting for daemon connection
```

[...]

PR #1510, opened 2023-08-28
Fix bug #1506: Treat rsync exit code 24 as INFO instead of ERROR

aryoda:
> Don't forget the changelog entry! ;)

Done :-)

@buhtz I hope the `gettext` translation does still work with my heavy string concatenation (required to avoid duplicated sub strings/prefixes in the messages)


common/snapshots.py
```diff
@@ -1241,11 +1242,27 @@ def takeSnapshot(self, sid, now, include_folders):
...
+        # (not as ERROR). The values are message strings for the snapshot log.
+        rsync_non_error_exit_codes = {
+            0: _("Success"),
+            24: _("Partial transfer due to vanished source files (see 'man rsync')")
+        }
+
+        rsync_exit_code_msg = _("'rsync' ended with exit code {exit_code}".format(exit_code=rsync_exit_code))
+
+        if rsync_exit_code in rsync_non_error_exit_codes:
+            self.setTakeSnapshotMessage(0,
```
buhtz, line 1255:
```
self.setTakeSnapshotMessage(0,
                            rsync_exit_code_msg + _(": {msg}".format(msg=rsync_non_error_exit_codes[rsync_exit_code])))
```

I'm not sure what your goal is here. The `format()` is executed **before** the translation.
The translator on Weblate won't get this string

    ": {msg}"

but something like this if me as a maintainer with German local on his developer machine did managed the translations.

    ": Teilweise Übertragung aufgrund von verschwundenen Quelldateien (siehe 'man rsync')"

Because of line 1248 and 1249 the message is still translated and then injected into `": {msg}".

To my understanding there is no need for translation in this line. Make it like this:

```
self.setTakeSnapshotMessage(
    0,
    rsync_exit_code_msg + ": {msg}".format(msg=rsync_non_error_exit_codes[rsync_exit_code]))
```

common/snapshots.py
```diff
@@ -1241,11 +1242,27 @@ def takeSnapshot(self, sid, now, include_folders):
...
+                                        rsync_exit_code_msg + _(": {msg}".format(
+                                            msg=rsync_non_error_exit_codes[rsync_exit_code])))
+        elif rsync_exit_code > 0:  # indicates a rsync error
+            params[0] = True  # HACK to fix #489 (params[0] and has_errors should be merged)
+            self.setTakeSnapshotMessage(1,
+                                        rsync_exit_code_msg + _(": See 'man rsync' for more details"))
```
buhtz: Here I would exclude the `: ` from the string to translate. Translators on Weblate can overlook and miss it.

Modify this line

    rsync_exit_code_msg + _(": See 'man rsync' for more details"))

to this

    rsync_exit_code_msg + ": " + _("See 'man rsync' for more details"))

common/snapshots.py
```diff
@@ -1241,11 +1242,27 @@ def takeSnapshot(self, sid, now, include_folders):
...
+        elif rsync_exit_code < 0:  # indicates a rsync error caused by a signal
             params[0] = True  # HACK to fix #489 (params[0] and has_errors should be merged)
             self.setTakeSnapshotMessage(1,
-                                        _("Error: 'rsync' ended with exit code {exit_code} (negative values are signal numbers, see 'kill -l')".format(
-                                            exit_code=rsync_exit_code)))
+                                        rsync_exit_code_msg + _(": Negative rsync exit codes are signal numbers, see 'kill -l' and 'man kill'"))
```
buhtz: The same here

    rsync_exit_code_msg + ": " + _("Negative rsync exit codes are signal numbers, see 'kill -l' and 'man kill'"))

buhtz (changes requested):

common/snapshots.py
aryoda: No I do see it too, will fix this :-)

aryoda:
> The format() is executed before the translation.

So which string does a translator see on Weblate for this code in line 1252:

```
 rsync_exit_code_msg = _("'rsync' ended with exit code {exit_code}".format(exit_code=rsync_exit_code))
```

 1. 'rsync' ended with exit code {exit_code}
  
 or

2. 'rsync' ended with exit code 24

I was hoping the translator does see 1.

If the translator sees 2. I am just curious how gettext/Weblate is able to extract all strings from the source code incl. applying all possible formats first (without executing the code with all possible values)!

buhtz:
Yes, this topic can make headaches. 😄 

> > The format() is executed before the translation.
> 
> So which string does a translator see on Weblate for this code in line 1252:
> 
> ```
>  rsync_exit_code_msg = _("'rsync' ended with exit code {exit_code}".format(exit_code=rsync_exit_code))
> ```
> 
> 1. 'rsync' ended with exit code {exit_code}
> 
> or
> 
> 2. 'rsync' ended with exit code 24
> 
> I was hoping the translator does see 1.

The translator on weblate will see nothing because the gettext utilities (e.g. `xgettext`) will ignore that line. They ignore it because it is not a string only. Tools like `xgettext` looking for patterns like `_("...")`. But you have `_("...".format(foo))` which is not only a string.

If you want number 1 then you have to give exact that string to `_()` and nothing else.

    _("'rsync' ended with exit code {exit_code}")

But you will fill in the variable `exit_code` when BIT code is executed right? Just add `.format()` to the translated string returned by `_()`.

    _("'rsync' ended with exit code {exit_code}").format(exit_code=rsync_exit_code)

The `_()` is a function taking a string and return another one. In German it will return `"'rsync' wurde mit exit code {exit code} beendet"`. So when the code is executed by Python with German local the `format()` is practically something like this:

    "'rsync' wurde mit exit code {exit code} beendet".format(exit_code=rsync_exit_code)

> If the translator sees 2. I am just curious how gettext/Weblate is able to extract all strings from the source code incl. applying all possible formats first (without executing the code with all possible values)!

The gettext tools (to my knowledge) all working on text-file-parsing. They do not execute code. Your example is ignored because it is not just a string.

I see no need for "applying all possible formats first". The opposite should be the case. Strings should be translated **before** applying `format()`. The all possible formats should be applied **after** translation.

aryoda:
@buhtz THX a lot, static code analysis (searching for plain strings) is the simple magic in providing strings to Weblate

So if I understand you right I should change line 1252 from

```
rsync_exit_code_msg = _("'rsync' ended with exit code {exit_code}".format(exit_code=rsync_exit_code))
```
to

```
rsync_exit_code_msg = _("'rsync' ended with exit code {exit_code}").format(exit_code=rsync_exit_code)
```

so that translators can rearrange the position of the `{exit_code}` variable in the translated text according to the target language?

buhtz:
> So if I understand you right I should ...

Correct.

aryoda:
OK, all translations should now be enabled correctly

buhtz (approved):

PR #1511, opened 2023-08-29
second try: "feat: Approach translators dialog message after 10 manual starts of BIT GUI"

buhtz:
This is a follow up to PR #1504 which wasn't merged into `dev` but into a temporary branch.

This PR here do merge #1504 and some minor modifications.


aryoda:
Sorry, I don't want to be pedantic, but I think "OK" is always the right-most button
(otherwise users may loose changes by quickly clicking the right-most button because they are used to this UX).

Furthermore I get a traceback now when I open

    Help > Translation

```
  File "/home/user/dev/backintime/qt/app.py", line 1798, in slot_help_translation
    self._open_approach_translator_dialog()
  File "/home/user/dev/backintime/qt/app.py", line 1767, in _open_approach_translator_dialog
    name, perc = tools.get_native_language_and_completeness(code)
  File "/home/user/dev/backintime/common/tools.py", line 278, in get_native_language_and_completeness
    completeness = languages.completeness[language_code]
KeyError: 'en'
```

Can you reproduce this?

buhtz:
> Sorry, I don't want to be pedantic

I hope you always be! 😄 

> but I think "OK" is always the right-most button (otherwise users may loose changes by quickly clicking the right-most button because they are used to this UX).

I have no feeling or strong opinion about this but regarding to "the internet" the order of such buttons is different between operating systems and even desktop environments. Currently Qt handles the ordering in BIT by itself via the `QDialogButtonBox`. On KDE the order would be OK - Cancel, on Gnome Cancle - OK. I never noticed differences like this.

I don't know if it is a good idea to work against the standards.

> Furthermore I get a traceback now when I open
> ...
> KeyError: 'en'
> 
> Can you reproduce this?

Very good. I can reproduce. It is because "en" is not a language in the meaning of gettext it is just our python source.
I think I fixed this.

 - "en" is no recognized always as 100 complete translated
 - The approachtranslatorsdialog won't show up on systems using "en" on the end of the countdown.
 - But the dialog shows up when using Help -> Translation.

aryoda:
> On KDE the order would be OK - Cancel, on Gnome Cancle - OK

Yes, there is no global standard (I was overkeen by saying "always" ;-)

We just have to stick to one standard in all our Dialogs otherwise user may click on the wrong button because it is at an unexpected place.

buhtz:
I assume there was a missunderstanding. It is fixed now, isn't it?

image

The "default order" QDialogButtonBox do use for Ok and Cancle is different from Apply and Cancel. In my desktop environment (Xfce) it is `[Cancel] [OK]` and `[Apply] [Cancel]`. My modification was only that I replaced `QDialogButtonBox.Apply` with `QDialogButtonBox.Ok`, nothing more. The implicit effect is that the order now is fixed.

aryoda:
> I assume there was a missunderstanding. It is fixed now, isn't it?

Ah, yes it is fixed (I was confused because Qt5 on Gnome is still using a different buttom order but it is consistent now :-)

image

Ready to be merged now IMHO...

aryoda (approved):

buhtz:
Closing #1473

PR #1512, opened 2023-08-29
translation: update (2023-08-29)

buhtz:
Synchronized translations on Weblate between modified source strings in the upstream repositories py-files.

Additionally added a quick-step-guide how to do this to our doc-dev section.

Improved the update_language_files.py script to produce "sorted" content for the po(t) files. That should make it easier to make "diff" between commits.


buhtz:
@aryoda I realized you introduced new translatable strings in #1510 . So I will wait with this PR here until you merged yours.

buhtz:
It seems to me that updating the pot and po files everytime shuffles the order of the stirngs in that files. That produce messy diff outputs. I modified the update script that way that the output now should be ordered by-file hoping this will produce cleaner diff message next time.

buhtz:
My assumption was wrong. The "textwrap" package to break inside words if they are to long. So there is no problem.

~This is a never ending story.... 😆
The current code in `ApproachTranslatorsDialog` do use Pythons own `textwrap` package to insert linebreaks in the text to make it look nicer in the resulting dialog. The package do recognize blank/space characters between words. My goal was not not bother the translators on weblate handling `\n` characters.
But there are languages that don't have blanks between words.
https://en.wikipedia.org/wiki/Line_breaking_rules_in_East_Asian_languages
The line break rules are quit easy but implementing it ourselfs just for one dialog is IMHO to much. To my current knowledge there is no Python package that is able to handle Japanese, Chinese, Korean (and more?). There is a "package" from Google that do connect to an extern service exposing our users to Google. Not on option of course.
I can detect the language and then decide if "textwrap" or an alternative mode should be used. The "alternative mode" would just be to insert `\n` each 65 characters and hopping it looks nice and understandable to the native speakers. ;)~

aryoda:
> @aryoda I realized you introduced new translatable strings in #1510 . So I will wait with this PR here until you merged yours.

@buhtz PR #1510 is merged now into DEV, you can proceed

aryoda (approved):

buhtz:
Hui... Translating a multi-paragraph text and dealing with dynamic line wrapping and languages having double-width characters (east Asian languages) is a "fascinating topic". :rofl: The "solution" here is not perfect and not beautiful in all cases but pragmatic.
I have a lot of ideas how to improve the visual result but this would produce a lot of more code and cost performance. The cost-benefit ratio wouldn't be so good.

Good news: The Japanese version of the dialog was translated by and approved by a native speaker with expertise in translation.

Issue #1517, opened 2023-09-01
Backup locks-up system


sackwort:
Since moving to MX Linux three months ago, whenever Back In Time is performing a backup the system locks-up. Say a backup takes four minutes, every 20 seconds or so I can't click anything or operate the desktop, or if I'd already clicked something it doesn't take effect for ten seconds or so.

I've been using Back In Time for years on various Linus distros without this issue.

Could this be a Debian issue? If so, I'd be grateful of any tips/advice.

emtiu:
Can you tell us what drive/filesystem/server the backups are written to/from? Your problem description sounds a bit like I/O delay, maybe from a slow external hard-drive, or a busy network connection.

sackwort:
Thanks for your quick reply. Sorry, I should've mentioned this.  I'm using an external USB-3 HDD to backup to, which is formatted EXT4 as is my PC. The external drive is not a fast one, but I've never had this issue on any other Linux desktop.

I also should have mentioned that I don't get this issue when doing a restore.

buhtz:
Can you please give us some more version information.

Which version of backintime (BIT) do you use currently?
How and from which source did you install it? I wasn't able to find "backintime" on MX Linux website.

Which version of BIT on which GNU Linux distribution did you use before when you got no issues?

sackwort:
I've attached the output of "backintime --diagnostics", which is something else I forgot to post. Sorry.

I installed it on MX via the Terminal ("sudo apt install backintime-qt").

As for using it on other distros, I've always used the latest available from their repositories. I can't really remember what versions, to be honest. On occasion, I've also used the PPA.
[backintime-diagnostics.txt](https://github.com/bit-team/backintime/files/12498118/backintime-diagnostics.txt)

```
{
    "backintime": {
        "name": "Back In Time",
        "version": "1.3.3",
        "latest-config-version": 6,
        "local-config-file": "/home/UsernameReplaced/.config/backintime/config",
        "local-config-file-found": true,
        "global-config-file": "/etc/backintime/config",
        "global-config-file-found": false,
        "started-from": "/usr/share/backintime/common",
        "running-as-root": false,
        "user-callback": "/home/UsernameReplaced/.config/backintime/user-callback",
        "keyring-supported": true
    },
    "host-setup": {
        "platform": "Linux-6.1.0-11-amd64-x86_64-with-glibc2.36",
[...]

PR #1518, opened 2023-09-01
feat: Diagnostic output using alternative os-release infos

buhtz:
Because of #1517 I realized that GNU Linux distros have a lot of variants to identify them self. Sometimes they do not follow the standards and sometimes they violate them (like MX Linux in #1517 identifing itself as "Debian").

In addition to the former behavior BIT now add information from `/etc/*release` files if they are present.
I removed the use of `platform.freedesktop_os_release()` (added in Python 3.10) because it does nothing more then reading the /etc/os-release file what we can do our self.

Here is an output example of content from files I created myself because Debian do offer a "os-release" file.

image

In relation to #1517 on MX Linux the output would look like this:

```
"host-setup": {
    "OS": {
        "os-release": "Debian GNU/Linux 11 (bullseye)",
        "/etc/lsb-release": "MX Linux 21.3 etc pp",

```

### DIFF
--- a/common/diagnostics.py
+++ b/common/diagnostics.py
@@ -381,10 +381,11 @@ def _get_os_release():
 
     # read and parse the os-release file (first)
     fp = etc_path / 'os-release'
+
+    osrelease = {'os-release': _get_pretty_name_or_content(fp)}
+
     if fp in os_files:
         os_files.remove(fp)
-        osrelease = {
-            'os-release': _get_pretty_name_or_content(fp)}
 
     # alternative release files
     for fp in os_files:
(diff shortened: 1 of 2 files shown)


common/diagnostics.py
```diff
@@ -340,28 +340,58 @@ def get_git_repository_info(path=None):
...
+    osrelease = {
+        'os-release': _get_pretty_name_or_content(fp)}
 
-    try:
-        with fp.open('r') as handle:
-            osrelease = handle.read()
-    except FileNotFoundError:
-        return '(os-release file not found)'
+    # alternative release files
+    for fp in Path('/etc').glob('*release'):
```
emtiu: You might also include /etc/*version here, because some distributions use that.

PR #1520, opened 2023-09-04
Update README.md

Slater91:
Reviewed some wording at the very end of the readme: "older then" is incorrect, and "the last stable release" implies there won't be another one so probably "latest" should be used instead.

Thanks for your work! :)


aryoda (approved):
THX a lot for making it so easy for us to fix this via your PR :-)

PR #1521, opened 2023-09-05
fix: Check SSH login works on machines with limited commands (#1442)

### LINKED ISSUE #1442: How to use ssh if remote host does not support echo command (Hetzner Storage Box)
Do you have any idea how I can use backintime if rsync and standard file operation commands are available but no echo command which backintime is using to check the connection?

`ssh -o PreferredAuthentications=publickey -p 22 -o ServerAliveInterval=240 -o IdentityFile=/home/[user]/.ssh/[id]_rsa backup@[my-local-ip] echo "Hello"`

fails with 
Command not found. Use 'help' to get a list of available commands.

`ssh -o PreferredAuthentications=publickey -p 22 -o ServerAliveInterval=240 -o IdentityFile=/home/[user]/.ssh/[id]_rsa backup@[my-local-ip] ls`

works fine.

buhtz:
Two functions checking login and cipher on a SSH remote host now using `exit` command instead of `echo`. This should work on machines with limited commands (e.g. Hetzer storage boxes) also.

Fix #1442

### DIFF
--- a/common/config.py
+++ b/common/config.py
@@ -385,7 +385,7 @@ class Config(configfile.ConfigFileWithProfiles):
                 if path == snapshots_path:
                     self.notifyError(
                         '{}\n{}'.format(
-                            _('Profile: "{name}"').format(name=f'"{profile_name}"'),
+                            _('Profile: "{name}"').format(name=profile_name),
                             _("Backup folder cannot be included.")
                         )
                     )
--- a/common/config.py
+++ b/common/config.py
@@ -397,7 +397,7 @@ class Config(configfile.ConfigFileWithProfiles):
                         self.notifyError(
                             '{}\n{}'.format(
                                 _('Profile: "{name}"').format(
-                                    name=f'"{self.currentProfile()}"'),
+                                    name=self.currentProfile()),
                                 _("Backup sub-folder cannot be included.")
                             )
                         )
--- a/common/config.py
+++ b/common/config.py
@@ -47,7 +47,7 @@ except ImportError:
 # The bigger problem with config.py is that it do use translatebale strings.
 # Strings like this do not belong into a config file or its context.
 try:
-    _('Cancel')
+    _('Warning')
 except NameError:
     _ = lambda val: val
 
@@ -385,7 +385,7 @@ class Config(configfile.ConfigFileWithProfiles):
                 if path == snapshots_path:
                     self.notifyError(
                         '{}\n{}'.format(
-                            _('Profile: {name}').format(name=f'"{profile_name}"'),
+                            _('Profile: "{name}"').format(name=f'"{profile_name}"'),
                             _("Backup folder cannot be included.")
                         )
                     )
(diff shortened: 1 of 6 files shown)


common/config.py
```diff
@@ -385,7 +385,7 @@ def checkConfig(self):
                 if path == snapshots_path:
                     self.notifyError(
                         '{}\n{}'.format(
-                            _('Profile: {name}').format(name=f'"{profile_name}"'),
+                            _('Profile: "{name}"').format(name=f'"{profile_name}"'),
```
aryoda: This looks strange:
_{name}_ is already surrounded by quotation marks
but _format()_ uses an f-string which again adds quotation marks.

Is it really intended to create two surrounding quotation marks?

common/config.py
```diff
@@ -396,7 +396,7 @@ def checkConfig(self):
                     if path[: len(snapshots_path2)] == snapshots_path2:
                         self.notifyError(
                             '{}\n{}'.format(
-                                _('Profile: {name}').format(
+                                _('Profile: "{name}"').format(
                                     name=f'"{self.currentProfile()}"'),
```
aryoda: same comment as above regarding duplicated quotation marks...

qt/settingsdialog.py
```diff
@@ -1009,14 +1009,14 @@ def __init__(self, parent):
         # additional rsync options
         hlayout = QHBoxLayout()
         layout.addLayout(hlayout)
+        tooltip = _('Options must be quoted e.g. {example}.').format(
```
aryoda, line 1012: Is quotation required for the option name or option value or both?
The example looks like quotation is required for the option value only.
In this case I would clarify this in the tooltip, eg. via

'Option values must be quoted, e.g. {example}'

aryoda:
> Two functions checking login and cipher on a SSH remote host now using `exit` command instead of `echo`. This should work on machines with limited commands (e.g. Hetzer storage boxes) also.

I have tested using `exit` instead of `echo` with a profile accessing ssh on my Synology NAS and it did work.

Is there a way for a negative test to see if `exit` may fail at all?

**Edit:** I did two trivial negative tests by changing the user name and then the host name to see if the failing login is recognized and it did work:

`ssh -o ServerAliveInterval=240 -o LogLevel=Error -o IdentityFile=/home/<user>/.ssh/id_rsa -o PreferredAuthentications=publickey -p 22 username@my.ssh.com exit`

Clever solution :+1:

qt/settingsdialog.py
buhtz: No that is OK that way. The result in GUI (Settings -> Export Options, with SSH profile).

grafik

common/config.py
buhtz: Yeah, you are right. That is chaos.

The question is if profile names need to be quoted or not?

If this should be part of that translatable string is another question.

buhtz:
In the beginning I tried to remove quotation marks from translatable strings to not confuse translators.
But I realized that even quotation marks need to be translated. For example in Japanese you would write

    _('「{name}」')

I will take this into account when reworking the source strings next time. I won't do this before the release because we would lose some translations.

common/config.py
buhtz: Stepped into my own trap...

PR #1522, opened 2023-09-06
feat: Update translation

buhtz:
Just an update of translations out of the line.

I got an icelandic translation via email because the user had problems with the register and login process at Weblate. This is known problem and on the todo list of the admins there.


buhtz:
Merging this immediately because I need to "Unlock" the Weblate project again.

PR #1525, opened 2023-09-07
fix: Translate not before language setup is done

### LINKED ISSUE #1523: Translation happens before setup language takes effect
The OS used language and the explicit language setup in BIT are mixed in the GUI.

On a none-English OS explicit setup a language (e.g. "English") in BIT. You will see strings in the GUI translated with the OS current local (e.g. German) instead of the language setup in BIT (e.g. English).

image

I know why this happens and it is not much to fix. Sorry but I would like to fix this before 1.4.0. 😄

buhtz:
Translations via gettext starting after the language setup is finished and using the setup language. It is relevant when a language was setup in BIT different from the operating systems used local.

Technically class variables from Config class are moved into the instance of Config or elsewhere.

Because it is a minor but not related to an existing release I decided not to make a changelog entry.

Fix #1523

### DIFF
--- a/common/config.py
+++ b/common/config.py
@@ -95,24 +95,6 @@ class Config(configfile.ConfigFileWithProfiles):
     DISK_UNIT_MB = 10
     DISK_UNIT_GB = 20
 
-    REMOVE_OLD_BACKUP_UNITS = {
-                DAY: _('Day(s)'),
-                WEEK: _('Week(s)'),
-                YEAR: _('Year(s)')
-                }
-
-    REPEATEDLY_UNITS = {
-                HOUR: _('Hour(s)'),
-                DAY: _('Day(s)'),
-                WEEK: _('Week(s)'),
-                MONTH: _('Month(s)')
-                }
-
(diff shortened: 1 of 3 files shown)


common/config.py
```diff
@@ -397,7 +379,7 @@ def checkConfig(self):
                         self.notifyError(
                             '{}\n{}'.format(
                                 _('Profile: "{name}"').format(
-                                    name=self.currentProfile()),
+                                    name=profile_name),
```
aryoda, line 382: Good change to use the name instead of the ID!

I personally even love to see not only the name  but also the ID (eg. "Main Profile [1]")
which helps eg. in screen shots in issues...

aryoda (approved):

common/config.py
buhtz: Good idea. But can we keep this in a separate issue?
There are a lot of code locations where the profile name/id is shown.

aryoda: Yes of course, no change required (we could use a bigger GUI refactoring for this)

PR #1530, opened 2023-09-12
trans: Update translations

### LINKED ISSUE #1505: Final update of translations before release
Before release make a final update of translations from weblate.

buhtz:
Sync translations with Weblate translation platform.

Additionally fix a minor issue about translating the default title of an info-messagebox before the setup language came into account.

Fix #1505


buhtz:
Just ignore the problem with the one "ppc64le" machine on TravisCI. This is a known problem on the side of Travis.

common/po/messages.pot
aryoda, line 1: Is it intentional that this file ending is *.pot instead of *.po (like the other ones)?

aryoda (approved):

common/po/messages.pot
buhtz: Yes. The "T" is for "Template". You won't find a translation in a pot-file. All other po-files are related to a specific language. And the po-files are updated based on the pot file.

More detailed steps:
 - Scanning the py-files for translatable strings results in a new or updated pot-file (messages.pot). We use `xgetext` for that).
 - Then the po-files are updated based on the pot-file. We use `msgmrg` for that.

Technically you can name that files how ever you want.

PR #1544, opened 2023-09-25
build: Ignore WARNING in specific unittest

### LINKED ISSUE #1539: More robust unit tests: Strict checking of output as expected causes many false positives
## Background

Several unit tests do check the captures console output  ("actual") against the expected output.

## Impact

This is not very robust (causes many false positives due to unexpected warnings mainly caused by different external setup constellations).

## Current implementation

Currently well known "false positive" warnings are filtered out from the actual console output:

https://github.com/bit-team/backintime/blob/dev/common/test/test_backintime.py#L143-L150

## Proposals

Either

- filter out all warnings in a generic way
- or switch from console output checking to another method (eg. comparing the created snapshot instead)

## References

See also the side discussion in this PR which added another tolerated warning:

https://github.com/bit-team/backintime/pull/1537#issuecomment-1725050367

buhtz:
As discussed in #1539 the test `TestBackInTime::test_local_snapshot_is_successful()` now do ignore all log entries beginning with `Warning` or `WARNING`.

Also slightly improved SSH Server related testing code in `generic.py` and added a FAQ entry for it. The background is that I wasn't able to get my SSH tests run. The reasons was that I do use none-RSA keys.

Fix #1539 
IMHO no need for a CHANGELOG entry.

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -30,6 +30,9 @@
   * [Which additional features on top of a GUI does BIT provide over a self-configured rsync backup? I saw that it saves the names for uids and gids, so I assume it can restore correctly even if the ids change. Great! :-) Are there additional benefits?](#which-additional-features-on-top-of-a-gui-does-bit-provide-over-a-self-configured-rsync-backup-i-saw-that-it-saves-the-names-for-uids-and-gids-so-i-assume-it-can-restore-correctly-even-if-the-ids-change-great---are-there-additional-benefits)
   * [How to move snapshots to a new hard-drive?](#how-to-move-snapshots-to-a-new-hard-drive)
   * [How to move large directory in source without duplicating backup?](#how-to-move-large-directory-in-source-without-duplicating-backup)
+- [Testing & Building](#testing--building)
+  * [SSH related tests are skipped](#ssh-related-tests-are-skipped)
+  * [Setup SSH Server to run unit tests](#setup-ssh-server-to-run-unit-tests)
 <!-- TOC end -->
 ## Restore
 
@@ -901,7 +904,7 @@ moving the file/folder in the last snapshot, too.
 Some of the tests need an available SSH server.
 They get skipped if this is not the case.
 Plese see
-[Setup SSH Server to run unit tests](#setup_SSH_Server_to_run_unit_tests).
+[Setup SSH Server to run unit tests](#setup-ssh-server-to-run-unit-tests).
 
 ## Setup SSH Server to run unit tests
 The goal is to log into the SSH server via `ssh localhost` without using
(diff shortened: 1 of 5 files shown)


FAQ.md
```diff
@@ -30,6 +30,9 @@
   * [Which additional features on top of a GUI does BIT provide over a self-configured rsync backup? I saw that it saves the names for uids and gids, so I assume it can restore correctly even if the ids change. Great! :-) Are there additional benefits?](#which-additional-features-on-top-of-a-gui-does-bit-provide-over-a-self-configured-rsync-backup-i-saw-that-it-saves-the-names-for-uids-and-gids-so-i-assume-it-can-restore-correctly-even-if-the-ids-change-great---are-there-additional-benefits)
   * [How to move snapshots to a new hard-drive?](#how-to-move-snapshots-to-a-new-hard-drive)
   * [How to move large directory in source without duplicating backup?](#how-to-move-large-directory-in-source-without-duplicating-backup)
+- [Testing & Building](#testing--building)
+  * [SSH related tests are skipped](#ssh-related-tests-are-skipped)
+  * [Setup SSH Server to run unit tests](#setup-ssh-server-to-run-unit-tests)
```
aryoda, line 35: great, I was trying hard to find our ssh server setup instructions last weekend :-)

FAQ.md
```diff
@@ -893,3 +896,23 @@ moving the file/folder in the last snapshot, too.
...
+## SSH related tests are skipped
+Some of the tests need an available SSH server.
+They get skipped if this is not the case.
+Plese see
+[Setup SSH Server to run unit tests](#setup-ssh-server-to-run-unit-tests).
+
+## Setup SSH Server to run unit tests
+The goal is to log into the SSH server via `ssh localhost` without using
+a password.
+- Generate an RSA keypair executing `ssh-keygen`.
```
aryoda: I think it is also required for the unit test that the key is not password-protected.
Not sure if this adds security risks requiring more config changes...

common/test/test_backintime.py
```diff
@@ -141,12 +141,8 @@ def test_local_snapshot_is_successful(self):
         #       The same goes with Gtk warnings.
 
         line_beginnings_to_exclude = [
-            "WARNING: Failed to connect to Udev serviceHelper",
-            "WARNING: D-Bus message:",
-            "WARNING: Udev-based profiles cannot be changed or checked",
-            "WARNING: Inhibit Suspend failed",
-            "Warning: Ignoring XDG_SESSION_TYPE=wayland on Gnome. Use QT_QPA_PLATFORM=wayland to run on Wayland anyway",
```
aryoda, line 148: Should we keep the well-known warnings to be ignored in a comment at least?

aryoda (approved):

common/test/test_backintime.py
buhtz: Good idea. Added them as comment.

FAQ.md
buhtz: I added this point to the FAQ.

PR #1552, opened 2023-10-19
Correct spelling of initiate_translator in qt/messagebox.py

asciiprod:
this small typo slipped in when initiate_translator was added

### DIFF
--- a/qt/messagebox.py
+++ b/qt/messagebox.py
@@ -23,7 +23,7 @@ import qttools
 def askPasswordDialog(parent, title, prompt, language_code, timeout):
     if parent is None:
         app = qttools.createQApplication()
-        translator = qttools.initate_translator(language_code)
+        translator = qttools.initiate_translator(language_code)
         app.installTranslator(translator)
 
     import icon


aryoda:
THX a lot for reporting and even fixing this :smile: 

I will merge it soon and add an changelog entry too.

BTW: How did you find this bug?

emtiu (approved):

asciiprod:
I run a daily EncFS backup via cron which normally prompts me to input my SSH and encryption password. After upgrading to Fedora 39 which ships 1.4.0 the prompt was gone. So I started the backup manually in debug mode and got the error message that the method could not be found.
So yeah, this is actually a bigger bug than just a typo.

aryoda:
> So I started the backup manually in debug mode and got the error message that the method could not be found.
> So yeah, this is actually a bigger bug than just a typo.

THX for clarification, this is definitely a HIGH bug in a our released version.
I will check since which release it is contained and update the "known issues" in our README...

GUI testing of all possible user choices is really a pain (almost impossible to automate)

aryoda:
> So I started the backup manually in debug mode and got the error message that the method could not be found.

I am trying to describe the steps to reproduce this bug and have tried it in the BiT GUI but didn't succeed.

In the code I can see that the line with the typo is only called when no parent windows exists so I assume
you have started BiT in the console (without the GUI) to diagnose this. Could you please confirm (or falsify) this so that I can close the documentation of the issue?

THX :smile:

buhtz:
Bad fault caused by me.
Retrospectively can not be sure. But I assume it was an ignored linter warning or a not used linter. I do have two coding machines. On one of them I do not use linters by default because of performance reasons (rasperrypi4).

I have a solution in mind how to prevent problems like this in the future and will come up with it in a separate issue or PR.

About GUI testing: Qt seems to offer a testing framework. It might be possible to create tests for situations like this.

aryoda:
> I have a solution in mind how to prevent problems like this in the future and will come up with it in a separate issue or PR.

Perhaps we could activate linting on the github side  (via Travis CI)

buhtz:
My idea is to add linters (pycodestyle and/or ruff) as a unit test. I 
still do this in my own projects.

But BIT is far away from PEP8 conformance and would produce 1000 of 
errors. Have to play with include/exclude options oft that linters.

I will play with it when I am back.

PR #1554, opened 2023-10-26
fix(doc): Install sphinx_rtd_theme explicit

buhtz:
This fix installs spinx_rtd_them package via pip while building the docs on readthedocs.org.

Previous problem was broken doc builds because of that package missing. This was cased by modified behavior on ReadTheDocs build servers.

I avoid to use a "requirements.txt" file as stated in the docs of RTD. This file is "old-schoold" and its information redundant to a "state of the art" "pyprojects.toml" file. But we do not have one yet. So the current solution is kind of a workaround until we migrate the project to a more modern layout.

Additionally added the timezone (strftime format %Z) to the last-updated-timestamp. Opened a FeatureRequest at Sphinx to improve this. https://github.com/sphinx-doc/sphinx/issues/11738

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -22,4 +22,5 @@ python:
   install:
   - method: pip
     path: .
-    extra_requirements: sphinx_rtd_theme
+    extra_requirements:
+    - sphinx_rtd_theme
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -12,7 +12,7 @@ build:
     python: "3"
   jobs:
     post_create_environment:
-      - pip install sphinx_rtd_theme
+      - python -m pip install sphinx_rtd_theme
 # Build documentation in the docs/ directory with Sphinx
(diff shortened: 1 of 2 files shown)


.readthedocs.yaml
```diff
@@ -19,5 +25,8 @@ sphinx:
...
 # https://docs.readthedocs.io/en/stable/guides/reproducible-builds.html
 # python:
-#   install:
-#   - requirements: docs/requirements.txt
\ No newline at end of file
+#  install:
+#  - method: pip
+#    path: .
+#    extra_requirements:
+#    - foo
```
aryoda, line 32: Is "foo" a copy&paste left-over from some example config?

aryoda (approved):

.readthedocs.yaml
buhtz: Yes. Foo is just a placeholder. Everything in the comments is just a reminder.

Issue #1559, opened 2023-11-08
Unclear what enfcs.Decode does (Question to Germar)


buhtz:
Dear @Germar ,
please have a look at this piece of code realated to `encfstools.Decode` class.

https://github.com/bit-team/backintime/blob/6df8c901ceecd659eaadb993103da3c6e89b4041/common/encfstools.py#L507-L510

I added some debug prints but was not able to trigger that part of the code.
- What does it do?
- Is it relevant?

The reason for my question is if I can modify the source string `_('Take snapshot')` without interfering something.

Is this somehow related to this piece of code?
https://github.com/bit-team/backintime/blob/6df8c901ceecd659eaadb993103da3c6e89b4041/common/snapshots.py#L907-L908

Germar:
This is relevant to decode encrypted path in `takesnapshot.log`. Problem is that BiT writes translated phrase `Take snapshot` into the log for lines like:

https://github.com/bit-team/backintime/blob/6df8c901ceecd659eaadb993103da3c6e89b4041/common/encfstools.py#L526-L531

Everytime the translation changes for this phrase, this will break decoding old logs which has been created before last translation-update.

With this code I tried to fix this at least for the german translation-change. But that's a losing game anyways...

buhtz:
[I] Take snapshot (rsync: rsync: send_files failed to open "...
        ^^^^^^^^^^^^^

Do we talk about this string?

Do I understand that correct? Someone create encfs snapshots with English language. After some snapshots the users decided to switch the UI (or OS) language to something else and do some more snapshots with the same profile.
Then it is not possible to "decode" the old snapshot logs written in English?

It seems to me that we shouldn't touch the source string `_('Take snapshot`)`. Correct? 😄

Germar:
> Do we talk about this string?

Correct!



> Do I understand that correct? Someone create encfs snapshots with English language. After some snapshots the users decided to switch the UI (or OS) language to something else and do some more snapshots with the same profile.
> Then it is not possible to "decode" the old snapshot logs written in English?

Yes, that's correct. But not just because someone changed his locale, this will also happen if the translation of that string changes

> It seems to me that we shouldn't touch the source string `_('Take snapshot`)`. Correct? 😄

I would recommend to not change it. Also changing translation should be handled with care.

It would be better to not translate that string at all and create a regex with every possible translation for backwards compatibility

buhtz:
[...]

PR #1560, opened 2023-11-09
Translation: Minor modifications and update

buhtz:
Let me give you a tip for review: Ignore the content in the "po" folder and just have a short look at "messages.pot" their.

Some minor things I discovered while working on another topic.

- Add warning comment about `_('Take snapshot')` source string (see #1559)
- Improved source strings in _Auto-Remove_ tab of the settings dialog ([mailing list discusson](https://mail.python.org/archives/list/bit-dev@python.org/message/A3YWD4WSBTC4GX3IY5QN4Y2EQWFX6ZR6/)). See screenshot below.
- Improved sources strings in _Question Dialog about restoring files_.
- ~Merged sources strings `Error` and `Errors`. Reasons: This does prevent "Reused translation" errors in our translation system. Singular and Plural in German is similar "Fehler". --> **After investigating languages with multiple plural forms (e.g. Arabic, Polish) I'm sure this solution is not elegant enough.**~
- Fix a minor visual bug in language dialog.
- Sync between upstream po(t) files and Weblate platform.

Here is the new _Auto-Remove_ tab:
image

Here is the new _Restore Question_ in singular form:
image


buhtz:
After doing some research in the translation community I'm not convinced that using `ngettext()` for "Error" and "Errors" is a good solution. The reason is that there are languages with multiple plural forms (e.g. Arabic and Polish). It seems to me that `ngettetx()` make sense only when used with a concrete number/count. But in our situation we do use "Errors" here in the meaning of "All errors", which is the plural-form "Others" in Arabic and "Few" in Polish.

I'm doing further research for a more elegant solution to prevent "reused translation" errors.

*EDIT*: It might be the easiest to add `ignore-reused` as a flag to the source string via the Weblate platform. I have to investigate if this flag persists after using our github-weblate-sync process.

qt/logviewdialog.py
```diff
@@ -89,13 +89,18 @@ def __init__(self, parent, sid = None, systray = False):
...
         self.comboFilter.addItem(_('All'), 0)
+
+        # Note about ngettext plural forms: n=102 means "Other" in Arabic and
+        # "Few" in Polish.
+        # Research in translation community indicate this as the best fit to
+        # the meaning of "all".
         self.comboFilter.addItem(' + '.join((_('Errors'), _('Changes'))), 4)
         self.comboFilter.setCurrentIndex(self.comboFilter.count() - 1)
-        self.comboFilter.addItem(_('Errors'), 1)
+        self.comboFilter.addItem(_('Errors'))
```
aryoda: I didn't dig deeper but why remove the "1" as argument if the other `addItems` do need this argument?

aryoda (approved):

qt/logviewdialog.py
buhtz: I have not noticed. You are right that is a mistake. The `1` is the value of the combo box entry and has nothing to do with GNU gettext.

buhtz:
I double checked the diff again.

Issue #1558, opened 2023-11-07
Check and fix linter findings of invalid member names (E1101)


aryoda:
`pylinter` does  about some missing member names in the `qt` src folder and a lot more in the `common` src folder.

This should be checked and fixed (or do [suppress the false positive linter finding via code directive](https://techoverflow.net/2021/04/05/how-to-ignore-pylint-problem-for-a-specific-single-line-of-code/)).

## qt

```
~/dev/backintime (doc/ssh_setup)  > python -m pylint --disable=all --enable=E1101 qt
************* Module playground
qt/playground.py:5:21: E1101: Module 'qttools' has no 'initate_translator' member; maybe 'initiate_translator'? (no-member)
************* Module snapshotsdialog
qt/snapshotsdialog.py:287:12: E1101: Module 'restoredialog' has no 'restore' member (no-member)
qt/snapshotsdialog.py:292:12: E1101: Module 'restoredialog' has no 'restore' member (no-member)

-----------------------------------
Your code has been rated at 9.96/10
```

## common

```
~/dev/backintime (doc/ssh_setup)  > python -m pylint --disable=all --enable=E1101 common
************* Module encfstools
common/encfstools.py:54:39: E1101: Instance of 'EncFS_mount' has no 'mode' member (no-member)
common/encfstools.py:54:50: E1101: Instance of 'EncFS_mount' has no 'path' member (no-member)
common/encfstools.py:62:79: E1101: Instance of 'EncFS_mount' has no 'mode' member (no-member)
common/encfstools.py:69:15: E1101: Instance of 'EncFS_mount' has no 'reverse' member (no-member)
common/encfstools.py:73:22: E1101: Instance of 'EncFS_mount' has no 'path' member (no-member)
common/encfstools.py:115:11: E1101: Instance of 'EncFS_mount' has no 'config_path' member (no-member)
common/encfstools.py:116:31: E1101: Instance of 'EncFS_mount' has no 'path' member (no-member)
common/encfstools.py:118:31: E1101: Instance of 'EncFS_mount' has no 'config_path' member (no-member)
common/encfstools.py:163:11: E1101: Instance of 'EncFS_mount' has no 'reverse' member (no-member)
************* Module mount
common/mount.py:477:68: E1101: Instance of 'MountControl' has no 'mode' member (no-member)
common/mount.py:530:30: E1101: Instance of 'MountControl' has no 'log_command' member (no-member)
[...]

Issue #1489, opened 2023-07-28
Test & Build: Distinguish between "unit", "system" and "integration" tests


buhtz:
Blocked by
- #1575

# Problem
BITs test suite do have three kinds of tests in it: unit, integration and system tests. Integration and system tests are valuable but they do not run under every circumstances (e.g. chroot build environments or other sandboxes). Debian do has the problem ([Debian#940319](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=940319)) for example.

# Proposal
Separate between unittests and all other tests to account for the different characteristics of these test categories.

# Solution
Currently BIT do use `make` as build system also to invoke tests via `pytest` or `python -m unittest`. Create folder `common/tests/unittests` containing unittests only. With this solution a `make test` will invoke all tests (including the ones in the sub-folder `tests/unittests`) and `make unittest` will invoked only the tests in the sub-folder. This should be easy to realize via our `configure` script.
More accurate we also could create `tests/system` and `tests/integration` which would force us to think about what test kind we do have. Another side effect would be a more realistic coverage number. IMHO "code coverage" should count only for unittests.

The usual `make unittest` then could be used by Debians build system (see below).

# More background
Related to the [Debian bug ticket #940319](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=940319) I did some more research and also got some valuable advice of people around Debians Python Packaging Team. The Debian ticket is about the problem that our tests want to write to HOME folder which is not possible in Debians build system because they do `chroot` as a sandbox solution.

But fixing the HOME problem with using a temp folder or a fake in-memory-filesystem won't solve the problem described in this Issue here. Our tests do create ssh keys, installing them, contacting an ssh server, etc. There is no way to mock that and there is no need for it. Integration and system tests are also valuable and they still prevented us from some mistakes.

# Which tests to separate?
Looking through the code in `common/tests` I only can find `diagnostics.py` to be a real unittest without dependencies to other BIT components. Also `argparse.py` is a candidate for a unittest but need some more research and modificaton. All other tests seem to be real integration- or system-tests to me. No other unittests do exist.

Sorry, this issue is not fun. 😿

buhtz:
Related to #1463

buhtz:
I tried to fix this with [`buhtz:1489testcategories`](https://github.com/buhtz/backintime/tree/fix/1489testcategories) but messed up with TravisCI, Coveralls and configure-make script I'm not familiar with.

[...]

PR #1562, opened 2023-11-10
build: Add pylint as unit test checking for "no-member" (E1101)

buhtz:
Add `pylint` as a unit test for `common` and `qt` checking for `E1101` (no-member) as happend in #1558.

~Draft:  Some more work todo in this PR. Some refactoring need to be done because of E1101. But feel free to review the `test_lint.py` file first.~

- Created `test_lint.py` for `common`.
- Created a `test` folder for `qt`.
- Activate `qt` tests in TravisCI.
- Add `pylint` as build dependency in `CONTRIBUTION.md`.

Keep in mind that `pylint` is an extremely slow linter compared to others. But it can catch lints and bugs (e.g. E1101) others can not. So there is value in the cost.


buhtz:
OK, with merging #1564 and #1665 all pylint-E1101 related problems should be fixed.

.travis.yml
```diff
@@ -43,9 +45,11 @@ install:
 
 script:
   # compile all files - ensure that syntax is correct
-  - python -m compileall common common/test common/plugins qt qt/plugins
+  - python -m compileall common common/test common/plugins qt qt/test qt/plugins
   # run unit tests - ensure that functionality is correct
   - cd common && ./configure && make unittest-v
+  - cd ..
+  - cd qt && pytest
```
aryoda, line 50: Which tests are executed by pytest here (we don't have any tests so far in qt)?
Or is this meant as preparation of upcoming GUI unit tests?

common/test/test_lint.py
```diff
@@ -0,0 +1,67 @@
+import unittest
+import pathlib
+import subprocess
+from importlib import metadata
+from typing import Iterable
+
+# PACKAGE_NAME = 'buhtzology'
+
+
+class MirrorMirrorOnTheWall(unittest.TestCase):
```
aryoda, line 10: Cool name ;-)


I am not sure if  `pylint` checks should be implemented as unit test since this impacts package maintainer too
regarding dependencies and possible false positives (eg. ARCH AUR package installs do ALWAYS execute the unit tests).

Could we somehow trigger `pylint` only via Travis CI but without unit tests (similar to how the code coverage estimation is calculated)?

**Edit:** Perhaps a simple solution would be to protect the unit test execution with `if_travis()`... but this would hinder local execution on the other hand which is required before sending any PR IMHO...

.travis.yml
aryoda: Ah I do understand now, this is a "forward reference" :-)
You have added the first unit test below...

buhtz:
Seems to work.

aryoda (approved):

aryoda:
@buhtz Great work!

I see a possible impact on downstream package maintainers if we do call the linter in a unit test
(eg. ARCH AUR package installs do ALWAYS execute the unit tests but linter results should never have an impact on packaging or installation but only during development, testing and release preparation).

I would feel better if we could do the linter check somehow separately from the unit test but I don't see an easy way how to do this ATM.

Since we are doing "just" a code check of explicitly enabled linter rules I think it is OK to go this way for now because new rules in newer linter releases on different platforms should not become active to interfere with new **unexpected** findings at the packaging/release point-in-time.

buhtz:
This seems to be related to #1489 which is about separating tests in sub-folders and divide between unit, system and integration tests. There could be a 4th category "misc" for exotic stuff like linters.
In that case distro maintainers can decide which part of the tests to run.

> I see a possible impact on downstream package maintainers if we do call the linter in a unit test (eg. ARCH AUR package installs do ALWAYS execute the unit tests but linter results should never have an impact on packaging or installation but only during development, testing and release preparation).

To my understanding I see no problem here. The linter is a build dependency (listed in CONTRIBUTING.md) and should be installed by the AUR maintainer. The linter will run and should be green. Or do you expect linter result on AUR somehow different from our TravisCI or our local machines?

> I would feel better if we could do the linter check somehow separately from the unit test but I don't see an easy way how to do this ATM.

TravisCI is one way. But then linting is not done automatically my local machine.
In other projects with more clean and PEP8 conform code I do use multiple linters in unit tests. This is easier to me then configuring multiple linters in an IDE. And unit tests linting force every developer to take care of the code.

> new rules in newer linter releases on different platforms

On the long run we could fix the version of the linter. This can be technically implemented when we migrate to pyproject.toml using project structure.

aryoda (approved):

PR #1568, opened 2023-11-10
add Codespell workflow and fix some typo

Fantu:
codespell github workflow help to spot typo and show them in pull request
I also fixed some typo, major spotted with codespell


buhtz:
Dear Fantu,
thanks for your contribution.

Modifying infrastructure should be discussed with the maintenance team first. I am sorry but we do not use workflow/actions or something else that is exclusive to Microsoft GitHub. The goal is to prevent a locked-in effect. We plan on the long run to migrate to another code hoster.

But for my learing: What is a "workflow" technically? Is it a VM or container?
And what application (a linter for example) do run in that container to check the spellings? It seems to me that this tool do check more than just the py-files?
If you can recommend a good tool we can integrate it into our current infrastructure (unit tests).

**EDIT**: Is it just https://github.com/codespell-project/codespell/ ?

**EDIT2**: Seems like a nice tool, easy to integrate in our unit tests. And it can be configured via `pyproject.toml` file (which we will have after migrating the project structure). To my current knowledge I see no need to use a closed source and locking-in Microsoft product for this.

CHANGES
```diff
@@ -326,15 +326,15 @@ Version 1.1.4 (2015-03-22)
...
 * Fix bug: failed to restore file owner and group
-* cache uuid in config so it doesn't fail if the device isn't pluged in (https://launchpad.net/bugs/1426881)
+* cache uuid in config so it doesn't fail if the device isn't plugged in (https://launchpad.net/bugs/1426881)
 * add warning about wrong Python version in configure
 * prevent snapshots from being removed with restore and delete; show warning if restore and delete filesystem root (https://answers.launchpad.net/questions/262837)
 * Fix bug: OSError in free_space; add alternate method to get free space
 * add bash-completion
 * Fix bug: ugly theme while running as root on Gnome based DEs (https://launchpad.net/bugs/1418447)
-* Fix bug: filename with broken charset throwed UnicodeError exception (https://launchpad.net/bugs/1419694)
+* Fix bug: filename with broken charset trowed UnicodeError exception (https://launchpad.net/bugs/1419694)
```
buhtz: That is not correct. Is it `thrown`?
I assume you used the auto-correct feature of codespell? Using tools that manipulate the code automatic (e.g. `black`) need to be mentioned when opening PRs and is not recommended.

buhtz (changes requested):
The workflow should be deleted. Integrating codespell will be handled in a separate issue.
Except one mistake (see other comment) in the rest looks OK.

But please let's wait until the other PRs (#1562, #1567, #1560) from today are merged. I expect conflicts.

CHANGES
Fantu: @buhtz I did changes manually, major was spotted by codespell, after I opened the files with errors and fixed them excluding the cases where are code, other languages or other things to not changes. In the editor I have also an extension which underlines words that are not in the English vocabulary, most are not to be considered while some caught my eye because they were actually errors.
others was related to function name, for example:
`SotedcomoBox, SnapshotComba, ProfilCombo` for each I searched the function with correct name in the file and I corrected them, for example to: `SortedcomboBox, SnapshotCombo, ProfileCombo`
regarding the changelog, probably because I was already going through and checking files for an hour, it hadn't occurred to me that it could refer to "Python Throw Exception", I think I made a mistake, sorry, I hope I haven't made any other mistakes
I hope I haven't made any other mistakes also due to my imperfect English^^'

Fantu:
as you prefer if you want to abandon github, however it doesn't seem to me that Microsoft has done enough damage in these 5 years, or am I wrong?

CHANGES
buhtz: I really appreciate your work. You have invested a lot of time here and also on the Debian site. And I apologize if I might have become a bit harsh.
It is easy to drive me emotional when it comes to Microsoft, Ubuntu or something similar. ;) And please be aware that the "BIT Team" currently have not official consents about things like this. Most of the time it is my personal opinion.

Please fix the "thrown" typo and remove the workflow-codespell file from your PR. Then we can merge it.

I will add codespell on my todo list and see if it make sense to integrate it into our unittests (like we do with PyLint in #1562). It might give to much false positives. Also getting spellchecking working in my Emacs setup is a long existing todo on my list.

buhtz:
> it doesn't seem to me that Microsoft has done enough damage in these 5 years, or am I wrong?

To my knowledge there was no damage explicit to BIT. But they disrespect our LICENCE with feeding CoPilot with our repository.
Microsoft (MS) banned open source developers only because they are from "bad countries" (enemies) for example.
They use our work to feed their CoPilot algorithm.
Beside that behavior it is a matter of principles to me. GitHub is closed source. The company behind GitHub has proven more then once that they do not respect privacy of their users. They have no respect to Free Software and the four rights but using the Open Source label to "green wash" ("open wash" maybe?) their own image.
Their responsibilities are tide to USA government.

Technically currently it is no problem to migrate BIT to another code hoster because we do not use GitHub specific features. Even our TravisCI config might be easy to migrate if we would do this. We should keep the relationship in that current "friends with benefits" state and do not go further.

Fantu:
those defined as "artificial intelligence" are now being used everywhere, I think that they largely bring negative aspects, but it now seems like an unstoppable trend implemented even where it is counterproductive.
I suppose that even if it hadn't been acquired by Microsoft, Github would sooner or later have created a similar AI, whether for money and/or customer request.
the open source code being open and reachable by anyone is at risk of being "abused", it was before without "AI" and it is even more so now (any repository, even self-hosted, if it is indexed even on just one search engine, it is). only if it is inaccessible would it be protected (but then it would not be open source), and protected it still remains relative (until it is hacked)
just as for privacy, unfortunately I have seen that trying to achieve it involves an enormous effort and amount of time with poor results (and it gets worse and worse :( ) so personally I no longer waste too much time and energy on it. in practice, if you want true privacy (or almost) you should go and live isolated and without information technology^^'

Fantu:
removed the workflow and the wrong change

> I will add codespell on my todo list and see if it make sense to integrate it into our unittests (like we do with PyLint in #1562). It might give to much false positives.

there was big list of "false positive", but I already excluded them in codespell excluding po files (for major) and few words for others, on codespell workflow was:
```
skip: "*.po"
ignore_words_list: manuel,dum,sistem,clude
```
there is something similar using codespell from command line tool installed locally and I suppose also using it from pylint (but I never used it and I not checked)

aryoda:
@Fantu Thanks a lot for investing your time to help us saving time!

buhtz:
> ```
> ignore_words_list: manuel,dum,sistem,clude
> ```

Can you explain why `manuel` needs to be lower case no matter that the source string is upper case? Even after reading the docu I don't get it. I opened [an Issue at codespell about it](https://github.com/codespell-project/codespell/issues/3210).

Fantu:
word to ignore must be always all lowercase, as I saw works only putting lowercase (like dictionary) and after will work also for uppercase or mixed result (so for example manuel in ignore word list will skip Manuel, manuel or MANUEL)
from man:
```
-L WORDS, --ignore-words-list WORDS

    comma separated list of words to be ignored by codespell. Words are case sensitive based on how they are written in the dictionary file
```
and example of codespell dictionary:
https://github.com/codespell-project/codespell/blob/master/codespell_lib/data/dictionary.txt
I didn't understand either initially and I understood after seeing the dictionaries from the codespell source

PR #1569, opened 2023-11-11
build: Codespell setup

buhtz:
This is a setup for codespell. The config file `.codespellrc` is used for this. I hope that most of the variables in there are documented in an understandable way.

Nice to know: Codespell is not a usual spellchecker. The later do mark everything unknown (not existent in a dictionary) as an error. Codespell works the other way around. It has a dictionary of errors. It can not find every error or misspell but it can find the usual ones. It prevents to much false positives. I really like that idea.

Install codespell via pip and just run `codespell` in the root directory of the repo.
The question now is if and how we should integrate that in our infrastructure.

I would vote to just describe it as an optional but recommended step in the "How to prepare a release" docu.

Technically it is possible to add it to TravisCI. But I don't see much advantage of it. See next point.

I also don't think it is a good idea to add this to the unit tests like we do with PyLint. Codespell can produce false positive errors. They need to be handled by ignore rules etc pp. Contributors need to know how codespell works before they can do this. Or we as maintainers need to catch up situations like this when contributors open PRs with failing TravisCI runs.
This is IMHO to much work and "noise" for some simple typos.

I would like to keep it simple.

**EDIT**: Additionally I activated PyLint warning W1401 about invalid escape sequences (see #1567).


buhtz:
The decision to make now is how to integrate that in our infrastructure:

 1. Just describe it as an optional but recommended step in the "How to prepare a release" docu.
 2. Add it to TravisCI.
 3. Something else?

I would prefer 1. in the first place. The 2. would confuse new contributors because they may do not understand why TravisCI fails when they opened a PR. And they may not know how to fix a false positive because they don't know how codespell works.

Fantu:
codespell github workflow have a warn only option that don't make fails the test but only list the error spotted, I don't know travis if possible with it do a similar thing (on the codespell part only)

buhtz:
> codespell github workflow have a warn only option that don't make fails the test but only list the error spotted, I don't know travis if possible with it do a similar thing (on the codespell part only)

I am aware of that "feature". But this "activate" the locked-in effect. It is not bad but only nice to have.

aryoda:
I think it is enough for now to add the codespell check to the "how to prepare a release" documentation.

Should we add the `codespell` package as an optional "build" dependency or mention the installation only in above documentation?

aryoda (approved):

PR #1576, opened 2023-12-04
FAQ improvment

buhtz:
Add "11 leading characters in snapshot log" topic to the FAQ.
Tried to improve the information about "incremental" backups.
Reorganized the FAQ with introducing the section "General".

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -1,5 +1,11 @@
 # FAQ - Frequently Asked Questions
 <!-- TOC start - via https://derlin.github.io/bitdowntoc/-->
+- [General](#general)
+  * [Does Back In Time create incremental backups?](#does-back-in-time-create-incremental-backups)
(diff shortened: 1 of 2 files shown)


aryoda:
> Add "11 leading characters in snapshot log" topic to the FAQ

Very good idea!

On my personal list I have a TODO to add a kind of hoover-over tooltip in the changes list which shows the translation of the characters under the cursor as tooltip so that nobody must use the browser to understand this.
Alternatively a "Help" button in the dialog could be used to explain these special characters...

buhtz:
> add a kind of hoover-over tooltip in the changes list which shows the translation of the characters under the cursor as tooltip

That is a good idea. I also do have some thoughts about improving the GUI with better using the abilities of Qt and GUI design. The current solution is just a workaround and not user friendly.

The FAQ is like an entry to me. In the future we can migrate (or copy) the information into the manual, too.

Issue #1580, opened 2023-12-08
BIT hangs when starting a backup as root


jim-moe:
BackInTime v1.4.1 (from opensuse OSS main repo)
Opensuse tumbleweed v20231130
linux 6.6.3-1-default x86_64
python 3.11.6
rsync  version 3.2.7  protocol version 31

BIT hangs when starting a backup as user root. This is not a problem for non-privileged users.
It is similar to the issue with `xdpyinfo`; I could bypass that by killing the `xdpyinfo` instance.

The hang occurs here, shown by ps. When this process is killed, BIT proceeds to backup.
`root 31136 35.7 0.0 190892 26960 ? SNl  09:00  19:36 \_ /usr/bin/python3 /usr/share/backintime/common/qt5_probing.py`
There was nothing unusual in the system journal. There was nothing noted in` takesnashpot_.log`.

```
$ backintime --diagnostics
Traceback (most recent call last):
  File "/usr/share/backintime/common/backintime.py", line 1190, in <module>  startApp()
  File "/usr/share/backintime/common/backintime.py", line 507, in startApp
    args = argParse(None)
           ^^^^^^^^^^^^^^
  File "/usr/share/backintime/common/backintime.py", line 568, in argParse
    args, unknownArgs = mainParser.parse_known_args(args)
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.11/argparse.py", line 1902, in parse_known_args
    namespace, args = self._parse_known_args(args, namespace)
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.11/argparse.py", line 2114, in _parse_known_args
    start_index = consume_optional(start_index)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.11/argparse.py", line 2054, in consume_optional
    take_action(action, args, option_string)
  File "/usr/lib64/python3.11/argparse.py", line 1978, in take_action
    action(self, namespace, argument_values, option_string)
  File "/usr/share/backintime/common/backintime.py", line 742, in __call__
    diagnostics = collect_diagnostics()
                  ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/share/backintime/common/diagnostics.py", line 74, in collect_diagnostics
    'OS': _get_os_release()
          ^^^^^^^^^^^^^^^^^
[...]

PR #1581, opened 2023-12-09
fix: Full os-release path in --diagnostics

buhtz:
This fix a minor bug (discovered in #1580) in context of diagnostic output handling os-release file.

IMHO no need for a changelog entry because it is very minor.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,7 @@ python:
   - "3.9"
   - "3.10"
   - "3.11"
+  - "3.12"
 
 before_install:
   # disable mongodb as we don't need it and it sometimes temporary fails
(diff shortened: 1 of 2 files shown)


buhtz:
Ignore the failed TravisCI. It is only about the one PC64le machine using Python 3.12. Support will fix this soon.

I activate Python 3.12 (the latest stable) at TravisCI because Debian started to migrate to it in its testing repo. Can we keep it? Any arguments against it?

aryoda:
> I activate Python 3.12 (the latest stable) at TravisCI because Debian started to migrate to it in its testing repo. Can we keep it? Any arguments against it?

Adding 3.12 is a good idea

aryoda (approved):

Issue #1573, opened 2023-11-21
How to automatically retake snapshot after rsync error 23 using user-callback


jean-christophe-manciot:
Here is the workflow inside my user-callback:
```
#!/bin/bash

profile_id="$1"
profile_name="$2"
backintime_status="$3"
...
if [[ ${profile_id} -eq 1 ]]; then
        # Working around a bug
        snapshot_log_file=/root/.local/share/backintime/takesnapshot_.log
else
        snapshot_log_file="/root/.local/share/backintime/takesnapshot_${profile_id}.log"
fi
case $backintime_status in
        1) #Backup process begins
                printLog "1 - backup process begins"
                ;;
        2) #Backup process ends
                printLog "2 - backup process ends"

                # Waiting for the log to be complete
                sleep 1m
                rsync_return_code=$(grep -P "'rsync' ended with exit code " "${snapshot_log_file}" | sed -E "s|^.*'rsync' ended with exit code ([0-9]+).*$|\1|g")

                case $profile_name in
                        "Main profile")
                                case "$rsync_return_code" in
                                        0|24)
                                                /usr/bin/nice -n 19 /usr/bin/ionice -c2 -n7 /usr/bin/backintime --profile "$next_profile_name" backup
                                                ;;
                                        23)
                                                # Retaking the backup
                                                /usr/bin/nice -n 19 /usr/bin/ionice -c2 -n7 /usr/bin/backintime --profile "$profile_name" backup
                                                ;;
                                        *)
                                                # Saving snapshot log
                                                cp --force --preserve=all "${snapshot_log_file}" "${snapshot_log_file}.$(date --iso-8601=minutes)"
                                                exit 1
                                                ;;
                                esac
                                ;;
                ...
                esac
        ...
esac
```

Unfortunately, this does not work when **rsync error 23** happens; I get in **takesnapshot.log**:
```
...
notify-send Back In Time (user) : profile_name Error: rsync: [sender] read errors mapping "filename": No data available (61)
notify-send Back In Time (user) : profile_name 'rsync' ended with exit code 23: See 'man rsync' for more details
notify-send Back In Time (user) : profile_name Failed to take snapshot 2023-11-19 19:01:14.'
ERROR: [plugins/usercallbackplugin.py:94 UserCallbackPlugin.callback] user-callback returned 'INFO: Lock
INFO: Take a new snapshot. Profile: profile_id profile_name
INFO: Call rsync to take the snapshot
[...]

PR #1590, opened 2023-12-18
Add developer documentation for usage of lock, flock and other files in BiT ( related to #1573)

aryoda:
Mainly contributes a new dev doc file:

https://github.com/aryoda/backintime/blob/issue/1573_retry_backup/common/doc-dev/4_Control_files_usage_(locks_flocks_logs_and_others).md

This documentation is part of #1573

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -131,7 +131,7 @@ schedules.
 
 ## Known Errors and Warnings
 ### WARNING: A backup is already running
-_Back In Time_ uses signal files like `worker.lock` to avoid starting the same backup twice.
+_Back In Time_ uses signal files like `worker<PID>.lock` to avoid starting the same backup twice.
 Normally it is deleted as soon as the backup finishes. In some case something went wrong
 so that _Back In Time_ was forcefully stopped without having the chance to delete
 this signal file.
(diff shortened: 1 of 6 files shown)


buhtz:
Wow. That is awesome. This topic(s) made me headaches. Now I can sleep well. It is a good starting point to create isolated Classical/Detroit-style unit tests for that parts.

I fixed some URLs in the TOC and a paragraph to improve the readability. Be aware that online TOC generators using different flavours of markdown. I do struggle with this topic all the time even on Gitea/Forgejo and not only at GitHub. I haven't found a trustful generator yet.

About the content: I don't understand the difference between `worker<Profile ID>.lock` and `worker<Profile ID>.lock.flock`. Do they need to exist separate?

buhtz (approved):

aryoda:
> About the content: I don't understand the difference between `worker<Profile ID>.lock` and `worker<Profile ID>.lock.flock`. Do they need to exist separate?

If two processes check in parallel if the the `worker<Profile ID>.lock` already exists but **before** one of them has already created one there is small amount of time where both processes "think" they are the only ones.

This could cause
1.  one process to overwrite the just created lock file of the other process
2. run BiT in parallel

To solve this a **separate** *.flock file is used using a blocking file lock to indicate other processes: "I am about to check the existence of the *.lock file and will create a new one if possible".

The advisory (f)lock is blocking which means if the other process wants to set the (f)lock on an existing already locked file the function call to set the (f)lock **waits** until the other process has removed the (f)lock.
This is called **serialization**.

PR #1596, opened 2024-01-01
Install dbus configuration file to /usr/share not /etc

Fantu:
This was done in Debian packages in version 1.4.1-1

From https://dbus.freedesktop.org/doc/dbus-daemon.1.html:

> The standard system bus normally reads additional XML files from /usr/share/dbus-1/system.d. Third-party packages should install the default policies necessary for correct operation into that directory, which has been supported since dbus 1.10 (released in 2015).
> The standard system bus normally also reads XML files from /etc/dbus-1/system.d, which should be used by system administrators if they wish to override default policies.
> **Third-party packages would historically install XML files into /etc/dbus-1/system.d, but this practice is now considered to be deprecated: that directory should be treated as reserved for the system administrator.**

### DIFF
--- a/qt/configure
+++ b/qt/configure
@@ -162,10 +162,10 @@ addUninstallDir                                     "/share/dbus-1"
 addNewline
 
 addComment "dbus conf"
-addInstallDir                                    "/../etc/dbus-1/system.d"
-addInstallFiles "net.launchpad.backintime*.conf" "/../etc/dbus-1/system.d"
-addUninstallDir                                  "/../etc/dbus-1"
-addUninstallDir                                  "/../etc"
+addInstallDir                                    "/share/dbus-1/system.d"
+addInstallFiles "net.launchpad.backintime*.conf" "/share/dbus-1/system.d"
+addUninstallDir                                  "/share/dbus-1"
+addUninstallDir                                  "/share"
 addNewline
 
 addComment "polkit action"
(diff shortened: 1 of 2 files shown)


aryoda:
Thanks a lot for your PR (and giving us the context for the `dbus` config location change) :smile: 

Please give me a few days to find and perform a test case to avoid a possible regression.

qt/configure
```diff
@@ -162,10 +162,10 @@ addUninstallDir                                     "/share/dbus-1"
...
 
 addComment "dbus conf"
-addInstallDir                                    "/../etc/dbus-1/system.d"
-addInstallFiles "net.launchpad.backintime*.conf" "/../etc/dbus-1/system.d"
-addUninstallDir                                  "/../etc/dbus-1"
-addUninstallDir                                  "/../etc"
+addInstallDir                                    "/share/dbus-1/system.d"
+addInstallFiles "net.launchpad.backintime*.conf" "/share/dbus-1/system.d"
+addUninstallDir                                  "/share/dbus-1"
+addUninstallDir                                  "/share"
```
aryoda, line 175: I am just wondering if we still need to care about uninstalling from `/etc` too to avoid orphaned files in `/etc` if the "old" `make install` was used. This could lead to strange side effects if the same file exists in different versions in `/etc` and `/share`.

aryoda (approved):

PR #1600, opened 2024-01-06
Add option to restrict rsync to one file system

### LINKED ISSUE #1598: Add option to exclude network file systems
I'd like to be able to configure bit to ignore network file systems.

Use case: 
In user space, I mounted a smb share on a remote system using a mount point in my home directory.  BIT tried to back up that directory because I hadn't excluded it.

Obviously the work around is to exclude the mount point from the backup, but if a share gets mounted on a different directory, my backup's not know to exclude them.

It would be much better to have an option to exclude network file systems in general.

fallingrock:
Added the ability to include the ```--one-file-system``` option on the rsync command.

This will prevent rsync from going into file systems on other devices and systems.

Fix #1598


buhtz:
Dear David,
thanks a lot for this contribution. I will have a look onto it.

We need to discuss this in the team but I would say we wait until the upcoming release (round about 15th January) before we merge this. But it is just a (slow-down) feeling of myself. There is no strict technical reason.

buhtz:
I excluded the Travis job for Python3.12 on a ppc64le machine because of an unfixed Issue on the site of TravisCI. I am in contact with their support.

Open todos for the team:
 - Question: Wait until next release?
 - Please review the English strings. I am not an expert here.

CHANGES
```diff
@@ -1,4 +1,6 @@
 Back In Time
+Version 1.5.0-dev (development of release after upcoming release)
+* Improve: Support rsync '--one-file-system' in Expoert Options (#1600, #1598).
```
```suggestion
* Improved: Support rsync '--one-file-system' in Expert Options (#1600, #1598).
```

common/config.py
```diff
@@ -1259,13 +1259,21 @@ def setCopyUnsafeLinks(self, value, profile_id = None):
...
     def copyLinks(self, profile_id = None):
-        #?When  symlinks  are  encountered, the item that they point to
+        #?When symlinks are encountered, the item that they point to
         #?(the reference) is copied, rather than the symlink.
         return self.profileBoolValue('snapshots.copy_links', False, profile_id)
 
     def setCopyLinks(self, value, profile_id = None):
         return self.setProfileBoolValue('snapshots.copy_links', value, profile_id)
 
+    def one_file_system(self, profile_id = None):
```
emtiu:
```suggestion
    def oneFileSystem(self, profile_id = None):
```

common/config.py
```diff
@@ -1259,13 +1259,21 @@ def setCopyUnsafeLinks(self, value, profile_id = None):
...
+        #?Use rsync's "--one-file-system" to avoid crossing filesystem
+        #?boundaries when recursing.
+        return self.profileBoolValue('snapshots.one_file_system', False, profile_id)
+
+    def set_one_file_system(self, value, profile_id = None):
```
emtiu:
```suggestion
    def setOneFileSystem(self, value, profile_id = None):
```

common/tools.py
```diff
@@ -892,6 +892,9 @@ def rsyncPrefix(config,
     else:
         cmd.append('--links')
 
+    if config.one_file_system():
```
emtiu:
```suggestion
    if config.oneFileSystem():
```

qt/settingsdialog.py
```diff
@@ -1393,6 +1416,7 @@ def updateProfile(self):
         self.cbPreserveXattr.setChecked(self.config.preserveXattr())
         self.cbCopyUnsafeLinks.setChecked(self.config.copyUnsafeLinks())
         self.cbCopyLinks.setChecked(self.config.copyLinks())
+        self.cbOneFileSystem.setChecked(self.config.one_file_system())
```
emtiu:
```suggestion
        self.cbOneFileSystem.setChecked(self.config.oneFileSystem())
```

qt/settingsdialog.py
```diff
@@ -1587,6 +1611,7 @@ def saveProfile(self):
         self.config.setPreserveXattr(self.cbPreserveXattr.isChecked())
         self.config.setCopyUnsafeLinks(self.cbCopyUnsafeLinks.isChecked())
         self.config.setCopyLinks(self.cbCopyLinks.isChecked())
+        self.config.set_one_file_system(self.cbOneFileSystem.isChecked())
```
emtiu:
```suggestion
        self.config.setOneFileSystem(self.cbOneFileSystem.isChecked())
```

emtiu (commented):
Make the spelling consistent with the other options. Careful: Just editing here, I haven't had a chance to test this code.

buhtz:
Thanks for the review. Do you think we should wait until the (in two weeks) upcoming release?

> Make the spelling consistent with the other options.

I agree that consistence will improve code quality in most cases. But ( :smile: ) in this case my focus was more on resources. To the best of my recollection, the decision was not final, but we somehow agreed to stick to PEP8 as a minimal and in the Python world well known and accepted coding style. I would take this into account no matter that it will result in style-mixed code files.
IMHO saying "We are sticking to it because it has always been like this." won't help improving BIT in the long run.

In the end I will stick to the teams opinion here. I am very flexible about that. :heart:

CHANGES
buhtz: I would stick to `Improve` because we do write `Fix bug` instead of `Fixed bug`.
In the long run there are also [standards about how to write a changelog](https://common-changelog.org/). But I didn't want to bother you with this discussion yet. :laughing:

fallingrock:
Thanks.

As I said, I'm a newbie when it comes to python, so I'm not familiar with standards. I was copying existing code.

buhtz:
> As I said, I'm a newbie when it comes to python, so I'm not familiar with standards. I was copying existing code.

You did a good job. No problem. It is usual that PRs are slightly modified by maintainers.

buhtz:
OK, I updated to the latest upstream "dev".
I did some manual tests and checked the debug output. Seems to work.

How do we decide about the naming style of the methods involved in this PR? Keep the old non-PEP8 naming convention (e.g.  `oneFileSystem()`) or beginn with PEP8 conformance (e.g. `one_file_system()`) no matter that this will currently create a mix of two standards in the code? As a PEP8-fanatic I vote for the latter. 🤣 

Any objects against merging this?

aryoda:
> beginn with PEP8 conformance (e.g. `one_file_system()`) no matter that this will currently create a mix of two standards in the code

I also prefer using underscores as word separator due to better readability even though we will have a mix for next years (or decades ;-) I guess.

qt/settingsdialog.py
```diff
@@ -1018,6 +1022,25 @@ def __init__(self, parent):
...
+            'From \'man rsync\':\n'
+            + '\n'.join(textwrap.wrap(
+                'This tells rsync to avoid crossing a filesystem boundary '
+                'when recursing. This does not limit the user\'s ability '
+                'to specify items to copy from multiple filesystems, just '
+                'rsync\'s recursion through the hierarchy of each directory '
+                'that the user specified, and also the analogous recursion '
+                'on the receiving side during deletion. Also keep in mind '
+                'that rsync treats a "bind" mount to the same device as '
+                'being on the same filesystem.',
```
aryoda, line 1039: longest tooltip I ever saw ;-)

emtiu:
> How do we decide about the naming style of the methods involved in this PR? Keep the old non-PEP8 naming convention (e.g. `oneFileSystem()`) or beginn with PEP8 conformance (e.g. `one_file_system()`) no matter that this will currently create a mix of two standards in the code? As a PEP8-fanatic I vote for the latter. 🤣

I don't like the idea of introducing a new naming scheme with for one function, so that it is different from all its siblings.

But my objection is not very strong, so you may go ahead.

qt/settingsdialog.py
buhtz: There are much more of this from-rsync-manpage-copy-and-pasted-tooltips. That is why I introduced the wrapping, to prevent translators from dealing with line breaks them selfs.

buhtz:
> I don't like the idea of introducing a new naming scheme with for one function, so that it is different from all its siblings.

Sounds reasonable to me. Also might confuse extern contributors.

buhtz:
Now I set it back to camel case. :rofl: 
I think it is even better to modify such PEP8 related things (nearly) en bloc .

buhtz (approved):

Issue #1574, opened 2023-12-03
Backintime install script only uses python / python3 even if other versions are available


tguruswamy:
The install scripts (`configure`) assume that /usr/bin/python{3} are the correct binary to use for running backintime, and error out if that version is < 3.8. On systems with multiple python versions (e.g. Red Hat Enterprise Linux, SUSE Enterprise Linux/openSUSE Leap, Debian 10, Ubuntu LTS) this means backintime cannot be installed even if an appropriate python version is available (openSUSE Leap has python36 and python311, for example). More generally it means backintime cannot be reliably managed on systems with custom python installs, such as from conda or pyenv.

The install script should simply accept the full path to a python binary to be used, not only python / python3.
(Even better would be to use a standard python build system).

buhtz:
Dear Tejas,

thanks for your report and your feedback.

First of all: We do plan to migrate to a "standard python build system" and eliminate the make system (#1575). But this will take time and currently we do not have a schedule for this.

> The install scripts (`configure`) assume that /usr/bin/python{3} are the correct binary to use for running backintime

Can you point us to the specific location please and give us a line number. I was not able to find it in `configure`.

> and error out if that version is < 3.8

That is correct. We (as upstream) do support Python 3.8 and not older.
 
> On systems with multiple python versions (e.g. Red Hat Enterprise Linux, SUSE Enterprise Linux/openSUSE Leap, Debian 10, Ubuntu LTS) this means backintime cannot be installed even if an appropriate python version is available

I think you mix up some things here. I might be mistaken but Debian and Ubuntu do not offer multiple Python versions. Am I wrong? I don't know about the other distros but...

> (openSUSE Leap has python36 and python311, for example)

If this really is the case then please open a ticket there. Then the "openSUSE Leap" maintainer can contact us at upstream and tell us how we could modify BIT to fit SUSEs needs. Then we can discuss if this solution will violate other rules.

> More generally it means backintime cannot be reliably managed on systems with custom python installs, such as from conda or pyenv.

That is an often misunderstanding of conda/pyenv etc. A productive system do not have multiple Python versions. Conda & Co are for testing and development environments. BIT is an end-user application intended to run on a productive system.

> The install script should ...

As I told the `configure` do not point to a full path of the python binary. It just checks the version of the systems python.
I need more details to understand the real problem here.
If my understanding improved then I might be able to offer you a solution that do not violate our needs but also fit your needs.

[...]

PR #1602, opened 2024-01-08
Allow manual specification of python executable during build

tguruswamy:
common/configure and qt/configure now support `--python=PYTHON_PATH` option to manually specify the python executable. A full, absolute path is preferred but any valid path should work.

Previous --python and --python3 options also should still work, representing /usr/bin/python and /usr/bin/python3 respectively.

Tests pass on openSUSE Leap 15.5 and Tumbleweed.

This is the quick fix for #1574 .

### DIFF
--- a/common/configure
+++ b/common/configure
@@ -133,9 +133,12 @@ if [ -n "$unknown_args" ]; then
     echo "Unknown Arguments: $unknown_args"
 fi
 
-sed -e "s#^python3 #${PYTHON} #g" \
-    -e "s#^ssh-agent python3 #ssh-agent ${PYTHON} #g" \
-    -i $USR_BIN_FILES
+if [ -n "$(sed -e "s#^python3\? #${PYTHON} #gw /dev/stdout" -i $USR_BIN_FILES)" ]
+then
+    echo "Replacement of python path with \"${PYTHON}\" successful."
+else
+    echo "WARNING: Replacement of python path with \"${PYTHON}\" FAILED. Maybe you ran configure more than once?"
+fi
 
 #check languages
 mos=""
--- a/common/configure
+++ b/common/configure
@@ -17,12 +17,14 @@ USR_BIN_FILES="backintime backintime-askpass"
 
 usage () {
     echo "Usage:"
-    echo "$0 [--python | --python3]"
+    echo "$0 [--python | --python3 | --python=PYTHON_BINARY]"
     echo ""
     echo "--python"
     echo "\tuse 'python' to start Python3"
     echo "--python3"
     echo "\tuse 'python3' to start Python3"
+    echo "--python=PYTHON_BINARY"
+    echo "\tuse PYTHON_BINARY to start Python3"
 }
 
 addInstallFiles () {
@@ -113,7 +115,15 @@ onTravis () {
 unknown_args=""
 for arg in $*; do
     case $arg in
-        --python | --python3) PYTHON=$arg;;
+        --python=*)
+            PYTHON=$(echo $arg | cut -f2 -d'=')
+            ;;
+        --python3)
+            PYTHON="/usr/bin/python3"
+            ;;
+        --python)
+            PYTHON="/usr/bin/python"
+            ;;
         --help | -h) usage; exit 0;;
         *) unknown_args="$unknown_args $arg";;
     esac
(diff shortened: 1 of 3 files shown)


tguruswamy:
I note that you can no longer change the python path twice by rerunning configure -- run it once only. But hopefully there isn't anyone relying on that.

buhtz:
Thanks a lot for this contribution.
I am not into makefiles to review this PR.
Maybe we can ask on bit-dev for assistance?

aryoda:
@buhtz I will reviewing and testing this in the next days

> I note that you can no longer change the python path twice by rerunning configure -- run it once only. 

Do you mean a 2nd `./configure` does not change the `Makefile`'s Python3? Strange, I will test this on Ubuntu...

**Edit:** OK, it is because `sed` does not find `^python3` anymore to replace it (once it has been changed).
                  This is IMHO a source of problems that are hard to find.
                  Perhaps there is a better solution that allows changing Python with every `./configure` call.
                  @all Any proposals?

common/configure
```diff
@@ -123,14 +133,8 @@ if [ -n "$unknown_args" ]; then
...
 
-#patch python command
-#use 'python' or 'python3' to start Python Version 3.x
-case $PYTHON in
-    --python)   PYVERSION="" ;;
-    --python3)  PYVERSION="3";;
-esac
-sed -e "s/^python3\? /python${PYVERSION} /g" \
-    -e "s/^ssh-agent python3\? /ssh-agent python${PYVERSION} /g" \
+sed -e "s#^python3 #${PYTHON} #g" \
```
aryoda: Why is `#` instead of `/` used as substitution delimiter?
It does work surprisingly...

**Edit:** `man sed` says:

> cregexpc   Match lines matching the regular expression regexp.  The c may  be any character.

so I would have expected a leading backslash...n

common/configure
```diff
@@ -324,7 +328,7 @@ done
 
 # check python version
 PYTHON_VERSION_REQUIRED="3.8"
-PYTHON_VERSION_CURRENT=$(python${PYVERSION} --version | tr --delete 'Python ')
+PYTHON_VERSION_CURRENT=$(${PYTHON} --version | tr --delete 'Python ')
```
aryoda, line 334: beautiful code now, well done :smiley:

tguruswamy:
> **Edit:** OK, it is because `sed` does not find `^python3` anymore to replace it (once it has been changed). This is IMHO a source of problems that are hard to find. Perhaps there is a better solution that allows changing Python with every `./configure` call. @ALL Any proposals?

Yes, that's right. The proper way to do this (c.f. autotools) would be to have a template file ("backintime.in") which is used to create the real "backintime" every time configure is run. This would be a more invasive change, and may not be worth it to implement right now if you are rewriting the build system anyway.

common/configure
tguruswamy: Indeed, any character can be used as long (as it doesn't appear in `regexp`, of course) -- I picked `#` since that was already used in qt/configure for the same purpose.

aryoda:
> Yes, that's right. The proper way to do this (c.f. autotools) would be to have a template file ("backintime.in") which is used to create the real "backintime" every time configure is run. This would be a more invasive change, and may not be worth it to implement right now if you are rewriting the build system anyway.

I fully agree and given the fact that the python version is not changed on the build computer anymore (normally) it is IMHO a valid approach.

It would be great if you place just a short comment in the code as hint about this restriction. THX :-)

tguruswamy:
Here's a compromise to avoid rewriting too much -- a version which prints a warning if the sed python path replacement fails. Feel free to change the text to something you like better.

I removed the `ssh-agent` pattern because I could not find anywhere in the code where it might apply.

buhtz:
Dear @Fantu ,
as Debian co-maintainer can you please have a short look on this PR and tell is if it would negative influence the packaging process on your site?
Thanks in advance
Christian

aryoda:
> Here's a compromise to avoid rewriting too much -- a version which prints a warning if the sed python path replacement fails. Feel free to change the text to something you like better.

Perfect as is now!

> I removed the `ssh-agent` pattern because I could not find anywhere in the code where it might apply.

Just checked it, you are right, `ssh-agent` is only called from within our Python source code so no Python version must be specified:

https://github.com/bit-team/backintime/blob/2df7731f1eeab6f2c734f0dd48a75996adef9cd8/common/sshtools.py#L255-L305

So this is useless legacy code (`git blame` dates this back to 2015)

aryoda (approved):

Fantu:
> Dear @Fantu , as Debian co-maintainer can you please have a short look on this PR and tell is if it would negative influence the packaging process on your site? Thanks in advance Christian

Probably not, but I'm not sure.
You could compare the contents of the generated packages for example to see the differences, the latest version in Debian compared with the one generated by adding only this patch. Should be possible with [debdiff](https://manpages.debian.org/unstable/devscripts/debdiff.1.en.html).
I'm probably too busy in these days, if you can check it, if you have difficulty I'll try to do it when I can.

buhtz (approved):

aryoda:
@tguruswamy I have merged your PR now into our `dev` branch.
THX a lot for your excellent work helping to improve BiT! :heart:

PR #1611, opened 2024-01-20
Complete Slovenian translation

LolekLiam:
Completed the Slovenian translation. I still need to check the translations, marked as translated, because i noticed that some of them are not translated very well.

Edit: Now most if not all translations are fixed.


buhtz:
Info to the team: Some translators want to have their own PRs to improve their profile at Microsoft GitHub.

PR #1614, opened 2024-01-22
Use absolute path to the python executable now by default (closes #1574)

### LINKED ISSUE #1574: Backintime install script only uses python / python3 even if other versions are available

aryoda:
All starter scripts do use an absolute path to the python executable by default now via common/configure and qt/configure (see #1574)

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -14,6 +14,7 @@ Version 1.4.2-dev (development of upcoming release)
 * Build: Activate PyLint warning W1401 (anomalous-backslash-in-string).
 * Build: Add codespell config.
 * Build: Allow manual specification of python executable (--python=PYTHON_PATH) in common/configure and qt/configure
+* Build: All starter scripts do use an absolute path to the python executable by default now via common/configure and qt/configure (#1574)
 * Build: Install dbus configuration file to /usr/share not /etc (#1596)
 * Build: `configure` does delete old installed files (`qt4plugin.py` and `net.launchpad.backintime.serviceHelper.conf`) that were renamed or moved in a previous release (#1596)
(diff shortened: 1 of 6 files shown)


buhtz (approved):

Douglas1949:
Here are the screenshots we were talking about - attached; I hope you can see them?
Screenshot from 2024-01-25 12-28-48_BiT_Auto-remove-tab
Screenshot from 2024-01-25 12-30-07_BiT_write-access-issue

aryoda:
> Here are the screenshots we were talking about - attached; I hope you can see them?

Yes, it I can see them now, THX a lot!

I have copied your screen shots to #1617 which is the new correct issue for our discussion (not this one), let's continue our conversation there...

aryoda:
@Douglas1949 Could you please follow [my step-by-step instructions in this comment](https://github.com/bit-team/backintime/issues/1617#issuecomment-1910779484) to diagnose the smart remove problem? THX :-)

Douglas1949:
Thank-you very much.
I will do this tomorrow when I am fresh.
Sincerely
Douglas / Duncan

Issue #1587, opened 2023-12-14
"rsync error 23" (Partial transfer due to errors)


aryoda:
This issue is meant to centralize all "error 23" issues to discuss a common solution...

## Background

BiT Version 1.4.0 (2023-09-14) introduced the **evaluation of `rsync` exit codes for better error recognition**:

https://github.com/bit-team/backintime/blob/726bffa624d9a6da4ebd13ceb40721e1290e1eaf/CHANGES#L32-L42

Before this change `rsync` errors reported only via exit code were hidden and left users with the feeling that everything worked well (but didn't).

## Collected reasons for `rsync` exit code 23

The final error message in the snapshot log is always

> [I] Take snapshot (rsync: rsync error: some files/attrs were not transferred (see previous errors) (code 23) at main.c(1338) [sender=3.2.7])
> [E] 'rsync' ended with exit code 23: See 'man rsync' for more details

Here are typical snapshot log entries causing exit code 23 (shall be extended here to be extended step by step to get a holistic picture):

* `[I] Take snapshot (rsync: symlink has no referent: "/home/user/Documents/dead-link")`
* `[E] Error: rsync: [sender] send_files failed to open "/home/user/Documents/root_only_file.txt": Permission denied (13)`

Other known reasons: See https://github.com/bit-team/backintime/issues/1587#issuecomment-1856400319

## Current impact

Some of the reasons for exit code 23 may be out of control of the BiT user or are OK to be ignored since not considered as "severe" (eg. dead links).

Despite that BiT snapshots do always fail with an error now.

Even though `rsync` treats exit code 23 always as error the BiT user should have the freedom to treat some reasons for exit code 23 as INFO or WARNING only...

## Possible solutions

1. Add exit code 23 to the ignore list (as before) and let the user think everything went OK (until looking into the snapshot log)
    https://github.com/bit-team/backintime/blob/726bffa624d9a6da4ebd13ceb40721e1290e1eaf/common/snapshots.py#L1247-L1252

1. Find and use `rsync` options so that we can ignore minor exit code 23 reasons:
    https://unix.stackexchange.com/questions/598736/how-to-skip-copying-broken-links-with-rsync-copy-links

1. Introduce a new snapshot result state "Warning" (currently we have only "OK" and "with errors)
    https://github.com/bit-team/backintime/blob/726bffa624d9a6da4ebd13ceb40721e1290e1eaf/common/snapshots.py#L1110-L1131

## Challenges

[...]

Issue #1555, opened 2023-10-28
Rsync error 23 (partial transfer due to error) undiscovered


Hizoka76:
Hi,

since the **1.4.1-1** update, all my backups end up with **error 23**.

The backup crashes very quickly and never on the same files.

I was supposedly in version **1.3.3-3** before the update via the ppa **https://launchpad.net/~bit-team/+archive/ubuntu/stable**.

I have backups on an external hard disk that work.

I tried to run the command **backintime backup** :
```
Back In Time
Version: 1.4.1-1

Back In Time comes with ABSOLUTELY NO WARRANTY.
This is free software, and you are welcome to redistribute it
under certain conditions; type `backintime --license' for details.

INFO: Lock
INFO: Take a new snapshot. Profile: 1 Profil principal
INFO: Call rsync to take the snapshot
[...]

PR #1621, opened 2024-01-26
Work around for rsync error 23 (part of #1587) + new "experimental" snapshot log filter

aryoda:
* Work around: Relax `rsync` exit code 23: Ignore instead of error now (part of #1587)

* Feature: Exclude 'SingletonLock' and 'SingletonCookie' (Discord) and 'lock' (Mozilla Firefox) files by default (part of #1555)

* Feature (experimental): Add new snapshot log filter `rsync transfer failures (experimental)` to find them easier (they are normally not shown as "error").

  This feature is experimental because it is based on hard-coded error message strings in the `rsync` source code and may possibly not find all rsync messages or show false positives.

  The filter is meant only to support humans, **not** to automatically recognize errors when taking a snapshot!

  image

  The filter recognizes snapshot log entries that would otherwise require intensive human search in long logs
  because they are hidden in `[I]`nfo log entries, eg.:

  ```
  [I] Take snapshot (rsync: symlink has no referent: "/home/user/Documents/dead-link
  [I] Schnappschuss erstellen (rsync: IO error encountered -- skipping file deletion)
  [I] Schnappschuss erstellen (rsync: rsync error: some files/attrs were not transferred (see previous errors) (code 23) at main.c(1333) [sender=3.2.3])
  [I] Take snapshot (rsync: rsync error: some files/attrs were not transferred (see previous errors) (code 23) at main.c(1333) [sender=3.2.3])
  ```

Changelog is updated too...

### DIFF
--- a/common/snapshotlog.py
+++ b/common/snapshotlog.py
@@ -57,43 +57,50 @@ class LogFilter(object):
              INFORMATION:       re.compile(r'^(?:\[I\]|[^\[])'),
              ERROR_AND_CHANGES: re.compile(r'^(?:\[E\]|\[C\]|[^\[])'),
              RSYNC_TRANSFER_FAILURES: re.compile(
+                 # All links to rsync's source reference the commit 2f9b963 from Jun 27, 2023 (most-recent commit on "master" as at Jan 28, 2024)
                  r'.*(?:'
-                 r'Invalid cross-device link'
-                 r'|symlink has no referent'
-                 r'|readlink_stat\(.?\) failed'
-                 r'|link_stat .* failed'
-                 r'|receive_sums failed'
-                 r'|send_files failed to open'
-                 r'|fstat failed'
-                 r'|read errors mapping'
-                 r'|change_dir .* failed'
-                 r'|skipping overly long name'
-                 r'|skipping file with bogus \(zero\) st_mode'
-                 r'|skipping symlink with 0-length value'
-                 r'|cannot convert filename'
-                 r'|cannot convert symlink data for'
-                 r'|opendir .* failed'
-                 r'|filename overflows max-path len by'
-                 r'|cannot send file with empty name in'
-                 r'|readdir\(.*\)'
-                 r'|cannot add local filter rules in long-named directory'
-                 r'|failed to re-read xattr'
-                 r'|Skipping sender remove of destination file'
-                 r'|Skipping sender remove for changed file'
-                 r'|could not make way for'
-                 r'|system says the ACL I packed is invalid'
-                 r'|recv_acl_access: value out of range'
-                 r'|recv_acl_index: .* ACL index'
-                 r'|Create time value of .* truncated on receiver'
-                 r'|FATAL I/O ERROR: dying to avoid a \-\-delete'
-                 r'|IO error encountered'  # TODO reported in an issue but is this string really contained in the rsync source code?
-                 r'|some files/attrs were not transferred'  # TODO reported in an issue but is this string really contained in the rsync source code?
-                 r'|temporary filename too long'
-                 r'|No batched update for'
-                 r'|recv_files: .* is a directory'
-                 r'|no ftruncate for over-long pre-alloc'
-                 r'|daemon refused to receive'
-                 r'|get_xattr_data: lgetxattr'
+                 r'Invalid cross-device link'     # not directly contained in rsync's source code but may be catched and passed through as-is
+                 r'|symlink has no referent'      # https://github.com/WayneD/rsync/blob/2f9b963abaa52e44891180fe6c0d1c2219f6686d/flist.c#L1281
+                 r'|readlink_stat\(.?\) failed'   # https://github.com/WayneD/rsync/blob/2f9b963abaa52e44891180fe6c0d1c2219f6686d/flist.c#L1294
+                 r'|link_stat .* failed'          # https://github.com/WayneD/rsync/blob/2f9b963abaa52e44891180fe6c0d1c2219f6686d/flist.c#L1810
+                 r'|receive_sums failed'          # https://github.com/WayneD/rsync/blob/2f9b963abaa52e44891180fe6c0d1c2219f6686d/sender.c#L347
+                 r'|send_files failed to open'    # https://github.com/WayneD/rsync/blob/2f9b963abaa52e44891180fe6c0d1c2219f6686d/sender.c#L361
+                 r'|fstat failed'                 # https://github.com/WayneD/rsync/blob/2f9b963abaa52e44891180fe6c0d1c2219f6686d/sender.c#L373
+                 r'|read errors mapping'          # https://github.com/WayneD/rsync/blob/2f9b963abaa52e44891180fe6c0d1c2219f6686d/sender.c#L435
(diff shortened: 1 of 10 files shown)


common/snapshotlog.py
```diff
@@ -38,18 +38,65 @@ class LogFilter(object):
...
+    # RegExp syntax see: https://docs.python.org/3.10/library/re.html#regular-expression-syntax
+    # (?:...) = the matched substring cannot be retrieved in a group (non-capturing)
     REGEX = {None:              None,
              NO_FILTER:         None,
              ERROR:             re.compile(r'^(?:\[E\]|[^\[])'),
              CHANGES:           re.compile(r'^(?:\[C\]|[^\[])'),
              INFORMATION:       re.compile(r'^(?:\[I\]|[^\[])'),
-             ERROR_AND_CHANGES: re.compile(r'^(?:\[E\]|\[C\]|[^\[])')}
+             ERROR_AND_CHANGES: re.compile(r'^(?:\[E\]|\[C\]|[^\[])'),
+             RSYNC_TRANSFER_FAILURES: re.compile(
```
buhtz, line 59: These strings are from the rsync sources?
Maybe can you add a link to the related file in the rsync repo here as a comment?

buhtz:
The TravisCI job for ppce64le on Python 3.8 is failing. This should not happen. I can't see in the logs why this happens.

This was also the case with PR #1614. But we did not recognized because we are used to the fact that currently TravisCI do not work. There is another job "ppc64le Python 3.12" (the latest Python!) which do not work because TravisCI still has no Python interpreter with that version for this architecture installed. They do work on it.

First of all within this PR the travis.yaml file should modified that way that the known problematic job ""ppc64le Python 3.12" is inactive.

Why "ppc64le Python 3.8" is failing needs further investigation but I assume this is somehow because of PR #1614.

common/snapshotlog.py
aryoda: Great idea (and lot of work ;-) but I will enrich this to make maintenance easier in the future.
Question is: Permalink to the examined rsync version or dynamic link (may change).
I guess the purpose is compare underlying source code of the implemented regexp against the most recent rsync version so I will add permalink.

aryoda:
> First of all within this PR the travis.yaml file should modified that way that the known problematic job ""ppc64le Python 3.12" is inactive.
> 
> Why "ppc64le Python 3.8" is failing needs further investigation but I assume this is somehow because of PR #1614.

I will look into this and fix it

**Edit:** Ah, I just saw you are already fixing this, so I will wait (meanwhile trying to understand the impact of #1614 on TravisCI for ppc64le)

buhtz:
Mhm... I still have no idea and can not see the error message in the TravisCI output.
grafik

I removed pytest on my own machine to make sure that "unittest" package is used only. Then configure && make unitest-v was not able to reproduce the problem on my Debian 12.
Then I compared the generated Makefiles between TravisCI and my Debian 12.

On Debian:

    /usr/bin/python3 -m unittest -v -b test/test_applicationinstance.py

On Travis

    coverage run -p -m unittest -v -b test/test_applicationinstance.py

This makes sense and does not seems wrong.

But I do think that because of #1614 now "/usr/bin/python3" is used instead of "python3" in the previous versions.
But I don't see how this could make problems.

I am stopping here for this night and going to bed. I have no further ideas yet.

aryoda:
> But I do think that because of #1614 now "/usr/bin/python3" is used instead of "python3" in the previous versions.
> But I don't see how this could make problems.

I have injected a relative path again in `common` and `qt` via `./configure --python=python3` and it works again.

I have not yet tested it with an absolute path again (`configure` was not called for `qt` - perhaps this was the root problem).

It looked like `pytest` did work in `qt` but `after_success` did not work (`coverage combine` + `coveralls`).

Really strange...

I can do more tests Sa. night...

buhtz:
I tried to run the "coverage ..." commands on my local machine. There is no error.
On major difference to the TravisCI Ubuntu could be that Travis do not use Ubuntu's own Python but installs a specific one from a non-ubuntu source.

OK, I asked the community about how to make "make" more verbose: [r/pythoncoding](https://www.reddit.com/r/pythoncoding/comments/1ac68hn/make_hides_error_messages_on_travisci/) and [debianforum.de](https://debianforum.de/forum/viewtopic.php?t=188948).

I also asked the "coverage.py" maintainer and community about how coverage decide which Python interpreter it uses: [Python discuss](https://discuss.python.org/t/coverage-py-which-python-binary-is-used).

buhtz:
The `make -d` give some information.

overage run -p -m unittest -v -b test/test_applicationinstance.py

```
Putting child 0x8e31851a7f0 (unittest-v) PID 4469 on the chain.
Live child 0x8e31851a7f0 (unittest-v) PID 4469 
WARNING: 'import keyring' failed with: ModuleNotFoundError("No module named 'keyring'")
test_autoExit_other_running_process (test.test_applicationinstance.TestApplicationInstance) ... ok
```

Keyring is missing. I assume that "coverage.py" does use an Python interpreter different from the one used in this environment. Working on it ...

buhtz:
:boom: 
Our "bit-team" account on TravisCI don't have credits anymore. I contacted the support.

I have an idea about the problem but I am not sure.

There are ModuleNotFound errors about "keyring" and "packaging". Not sure how this could be triggered by our last binary-path-fix. It could be that these packages are not installed inside the virtual environment that is used.
Via "coverage debug sys" I was able to verify that "coverage" do use the Python binary that is installed inside the virtual environment.

Maybe it will help when installing the missing packages explicit via "pip".

I also modified the "coverage" call into "{$PYTHON} -m coverage" as an experiment. But to my knowledge it doesn't matter. Coverage is used only on Travis and it always use the virtual environments own python interpreter by default. So this modification shouldn't change something and could be undone.

buhtz:
Credits are back. Seems that people also working in the weekend at Travis.

common/snapshotlog.py
buhtz: Not sure about permalink or dynamic link. The purpose for me would be the opportunity to compare the current latest/stable rsync version source with the strings we do use here.
If we stick to this solution than we can add a little helper script (as manual step in the release process) doing this checkup for us and alarm if the strings have modified.

aryoda:
> I have an idea about the problem but I am not sure.

Really weird..

I'd say we should use the working `./configure --python=python3` for now since it is not the goal the test the infrastructure (which may have a problem) but BiT.

buhtz:
> I'd say we should use the working `./configure --python=python3` for now since it is not the goal the test the infrastructure (which may have a problem) but BiT.

Yes, good point. It works and doesn't matter that I don't know why it works. :laughing:

aryoda:
> > I'd say we should use the working `./configure --python=python3` for now since it is not the goal the test the infrastructure (which may have a problem) but BiT.
> 
> Yes, good point. It works and doesn't matter that I don't know why it works. 😆

I also really want to understand what's going wrong :smile: Since it only happens on one platform (ppc64le) I suspect a packaging issue on this specific Travis CI platform.

But my point is:

- I could merge this PR now since it works and shall be contained in the upcoming release (unless you want to do more testing on Travis CI with this PR - which is totally OK for me - I think the release can wait a few more days)

buhtz (approved):

common/snapshotlog.py
buhtz: Oh, damn! I didn't mean commenting each one of this message. 🤣 
Just one single comment. Very nice!

PR #1624, opened 2024-01-30
BiT Release v1.4.3 (Jan 30, 2024)

aryoda:
This PR contains the release version v.1.4.3


buhtz:
I remember the last time I was confused about that.
But isn't this release version "1.4.2" ? The latest stable release is "1.4.1".

README.md
buhtz, line 1: The problems-toc entry for qt5-probing is missing (near line 80).
This should work:

    - [`qt5_probing.py` may hang with high CPU usage when running BiT as `root` via `cron`](#qt5_probingpy-may-hang-with-high-cpu-usage-when-running-bit-as-root-via-cron)

aryoda:
> I remember the last time I was confused about that.
> But isn't this release version "1.4.2" ? The latest stable release is "1.4.1".

Yes, may be confusing, but I wanted to avoid confusion of 1.4.2-dev with 1.4.2

I can change this of course...

**Edit:** I think "not every BiT version must become a release so gaps in release version numbers are OK IMHO. But we can decide otherwise, I have no stake in this

emtiu (approved):

buhtz:
I have no hard opinion about it and will follow every decision.

But I would suggest to follow [semantic versioning point 9](https://semver.org/#spec-item-9).

And by definition `0-dev` is lower then `0`:
```
>>> from packaging import version
>>> version.parse("1.4.2") > version.parse("1.4.2-dev")
True
>>>
```

qt/app.py
```diff
@@ -88,7 +88,8 @@
                           QSortFilterProxyModel,
                           QDir,
                           QSize,
-                          QUrl
+                          QUrl,
+                          pyqtRemoveInputHook,
```
buhtz, line 92: Found this accidentally in context of PR #1625 and fixed it. Credits to @JATIN-RAGHAV.

qt/app.py
```diff
@@ -443,6 +444,9 @@ def _create_actions(self):
         """
 
         action_dict = {
+            # because of "icon"
+            # pylint: disable=undefined-variable
```
buhtz, line 448: Just for explanation: This "disable" is active only in this block (inside the `{ ... }`) and targets the `icon` package. That package is a "problem" because we do some unusual import hacks with it.

buhtz (approved):

PR #1625, opened 2024-01-31
fix: Consistent tooltips for menu entries and toolbar buttons

JATIN-RAGHAV:
I made the suggested changes for both snapshot and restore button. If it is not correct then please point out the errors. I wil try to fix them too.

Fix #1605

PR #1631, opened 2024-02-02
Build: Pylint improvement & Travis using PyQt

### LINKED ISSUE #1626: PyLint: Run on all py-files before reporting errors
Situation: Currently pylint runs on each py-file separately. If there is an error in one file the unittest fails without running pylint on the rest of the py files.

Problem: So there might be more pylint related errors in other files that are not shown in the output.

Solution: Run pylint once on all relevant files. Or collect the output of each relevant run and display it at the end.

buhtz:
- PyLint unit test does check all files before reporting errors (Fix #1626)
- Activate PyLint "undefined-variable" error (inspired by PR #1625)
- TravisCI now using PyQt and is prepared to run Qt related unit test except on "ppc64le" architecture.

## About PyQt on TravisCI
Spent to much hours into Travis config experiments. 😪 I added some comments into the travis config but also try to explain it here. The final solution is to install PyQt5 via pip only at architectures different from "ppc64le" and execute "pytest" for the qt module also only on none-ppce64le architectures.

PyQt is needed on TravisCI when we start using Qt tests. But it is also needed today because PyLint need to use it to determine variable-names. Without PyQt installed PyLint would throw "undefined-variable" errors on our qt code.
The primary problem is that PyQt on PyPi (installed via "pip") is not available for architecture "ppc64le". But Ubuntu support that architecture and offer such a package. But Ubuntu packages can not be used inside the Python virtual environment that is used by TravisCI. The option "system-site-packages" might be a workaround because it allow an isolated Python interpreter to access system wide Python packages (installed via "apt-get"). But TravisCI has the restriction that "system-site-packages" only works for the Python version shipped with the Ubuntu (3.8 in our case).
I also experimented about to use "system-site-packages" only for the "Python-3.8-ppce64le-job". But I was not able to make it work.
So many "But"'s...

### DIFF
--- a/common/test/test_lint.py
+++ b/common/test/test_lint.py
@@ -1,10 +1,7 @@
 import unittest
 import pathlib
 import subprocess
-from importlib import metadata
 from typing import Iterable
-
-# PACKAGE_NAME = 'buhtzology'
 
 
 class MirrorMirrorOnTheWall(unittest.TestCase):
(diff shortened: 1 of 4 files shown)


common/test/test_lint.py
```diff
@@ -16,24 +13,24 @@ def _collect_py_files(self) -> Iterable[pathlib.Path]:
...
+        pyproject.toml.
         """
-        p = pathlib.Path.cwd()
+        path = pathlib.Path.cwd()
 
         # Make sure we are inside the test folder
-        if p.name in ['qt', 'common']:  # happens e.g. on TravisCI
-            p = p / 'test'
+        if path.name in ['qt', 'common']:  # happens e.g. on TravisCI
+            path = path / 'test'
```
aryoda, line 22: funny code ;-)

aryoda:
> PyQt is needed on TravisCI when we start using Qt tests.

I am looking forward to find out if Qt5/6 unit tests of GUI parts are working at all on Travis CI since I guess that no desktop environment can be "started" (no xorg/X11 or Wayland)...

aryoda (approved):

PR #1633, opened 2024-02-02
Fix #1630 (make install tries to remove old file using absolute path)

### LINKED ISSUE #1630: `make install` uses absolute path without $(DEST) to delete files outside of the staging directory
I have introduced a clean-up code in the `Makefile` to remove installed renamed files which could cause havoc when a new BiT version is installed via `make install`.

@uffe has now reported that `make install` may cause an unwanted side effect ("messing outside the staging dir"):

https://aur.archlinux.org/pkgbase/backintime#comment-954580

```
#clean-up installed old files that were renamed or moved in later BiT versions
rm -f /etc/dbus-1/system.d/net.launchpad.backintime.serviceHelper.conf
rm: cannot remove '/etc/dbus-1/system.d/net.launchpad.backintime.serviceHelper.conf': Permission denied
make: *** [Makefile:14: install] Error 1
```

aryoda:
Fixes #1630

Tested with `make DESTDIR="/home/user/temp/bit" install` before and after the fix

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -2,6 +2,7 @@ Back In Time
 
 Version 1.4.4-dev (development of upcoming release)
 * TODO
+* Fix bug: `make install` ignored $(DEST) in file migration part (#1630)
 
 Version 1.4.3 (2024-01-30)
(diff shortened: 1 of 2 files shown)


buhtz (approved):
I am still a rookie when it comes to make. But to my understanding this looks OK.

PR #1635, opened 2024-02-03
Build: PyLint unit test skipped if not installed except on TravisCI

### LINKED ISSUE #1634: pylint unit test `test_lint.py` should be executed "on demand" only (not as installation test on clients)
In BiT relese 1.4.3 the `pylint` unit test is included an is always executed to check the source code (static code analyzer).

This is not required for users and on distros who install BiT from the source code via `make` and execute the unit tests.

In case of ARCH linux it creates an unnecessary dependency and slows down the installation process:

https://aur.archlinux.org/pkgbase/backintime#comment-954673

I suggest to
- either make the execution of this unit test optional (not by default)
- or offer a way to disable it (a small documentation may suffice if this works)

A new `Makefile` target `make dev-test` similar to `make test` vs. `make test-v` could also be a valid solution
which is the only target that executes the linter (meant for us as developer)

@buhtz Do you have an idea?

buhtz:
The PyLint unit test is executed only when PyLint is installed. Exception: On TravisCI the test is executed always.

Fix #1634


buhtz:
Feel free to improve my English. I am not happy with it.

CHANGES
```diff
@@ -1,6 +1,7 @@
 Back In Time
 
 Version 1.4.4-dev (development of upcoming release)
+* Build: PyLint unit test is skipped if PyLint isn't installed except TravisCI will raise an error (#1634)
```
```suggestion
* Build: PyLint unit test is skipped if PyLint isn't installed, but will always run on TravisCI (#1634)
```

common/test/test_lint.py
```diff
@@ -1,8 +1,15 @@
...
+import os
 import pathlib
 import subprocess
+import shutil
 from typing import Iterable
 
+ON_TRAVIS = os.environ.get('TRAVIS', '') == 'true'
+PYLINT_AVIALBE = not shutil.which('pylint') is None
+PYLINT_REASON = ('Using PyLint is mandatory on TravisCI and otherwise '
+                 'optional to systems where PyLint is available.')
```
emtiu:
```suggestion
PYLINT_REASON = ('Using PyLint is mandatory on TravisCI, on other systems'
                 'it runs only if `pylint` is available.')
```

qt/test/test_lint.py
```diff
@@ -1,8 +1,15 @@
...
+import os
 import pathlib
 import subprocess
+import shutil
 from typing import Iterable
 
+ON_TRAVIS = os.environ.get('TRAVIS', '') == 'true'
+PYLINT_AVIALBE = not shutil.which('pylint') is None
+PYLINT_REASON = ('Using PyLint is mandatory on TravisCI and otherwise '
+                 'optional to systems where PyLint is available.')
```
emtiu:
```suggestion
PYLINT_REASON = ('Using PyLint is mandatory on TravisCI, on other systems'
                 'it runs only if `pylint` is available.')
```

emtiu (commented):
Done :)

PR #1640, opened 2024-02-09
Build: Display last commit hash in "*-dev" version strings

### LINKED ISSUE #1637: Identify BiT dev build by git hash
BiT development builds report version as "backintime 1.4.4-dev" which actually provides little identification.

I propose appending the last commit id to "backintime 1.4.4-dev.c65fc91" so one can tell that his dev version needs updating

buhtz:
BIT uses the last commit hash in its version string if it is a "*-dev" version.

An approach to discuss of course. Because of circular imports I saw no other way as to create a new `version.py`. I assume we can fusion such files later back when there is more structure in the whole project. Important to notice is that now you can do `backintime.__version__` nearly everywhere in the code.

image

Close #1637


aryoda:
I feel uncomfortable to say this after seeing to much time and work invested into this PR but it is quite a comprehensive code change "just" to inject the Git commit hash.

I would prefer to change eg. only the `updateversion.sh` (by appending the commit number to the version number read from the `VERSION` file) .

I guess this way the same result could be achieved by only changing (adding) a few lines of code...

**Edit:** This would also avoid to require another file to be installed (`version.py`)

buhtz:
No problem. No need to feel uncomfortable. 👍 Every step is an improvment no matter if it gets merged or not.

You mean the "updateversion.sh" would determine if it is a "*dev" version and in this case adds the git-hash to it?

**EDIT**: See my last commit. This way?

emtiu (approved):

common/version.py
```diff
@@ -0,0 +1,47 @@
...
+        bit_path = pathlib.Path(tools.backintimePath('common'))
+        git_json_file = bit_path / 'git-version.json'
+
+        # Git info file available?
+        if git_json_file.exists():
+            git_info = json.loads(git_json_file.read_text('utf-8'))
+
+    # git info from current repo or info file
+    if git_info:
+        return f'{__version_base__}.{git_info["hash"][:8]}'
```
aryoda: This "magic number" is the max length of the hash we append to our version number I guess?
A constant instead of "8" would make this more clear...

aryoda:
What I really like now is that the version number is in `version.py` instead of somehow hidden in `config.py`...

> You mean the "updateversion.sh" would determine if it is a "*dev" version and in this case adds the git-hash to it?

Yes

> **EDIT**: See my last commit. This way?

Yes but now the git hash logic is duplicated in `configure` and `updateversion.sh` (and even in `tools.get_git_repository_info`) - not really DRY :wink: 

Also it is IMHO difficult to recognize for developers _"where `git-version.json` is coming from"_:

- The base version number is now a variable in `version.py`
- `make install` creates a `git-version.json`
- This file is installed into `/share/backintime/common`
- `updateversion.sh` does update all version numbers (in `version.py`, man pages and sphinx docs)
   incl. the git hash if the base version number ends with `-dev`
- BiT shows the version number
  - as-is if no `-dev` postfix is found
  - with the git hash ad-hoc appended if `-dev` is part of the version number and BiT is started directly from the repo
  - with the git hash from `git-version.json` appended if `-dev` is part of the version number and the installed BiT version is started

**I would prefer a slightly simpler approach with the same result:**

- `updateversion.sh` is good as it is in the PR (injects the git hash)
- `configure` should create only an `updateversion.sh` call in `make install` instead of creating a `git-version.json`
- `version.py` should then be simplified to use the `__version_base__ ` as-is (since it already contains the git hash if needed)

Anyhow the PR is OK and improves the situation so it is OK to merge it if you don't agree.

PS: Showing the git hash when BiT is started directly from within the git repo is not neccessary IMHO (no value added but it could stay in the code).

buhtz:
> Yes but now the git hash logic is duplicated in `configure` and `updateversion.sh` (and even in `tools.get_git_repository_info`) - not really DRY 😉

Sorry, this wasn't meant to be a complete solution. I will come up with a clearer PR the next days.

buhtz:
>  `configure` should create only an `updateversion.sh` call in `make install` instead of creating a `git-version.json`

Is this correct? Each call of `sudo make install` will update the version read from `VERSION` including the git hash? I try to imagine what the advantage is of this procedure.
This is currently my only one open question.

Beside of this last open question I would summarize:
 - `config.py`: `Config.VERSION` is removed and replaced by `version.py::__version__`.
 - `updateversion.sh`: Add the current git hash to the version string before it updates all versions in the repo.
 - `version.py` is just a temporary solution. It is not elegant but one step towards a better structured code base. IMHO we need to remove (or replace) code fragments from monolithic files like `config.py` into smaller files before we glue them together again. This might be a long year process. But it is IMHO easier and more secure than refactoring all at once. Another example of such an approach is the file `qt/qttools_path.py`,
 - `common/configure` & `qt/configure`: The file `VERSION` is no longer installed. Via `grep` I verified that it is not used by an installed BIT instance.
 - Function `get_git_repository_info()` moved from `diagnostics.py` to `tools.py`.
 - Function `tools.py::gitRevisionAndHash()` removed. It was a duplicate and buggy (e.g. can not handle `/` in branch names). I created `get_git_repository_info()` before I realized that this function exists.

updateversion.sh
```diff
@@ -47,9 +53,9 @@ update_sphinx_config () {
       -i $1
 }
 
-update_config () {
+update_app () {
```
aryoda: Are there future plans to extend this function with other updates? If not I would name it `update_app_version`...

common/configure
```diff
@@ -196,7 +196,6 @@ addInstallFile "../AUTHORS"          "/share/doc/backintime-common"
 addInstallFile "../LICENSE"          "/share/doc/backintime-common"
 addInstallFile "../README.md"        "/share/doc/backintime-common"
 addInstallFile "../TRANSLATIONS"     "/share/doc/backintime-common"
-addInstallFile "../VERSION"          "/share/doc/backintime-common"
```
aryoda, line 199: Should we keep this file in the installation target for "backwards compatibility" (though I am not sure if anyone is using it at all)

common/diagnostics.py
```diff
@@ -55,7 +56,7 @@ def collect_diagnostics():
 
     # Git repo
     bit_root_path = Path(tools.backintimePath(""))
-    git_info = get_git_repository_info(bit_root_path)
+    git_info = tools.get_git_repository_info(bit_root_path)
```
aryoda, line 59: really nice solution!

aryoda (approved):

aryoda:
Really good implementation, totally DRY and our current build system (`make install`) is responsible for updating the version number now (at least for the `dev` version).

For automatically updating the base version number for PRD releases we need another build system (as you have proposed).

THX for investing your time into that!

common/configure
buhtz: I am 99% sure. ;) BIT does not use it.
I'll ask at Debian and Arch if the distro maintainers do use it. But I assume they are not.
My assumption is that this file appeared in the release just out of habit.

Dear @graysky2 ,
dear @Fantu ,
as distro maintainers. Are you aware of that file named `VERSION` in BIT? Do you use it or need it somehow?

buhtz:
I'll wait on feedback from Debian and Arch about the `VERSION` file before merging.

Fantu:
From d/rules of Debian packaging is not used: https://salsa.debian.org/jmw/pkg-backintime/-/blob/debian/debian/rules?ref_type=heads
So should be used only by BIT internally

graysky2:
> as distro maintainers. Are you aware of that file named VERSION in BIT? Do you use it or need it somehow? 

Not to my knowledge, deleting these have no ill-effects as far as I can tell.

```
/usr/share/doc/backintime-common/VERSION
/usr/share/doc/backintime-qt/VERSION
```

PR #1642, opened 2024-02-14
Build: Add AppStream meta data

buhtz:
This PR does add meta data in AppStream format for the "qt" component.

This will fix a [packaging warning on Debian](https://appstream.debian.org/sid/main/issues/backintime-qt.html) (and possible other distros).

[AppStream](https://www.freedesktop.org/wiki/Distributions/AppStream/) (by freedesktop.org) is a format to describe software packages without technical details to users. The intention is to be independet from a specific operating system. AppStream is used in packaging software (e.g. synaptics) and "software centers" (e.g. "gnome-software") and other app-store-like applications. The target audience are desktop users. Beside fixing the Debian warning it increase the visibility of BIT in several software cataloges.

- Please carefully check the content of the `<summary>` and `<description>` tags in the xml file. Keep the target audience in mind. The text shouldn't be to nerdy. The syntax is not Markdown and not HTML. There is just a small subset of HTML tags. No links allowed for example.

- ~I will investigate further if and how to treat that file in our make system.~ Done.

- Localization via GNU gettext and integration into Weblate is possible. I will do this in another PR also including localization of our *.desktop files.

Here you see how this meta data would be rendered with Gnome Software Center (via `gnome-software --show-metainfo *.xml`)
bit_gnomesoftware_example


io.github.bit_team.back_in_time.gui.xml
```diff
@@ -0,0 +1,88 @@
...
+    Docu:
+        https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html
+
+    Validate (lint) on shell:
+        appstreamcli validate *xml
+-->
+
+
+<component type="desktop-application">
+  <id>io.github.bit_team.back_in_time.gui</id>
```
buhtz: There are strict and hard rules about the creation of such an `<id>`. It is derived from an existing website.
We could also use `io.readthedocs.*` or `org.codeberg.translate.*`. But the current solution was recommended by the AppStream online generator I used. I just added the `.gui` at the end to keep the opportunity open to devide between "common", "qt" and other components in the future.

qt/backintime-qt-root.desktop
```diff
@@ -12,3 +12,4 @@ Comment=Simple backup system
 Comment[de]=Ein simples Backup-Programm inspiriert von »Time Machine«, »TimeVault« und dem »flyback project«.
 Comment[sl]=Enostaven sistem ustvarjanja varnostnih kopij
 Keywords=automatic;snapshot;restore;rsync;root;
+X-AppStream-Ignore=true
```
buhtz: AppStream meta data is somehow connected to that desktop-files.
The meaning is that this "root" desktop file is ignored. For example the Debian validators won't look for extra appstream data for that desktop file.

io.github.bit_team.back_in_time.gui.xml
```diff
@@ -0,0 +1,88 @@
...
+        - Default desktop theme
+        - Include windows only (incl. boders) and use alpha mask for
+          rounded window corners
+        - Without mouse pointer
+        - Add caption (without fullstop)
+        -->
+    <screenshots>
+      <screenshot type="default">
+        <caption>Main window</caption>
+        <image>https://translate.codeberg.org/media/screenshots/bit_16to9_960x540_mainwindow.png</image>
```
buhtz: It is possible to point to screenshots in the local file system.
But I don't like the idea to "install" png files for each users. Waste of resources because of of our users do not use software centers and won't never see that pictures.

It is also possible to use a screenshot folder in our upstream repo or pointing to screenshots used in our documentation at readthedocs.org.

Not sure about a final and secure solution but currently I would keep it that way.

qt/configure
```diff
@@ -172,6 +172,7 @@ addComment "application"
 addInstallDir                           "/bin"
 addInstallFile "backintime-qt"         "/bin" "755"
 addInstallFile "backintime-qt_polkit"  "/bin" "755"
+printf "\tinstall --mode=644 io.github.bit_team.back_in_time.gui.xml /usr/share/metainfo\n"  >> ${MAKEFILE}
```
buhtz: @Fantu @graysky2 
Can you please have short look here. We now offer AppStream meta data and installing it into /usr/share/metainfo folder using the "install" make target for the "qt" component.

Seems to be OK to the Debian AppStream guidelines. But I would like to ask you for sure.

Fantu: I don't have enough time to check it. I have almost no knowledge about it anyway. I suppose that if you followed "Debian AppStream guidelines" (I suppose https://wiki.debian.org/AppStream/Guidelines) is ok.
Automatic check (https://appstream.debian.org/) is done for packages in repository, so will be checked on next version including it will be uploaded.

graysky2:
I have also never used this.

qt/configure
buhtz: ~Seems to be `/usr/share/app-info/xmls` on Arch but I have not validated that info yet.~ This info was wrong. Validated it on IRC with Arch users checking on their system.

There is also an [AppStream spec about the location](https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#spec-component-location). So we can rely on `/usr/share/metainfo`.

But on Arch (and I assume also on Debian) that folder is created only if a package with AppStream data is installed. I'll modify the make target to this.

qt/io.github.bit_team.back_in_time.gui.metainfo.xml
```diff
@@ -0,0 +1,88 @@
...
+      You only need to specify 3 things:
+    </p>
+    <ol>
+      <li>What folders to back up.</li>
+      <li>Where to save snapshots.</li>
+      <li>The backup frequency (manual, every hour, every day, every month).</li>
+    </ol>
+  </description>
+
+    <icon type="stock">document-save</icon>
```
aryoda, line 39: Will be interesting to see how our own icon can be used here...

qt/io.github.bit_team.back_in_time.gui.metainfo.xml
```diff
@@ -0,0 +1,88 @@
...
+  <metadata_license>FSFAP</metadata_license>
+  <project_license>GPL-2.0</project_license>
+
+  <description>
+    <p>
+      It is an easy-to-use backup tool for files and folders. It runs on GNU
+      Linux and provides a command line tool "backintime" and a Qt5 GUI
+      "backintime-qt". It uses "rsync" in the back to take manual or scheduled
+      snapshots and stores them locally or remotely through SSH. Each snapshot
+      is its own folder with copies of the original files, but unchanged files
```
aryoda: ...is **in** its own folder...?

qt/io.github.bit_team.back_in_time.gui.metainfo.xml
```diff
@@ -0,0 +1,88 @@
...
+    <url type="homepage">https://github.com/bit-team/backintime</url>
+    <url type="bugtracker">https://github.com/bit-team/backintime/issues</url>
+    <url type="faq">https://github.com/bit-team/backintime/blob/dev/FAQ.md</url>
+    <url type="help">https://backintime.readthedocs.io/</url>
+    <url type="translate">https://translate.codeberg.org/engage/backintime</url>
+    <url type="contact">https://mail.python.org/mailman3/lists/bit-dev.python.org</url>
+    <url type="vcs-browser">https://github.com/bit-team/backintime</url>
+    <url type="contribute">https://github.com/bit-team/backintime/blob/dev/CONTRIBUTING.md</url>
+
+    <launchable type="desktop-id">backintime-qt.desktop</launchable>
```
aryoda, line 62: We have two GUI launcher scripts (normal + root). Does the schema allow to specify both here?

qt/io.github.bit_team.back_in_time.gui.metainfo.xml
buhtz: When it is installed with the package and located in the right place then its name (e.g. `backintime-icon`) only would be enough. Otherwise using full path names is possible also.

qt/io.github.bit_team.back_in_time.gui.metainfo.xml
buhtz: I am investigating this. Not sure how to handle this. This will also influence translation management in the future because it decides if we have to translate the xml and both desktop files or the xml file only. The latter might be because the describing content of a desktop file seems to be ignored if an appstream file is available. I am testing this.

qt/io.github.bit_team.back_in_time.gui.metainfo.xml
buhtz: Both sounds OK to me. But my English shouldn't be the reference. 😆 
Modified it, also in REAMDE.md.

buhtz:
FYI I just summarizing open questions I am working on

1. I [asked at AppStream](https://github.com/ximion/appstream/discussions/594) about the consequences of having multiple `<launchable>` tags because the specs are not clear (to me) about that point.
2. I [asked at debian-python](https://lists.debian.org/debian-python/2024/02/msg00029.html) if upstream or a distro maintainer should take care of an appstream metadata file. I was not able to find an official guideline/policy about it. Facing the fact that some distro maintainers do not know what *.metainfo.xml is for I see no problem if upstream take care about the installation of that file.
3. I [asked at debian-desktop](https://lists.debian.org/debian-desktop/2024/02/msg00002.html) about an observed behavior that is contradicted to information I got from the Debian Appstream Wiki. Content from desktop-files (e.g. Name, description, comment, ...) should be overwritten by content from appstream metainfo files. But it was not in my tests.

buhtz:
I got no response about my questions. They are not so important or blocking to wait for it.
I vote to merge it.

> 1. I [asked at AppStream](https://github.com/ximion/appstream/discussions/594) about the consequences of having multiple `<launchable>` tags because the specs are not clear (to me) about that point.

I see no risk and can not imagine any negative consequences about having to `<launchable>` tags. Would keep them.

> 2. I [asked at debian-python](https://lists.debian.org/debian-python/2024/02/msg00029.html) if upstream or a distro maintainer should take care of an appstream metadata file. I was not able to find an official guideline/policy about it. Facing the fact that some distro maintainers do not know what *.metainfo.xml is for I see no problem if upstream take care about the installation of that file.

I would like to manage its installation by us at upstream. Some maintainers do not know about AppStream.

> 3. I [asked at debian-desktop](https://lists.debian.org/debian-desktop/2024/02/msg00002.html) about an observed behavior that is contradicted to information I got from the Debian Appstream Wiki. Content from desktop-files (e.g. Name, description, comment, ...) should be overwritten by content from appstream metainfo files. But it was not in my tests.

Not relevant yet. It become relevant in the future when I integrate the AppStream metainfo file and the desktop file into our translation pipeline. It seems that AppStream meta data do not overwritte the desktop-files content in all cases. I will treat them both with the same priority and see no problem with that. We just need to keep the description/comment strings in both files identically to lower the burden for translators.

aryoda (approved):
Let's try it "in real (dev) life" (I cannot estimate the impact by only reviewing the code)

willemw12:
I am getting an error, building backintime-git from the AUR:

    # Install application
    install -d /build/backintime-git/pkg/backintime-git/usr/bin
    install --mode=755 backintime-qt /build/backintime-git/pkg/backintime-git/usr/bin
    install --mode=755 backintime-qt_polkit /build/backintime-git/pkg/backintime-git/usr/bin
    install -d /usr/share/metainfo
    install: cannot change permissions of ‘/usr/share/metainfo’: No such file or directory
    make: *** [Makefile:41: install] Error 1
    make: Leaving directory '/build/backintime-git/src/backintime/qt'
    ==> ERROR: A failure occurred in package_backintime-git().

The install path should be relative to the package build output folder:

    install -d /build/backintime-git/pkg/backintime-git/usr/share/metainfo

buhtz:
Dear willemw12,
thanks for reporting back.

Can you please open a new issue for this. 
Please also post your full "Makefile".

The command `install -d /usr/share/metainfo` should create that folder. But the next error indicates this wasn't the case. Maybe you haven't had the rights creating that folder?

**EDIT**

See this line 179 in `qt/configure`:

    printf "\tinstall -d /usr/share/metainfo\n" >> ${MAKEFILE}

Can you please modify it to:

    addInstallDir "/share/metainfo"

Might this help?

willemw12:
Yes, with the usual "addInstallDir" it does build.

Issue #1643, opened 2024-02-14
No execution log


ceperman:
I think an overall execution log is needed, one that logs when BackInTime runs and what actions it takes or doesn't take. If there is one currently I've not found it.

For example, I look at my snapshots and don't see all that I expect - there is only one for this month and I expect to see more. I don't know if this is because (1) the schedule didn't call for them to be taken, (2) smart remove or some other action has removed them, (3) BackInTime didn't execute at all. or some other reason. At the moment I'm in the dark about what's happened.

buhtz:
Dear ceperman,
thanks for your report.

We have [a section about logging and lock files](https://github.com/bit-team/backintime/blob/dev/common/doc-dev/4_Control_files_usage_(locks_flocks_logs_and_others).md) in our developers documention. It is an ongoing investigation because the current maintenance are not the founders of BIT. But you won't find much in there about log files.

So I started to create two FAQ entries about log files. Please see my PR #1644 and feel free to review it and suggest improvements. There are the two new FAQ entries: [Where is the log file?](https://github.com/buhtz/backintime/blob/doc/1643faq_about_logs/FAQ.md#where-is-the-log-file) and [How to read log entries?](https://github.com/buhtz/backintime/blob/doc/1643faq_about_logs/FAQ.md#how-to-read-log-entries).

Would this information help you to answer your questions?

We are aware that there is room for improvement. Currently we do not have an active discussion or decisions about further development.

Back to your question I assume you refer to something that could be named as an "application log"? BIT do use syslog for its application log. This means you have to check your syslog or systemd-journal. See the new FAQ entries I linked to about how to read syslog entries.

Issue #1449, opened 2023-06-16
Redesign Schedule section in the GUI


buhtz:
Blocked by:
- #1850
- #1984
- And maybe some more refactoring stuff that need to be done beforehand.

Related to:
- #881 
- #1479

This is based on PR #1196 where it comes up that it might be a good idea to re-design the part of the GUI responsible for scheduling.

# Current situation
Currently it is just a drop down menu in the "General" tab of the "Settings" dialog of the profile settings.
![image](https://github.com/bit-team/backintime/assets/11861496/489ad719-880a-4d33-9129-5d9c9c836e4c)

## Problems
 - Now way to add "time restrictions" like hours where the job never should run. This is was PR #1196 originally was about.
 - Cron and Anacron like schedule options are mixed. To complex and hard to understand for some users now aware of the difference between cron and anacron. Some user might not understand that some jobs are missing when using "Every 6 hours" (cronlike).
 - Anacron is named in the dialog. But for real BIT do not use anacron but just simulate the anacron like behavior.
 - An "export" option is missing to create a user defined crontab line.

# Proposal
This is a quick and dirty (not finished) proposal in form of a GUI MockUp (used draw.io for that). I attached the original file if you like to edit.
 
 - I tried to reflect nearly the same behavior as in the current Schedule GUI. Exceptions are noted below.
 - Removed scheduling from "General" tab and created an extra tab "Schedule" for it.
 - Removed all crontab-based options (e.g. "Every 6 hours") because of the assumption that most users do misunderstand them and don't know that it means specific time points (e.g. 0:00, 6:00, 12:00, ...) and they would miss them if the machine isn't running.
 - Technical details (cron, anacron, etc pp) could be added as tooltips.
 - The tab now has 5 sections where only one of them can be active.
    1. "Run peridically after" is related to the anacron-like behavior. It means timespans.
    2. "On reboot" is cron and using `@reboot`. The "delay" is new and would be realized with a `sleep n && backintime` like cron command.
    3. "Every day/week/month/year" is related to cron and results in usual crontab lines. Of the 4 variants only one is visible depending on the value selected in the dropdown menu.
    4. "When drive get connected" use udev.
    5. "Expert/cronline" is new and opens the opportunity to advanced users creating their own crontab line. Maybe this could be somehow improved with 5 separated line edits etc.
  - The solution should be separated into multiple PR's of course.
  - See Qt classes `QDateTimeEdit`, `QTimeEdit`

[...]

PR #1644, opened 2024-02-14
doc: FAQ entries about log file

buhtz:
Add two FAQ entries about the location of log files and how to read them.

Related to #1643

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -5,6 +5,8 @@
 - [General](#general)
    * [Does _Back in Time_ support full system backups?](#does-back-in-time-support-full-system-backups)
    * [Does _Back in Time_ support backups on cloud storage like OneDrive or Google Drive?](#does-back-in-time-support-backups-on-cloud-storage-like-onedrive-or-google-drive)
+   * [Where is the log file?](#where-is-the-log-file)
+   * [How to read log entries?](#how-to-read-log-entries)
 - [Backups (snapshots)](#backups-snapshots)
    * [Does _Back In Time_ create incremental or full backups?](#does-back-in-time-create-incremental-or-full-backups)
    * [How do snapshots with hard-links work?](#how-do-snapshots-with-hard-links-work)
--- a/FAQ.md
+++ b/FAQ.md
@@ -97,7 +97,7 @@ There are three distinct logs generated:
    in the folder `~/.local/share/backintime/` and is named `restore_.log` for
    the main profile, `restore_2.log` for the second, and so forth.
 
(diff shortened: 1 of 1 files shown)


FAQ.md
```diff
@@ -82,9 +84,41 @@ which does not support `rsync`.
...
+   given time. It is stored within each snapshot and can be accessed through
+   the GUI.
+
+2. The _restore log_ contains messages specific to a particular restore
+   process. It is displayed in the GUI after each restore. It is also located
+   in the folder `~/.local/share/backintime/` and is named `restore_.log` for
+   the main profile, `restore_2.log` for the second, and so forth.
+
+3. The applications log is generated using the syslog feature of the operating
+   system. See [[How to read log entries?]] for further details.
```
aryoda: This syntax does not work (is not rendered). `[How to read log entries?](#how-to-read-log-entries)` should work but duplicates the headline (requires a careful global search&replace in case of changes.

Not sure if there is a better way for internal links

aryoda (approved):

buhtz:
Thanks for review. I'll wait for feedback from @ceperman .

ceperman:
Yes, that's great thanks. It looks like syslog contains the info I was asking for, and the faq describes how to find it - for me: 
`journalctl --grep backintime`

It also shows, by absence of logging, when BIT _didn't_ run when one might have expected it to, which is also very helpful. I thought BIT used anacron (old info in internet says it does) but that's obviously changed and the current docs makes it clear that it doesn't, so if the computer is switched off at the scheduled time, the backup doesn't get taken. My problem. RTFM!

buhtz:
Thanks for your feedback. Please let us know if you see more aspects for improvement in the docs. Don't hasitate to open multiple issues for this or sending an email to the mailing list.

> I thought BIT used anacron (old info in internet says it does) but that's obviously changed

Myself I am not sure if it was that way in that past. But currently BIT try to Imitate an anacron-like behavior. Maybe that is the reason why some people think anacron is involved. How the scheduling works is described [here](https://backintime.readthedocs.io/en/latest/settings.html#schedule) in the (quite out-dated) docs. The scheduling option "Repeatedly (anacron)" is not anacron but just a simple cronjob starting BIT every 15 minutes.

It is on my list to improve that scheduling section to make it more clear to users what is going on. See #1449

ceperman:
Thanks. I had not noticed that "Repeatedly (anacron)" was a scheduling option - I'd assumed it was a checkbox option that's missing from my oldish Linux Mint distribution version. Again RTFM!

PR #1632, opened 2024-02-02
Dependency: Migration to PyQt6

buhtz:
And here it comes... The Qt6 version of BIT.

No hurry with migration. Maybe doing some further manual tests.
The majority of modifications are just renaming some classes and enums because Qt moved some inside there library from one namespace to another. I am not aware of relevant modifications in the behavior or functionality between Qt5 and Qt6.

I did tests myself and found some minor problems and typos I was able to fix.

Close #1301
Close #1607

PR #1646, opened 2024-02-19
build: Move manpage-creation-script to root folder

buhtz:
The helper script creating "backintime-config.1" man page just moves one folder up.

The reason is that the script is contained in release tarballs and deb-files of BIT. But it is not necessary.

The CI is failing because of the [known PyLint problem](https://github.com/bit-team/backintime/pull/1625#issuecomment-1952208460) fixed in upcoming PR #1632. The "ppc64le" machine is not failing because Qt-tests are not executed on that architecture.

I "improved" the CI script to be more "verbose" about the special situation with Qt tests on ppc64le architecture.
image

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -78,7 +78,7 @@ script:
   - ./configure --python=python3
   - make
   # PyQt5 is not installed on "ppc64le"
-  - if [ "$TRAVIS_ARCH" != "ppc64le" ] ; then pytest ; else echo "On ppc64le no Qt testing because PyQt is impossible to install." ; fi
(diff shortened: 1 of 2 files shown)


aryoda (approved):

aryoda:
> The helper script creating "backintime-config.1" man page just moves one folder up.
> 
> The reason is that the script is contained in release tarballs and deb-files of BIT. But it is not necessary.

BTW: I guess the same holds true for `configure` and `Makefile` but this is quite difficult to solve

buhtz:
> BTW: I guess the same holds true for `configure` and `Makefile` but this is quite difficult to solve

I can confirm that Debian do not contain the make files. I am not sure about AUR.

PR #1650, opened 2024-02-22
fix: Open symlinked folders in fileview

### LINKED ISSUE #1476: FileView can not open symlinked folders
This issue needs further investigation. I won't touch it yet because I'm [in the middle of refactoring](https://github.com/buhtz/backintime/tree/feat/langmenu) `qt/app.py` while I uncovered this issue.

Reproduce:
 - Select "Now" in the timeline left to show the current state ot the filesystem and not any of the snapshots.
 - Double click on a folder which is a symlink.
 - Nothing will happen. The folder does not open.

The relevant code in the GUI is here:

https://github.com/bit-team/backintime/blob/1003d360d758bce78c6cd528ea41df3734fec95b/qt/app.py#L1338-L1353

In there the `self.sid.canOpenPath()` does return `False`.

It is not a serious bug. But I have the feeling that touching this will cause some more trouble and much more need to refactor. There might be a big workload hidden behind it. :smile:

buhtz:
Symlink folders in fileview now can be opened via double click in all circumstances.

I have to say that I am not sure about that solution or its implications. The related code is to complex (for me) to be sure.
But on my tests it worked.

Close #1476


buhtz:
I don't feel well merging this without proper review. So I am waiting. But there is no hurry.

aryoda:
> I don't feel well merging this without proper review. So I am waiting. But there is no hurry.

Sorry, I did review this (but without actually running the code) and the only thing I have found is the question about a better name for the `canOpenPath` function (something like eg. `IsExistingPathInsideSnapshotFolder` - or shorter).

Just propose a better name (or even decide to leave it as it is for now since there is a function doc string):

https://github.com/bit-team/backintime/blob/ed282ad53a3bbbfbab079ee71ea8b13fb17703c1/common/snapshots.py#L2623-L2632

Is there something special I should test (eg. risky code change)?

emtiu:
I won't have a chance for some time, but I'll be happy to test this in a VM.

buhtz:
> the only thing I have found is the question about a better name for the `canOpenPath` function (something like eg. `IsExistingPathInsideSnapshotFolder` - or shorter).

I renamed now and just used your proposal.

> Is there something special I should test (eg. risky code change)?

😁 It is primarily an irrational sensation. I lack a thorough understanding of that section of the code, which hinders me from realistically contemplating its implications. It's nothing serious, just a minor challenge.

No meed to merge it soon. It can wait some weeks if Michael want to test it.

PR #1651, opened 2024-02-22
refactor: Separate About dialog code & improved path handling in "config.py"

buhtz:
Separate the about dialog code into its own file and doing some refactoring and linter-fixes on it.

Additionally I discovered an opportunity to improve code in config.py that was related to the about dialog. I (re)moved some path handling code that conceptually do not belong to a config class. Of course I grep-ed the whole code base to be sure the (re)moved code is not used anywhere else.

Following linters are satisfied (without excluding any rules) on "qt/aboutdlg.py":
- ruff
- pycodestyle
- pylint
- pydocstyle
- codespell

PS: This was the last Qt6-related PR I was holding in my back. 🤣


common/tools.py
```diff
@@ -110,6 +111,38 @@ def sharePath():
...
+
+                            /usr/share/backintime/common
+                            /usr/share/backintime/qt
+    """
+    return os.path.abspath(os.path.join(__file__, os.pardir, os.pardir, *path))
+
+def docPath():
+    """Not sure what this path is about.
+    """
+    path = pathlib.Path(sharePath()) / 'doc' / 'backintime-common'
```
aryoda, line 133: `backintime-common` looks strange, is this correct?

aryoda: Looks correct, for some reason our `common/Makefile` does install the `LICENSE` file into:

`install --mode=644 ../LICENSE $(DEST)/share/doc/backintime-common`

common/tools.py
```diff
@@ -110,6 +111,38 @@ def sharePath():
...
+
+    # Dev note (buhtz, 2024-02-09):
+    # Don't know why this "if" exists. I assume this is an undocumented
+    # workraound to handle different approaches of distro maintainers.
+    # I am aware that on Debian this file usually do not exist at this
+    # location. But package maintainer made an exception in case of BIT
+    # so it do exist also on Debian.
+    license_file = pathlib.Path(backintimePath()) / 'LICENSE'
+    if license_file.exists():
+        path = backintimePath()
```
aryoda, line 149: I think this code is based (= the refactoring result) on

https://github.com/bit-team/backintime/blob/2ad871f1d2d92fbcbdbd5d7128b04a8f286c17c6/common/config.py#L148-L151

`git blame` says the `if` is from 2008 (by Dan)

Our generated `Makefile` (in `common`) does only install into `share/doc/backintime-common` but never into the the `backintime` "binary" path so I guess the `if` is
a) either a distro-specific exception for a distro package that (manually?) installs the `LICENSE` into another path
b) or a left-over from old code where the `LICENSE` was installed differently...

Perhaps an additional TODO comment would help to check all known distro packages about the installation target of the `LICENSE` to decide then if the code can be removed (no need to investigate this now IMHO)

aryoda (approved):
I am very happy that you have started to split the old "monster" `app.py` into "one file per dialog" :smile:

common/tools.py
buhtz: Strange to me, too. But I didn't want to change the logic. There might be a good reason for this code. I just moved it. It was in the `Config.__init__()` before:

    self._DOC_PATH = os.path.join(tools.sharePath(), 'doc', 'backintime-common')

buhtz: "backintime-common" is the distros package name.

PR #1655, opened 2024-02-23
fix: PyLint discover "name-error" do not hide "import-error" / Repair AboutDlg code / Minor fixes

### LINKED ISSUE #1653: Bug(s). "is_Qt6_working() not in tools" and uncaught PyLint errors "no-member" because of disabled PyLint error "import-error"
https://github.com/bit-team/backintime/blob/ff294f723f02fc2fef41afda6c5fdacc24747ba5/qt/plugins/systrayiconplugin.py#L75

Accidentally discovered a name error in my debug output while working on something else. Easy to fix because the function now is named `is_Qt_working()`.

But this brought me to the question why PyLint did not warn about it. It should because "no-member" (E1101) is activated in our PyLInt config. The answer is that PyLint is not able to import `tools` module and because of that is not able to tell if that member exists or not. But the "import-error" (E0401) is deactivated in our PyLint config.

I am working on it.

buhtz:
This PR began so nice but is a mess in the end. Sorry.

There was a hidden "no-member" error: `tools.is_Qt6_working()` was renamed to `tools.is_Qt_working()`. I fixed it.

That error was hidden because as described in #1653 PyLint wasn't that accurate as I thought. So I modified its configuration. "no-member" errors related to our own BIT code weren't detected because PyLint was not able to import our own modules. This should trigger an "import-error" warning but this was not activated in our config. I fixed it and also made sure PyLint can import our own code.

And "TaDa"... I found some other interesting problems via running PyLint. See my comments in the diff output.

What really bothers me is that the code in `qt/aboutdlg.py` looked older than I remembered it. There where exceptions. I am not sure but I assume I reverted some of my improvements via solving merge conflicts in the wrong direction.

Additionally I allowed myself to rename `common/qt5_probing.py` into `common/qt_probing.py`.

Close #1653

### DIFF
--- a/common/askpass.py
+++ b/common/askpass.py
@@ -16,15 +16,11 @@
 
 import os
 import sys
-try:
-    import gtk
-except:
-    pass
-
 import password
 import password_ipc
 import tools
 import config
+
 
 if __name__ == '__main__':
     """
@@ -52,5 +48,6 @@ if __name__ == '__main__':
     #temp mode
     fifo = password_ipc.FIFO(temp_file)
     pw = fifo.read(5)
+
     if pw:
         print(pw)
--- a/common/doc-dev/conf.py
+++ b/common/doc-dev/conf.py
@@ -61,7 +61,8 @@ copyright = '2016, Germar Reitze'
 author = 'Germar Reitze'
 
 # Don't edit this variable. It is updated automatically by "updateversion.sh".
-version = '1.4.4-dev.bb32455a'
+import backintime
+version = backintime.__version__
 # The full version, including alpha/beta/rc tags.
 release = version  # '1.3.3-dev'
 
--- a/common/doc-dev/conf.py
+++ b/common/doc-dev/conf.py
@@ -61,7 +61,7 @@ copyright = '2016, Germar Reitze'
 author = 'Germar Reitze'
 
 # Don't edit this variable. It is updated automatically by "updateversion.sh".
-version = '1.4.4-dev'
+version = '1.4.4-dev.117a3021'
 # The full version, including alpha/beta/rc tags.
 release = version  # '1.3.3-dev'
 
--- a/common/doc-dev/conf.py
+++ b/common/doc-dev/conf.py
@@ -61,7 +61,7 @@ copyright = '2016, Germar Reitze'
 author = 'Germar Reitze'
 
 # Don't edit this variable. It is updated automatically by "updateversion.sh".
-version = '1.4.4-dev.117a3021'
+version = '1.4.4-dev.bb32455a'
 # The full version, including alpha/beta/rc tags.
 release = version  # '1.3.3-dev'
 
--- a/common/test/test_lint.py
+++ b/common/test/test_lint.py
@@ -51,6 +51,8 @@ class MirrorMirrorOnTheWall(unittest.TestCase):
         # Pylint base command
         cmd = [
             'pylint',
+            # Make sure BIT modules can be improted (to detect "no-member")
+            '--init-hook=import sys;sys.path.insert(0, "./../qt");',
             # Storing results in a pickle file is unnecessary
             '--persistent=n',
             # autodetec number of parallel jobs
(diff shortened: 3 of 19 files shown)


common/askpass.py
```diff
@@ -16,16 +16,12 @@
 
 import os
 import sys
-try:
```
buhtz, line 19: I grep-ed it and can't find a reasons to keep it.

common/test/test_lint.py
```diff
@@ -51,6 +51,8 @@ def test_with_pylint(self):
         # Pylint base command
         cmd = [
             'pylint',
+            # Make sure BIT modules can be improted (to detect "no-member")
+            '--init-hook=import sys;sys.path.insert(0, "./../qt");',
```
buhtz, line 55: This hook make pylint able to import our BIT modules and decide about "naming-errors".

common/test/test_lint.py
```diff
@@ -74,12 +76,11 @@ def test_with_pylint(self):
             'E0602',  # undefined-variable
             'E1101',  # no-member
             'W1401',  # anomalous-backslash-in-string (invalid escape sequence)
+            'E0401',  # import-error
```
buhtz, line 79: This allow us to know if PyLint has problems to import something.

common/tools.py
```diff
@@ -389,7 +389,8 @@ def get_git_repository_info(path=None, hash_length=None):
     """
 
     if not path:
-        path = pathlib.Path.cwd()
+        # This should be the repositories root folder
```
buhtz: This function did not worked as expected in my manual tests. Not sure if it was the case in the past or if this is because of my wrong-solved merge conflicts.

qt/aboutdlg.py
```diff
@@ -80,9 +81,13 @@ def __init__(self, parent=None):
         vlayout.addLayout(hlayout)
 
     def _create_name_and_version_label(self):
```
buhtz, line 83: Here I fixed that old code. I am sure I have done this in the past. But the old stuff came back somehow.

qt/plugins/systrayiconplugin.py
```diff
@@ -72,7 +72,7 @@ def init(self, snapshots):
 
         try:
 
-            if tools.is_Qt6_working(systray_required=True):
+            if tools.is_Qt_working(systray_required=True):
```
buhtz, line 75: This is where the story begun. 😆 
That name error appeared in DEBUG output. Then I started to investigate why PyLint can not see that error, too.

buhtz:
TravisCI fails because of import errors related to `dbus`. Working on it...

I hate that "ppc64le" architecture. 👿 Next time I will ignore every error/problem related to that architecture and just disable the related checks. I really see no value in keeping this architecture on TravisCI because Travis use a virtual environment and the the regular Ubuntu for ppc64le. We better should catch a ppc64le user and ask him for running the test suite on his own (real!) machine on our request.

buhtz:
Done. 🎈

common/askpass.py
aryoda: confirmed (`gtk` package is not used)

common/doc-dev/conf.py
```diff
@@ -61,7 +61,7 @@
 author = 'Germar Reitze'
 
 # Don't edit this variable. It is updated automatically by "updateversion.sh".
-version = '1.4.4-dev'
+version = '1.4.4-dev.bb32455a'
```
aryoda: I am not sure if we should merge "hashed" version numbers or not (may not harm but also be misleading if we don't `make`/`updateversion` before each push.

OTOH it does not hurt I think

aryoda (approved):

aryoda:
> We better should catch a ppc64le user and ask him for running the test suite on his own (real!) machine on our request.

I can provide access to real `ppc64le` hardware (server in Toulouse) if really required but I think we should use this only in case of user-reported issues (I do not expect severe findings if tests on other architectures do succeed on Travis CI)

common/doc-dev/conf.py
buhtz: Ah, yes. I did a `make install` and forget that.
Mhm... I could add a unit test that run on Travis only checking that.

buhtz:
> I can provide access to real `ppc64le` hardware

I think first I have to cool down about that topic. :rofl: :snowman_with_snow: But then we can discuss if we need automated tests on that architecture.

PR #1656, opened 2024-02-23
fix(build): Use addInstallDir to create metainfo folder

### LINKED ISSUE #1654: Build error AUR package backintime-git "install: cannot change permissions of ‘/usr/share/metainfo’: No such file or directory"
https://github.com/bit-team/backintime/pull/1642 causes a build error of the Arch AUR backintime-git package:

     "install: cannot change permissions of ‘/usr/share/metainfo’: No such file or directory"

The AUR package build file can be found here: https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=backintime-git. Temporarily comment out the check() function if some tests fail,

The problem and the solution are already mentioned here https://github.com/bit-team/backintime/pull/1642#issuecomment-1961257140:

    in qt/configure, change:

    printf "\tinstall -d /usr/share/metainfo\n" >> ${MAKEFILE}

    to:

    addInstallDir "/share/metainfo"

buhtz:
Fix #1654 

@willemw12 Can you test this on your AUR before merging?

### DIFF
--- a/qt/configure
+++ b/qt/configure
@@ -176,9 +176,8 @@ addComment "application"
 addInstallDir                           "/bin"
 addInstallFile "backintime-qt"         "/bin" "755"
 addInstallFile "backintime-qt_polkit"  "/bin" "755"
-printf "\tinstall -d /usr/share/metainfo\n" >> ${MAKEFILE}
+addInstallDir "/share/metainfo"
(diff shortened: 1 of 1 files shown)


willemw12:
Builds and installs successfully. Tested with the new PyQt6 dependency.

aryoda (approved):

PR #1658, opened 2024-02-24
test: User-callback

### LINKED ISSUE #1648: Unit test: User callback
This issue need to be solved before working on PR #1269 .

To my knowledge user-callback behavior is not covered by any of the current tests. Of course the code is executed in some tests (and covered in a technical way) but not tested.

IMHO the **behavior to test** would be if the callback reasons/steps are called on the right position on the timeline after/before some other specific steps of the taking a snapshot process.

*Problem*: My assumption is that this is nearly impossible in the current state of the productive code. The code is not isolated enough. And especially in testing user-callback behavior to many elements of code are involved. I see no way doing this with "unit tests" or "integration tests" in the near time.

*Solution*: A "dirty" approach could be to do a full snapshot run in an "system test" and then parsing the log output for the expected messages (or their patterns) and if they appear in the correct order. We still have tests doing full backups. They do use some helper code from `test/generic.py`.

buhtz:
Try to close #1648 creating a system test checking the output produced by a user callback while running a backup.

This PR might not be ready to merge but is read to review. 

Do run the test via:

    $ cd common
    $ pytest test/test_plugin_usercallback.py::SystemTest -vs


aryoda:
> A backup was created one time. But each next run of that test gave me a `Nothing changed, no new snapshot necessary` messages. **But** that test do create a temporary folder and delete it after every run. I am pretty sure about that. So how can there be a no snapshot needed message?

IMHO the `takeSnapshot()` function is quite complicated for historical reasons (a lot of branches and the core logic "hidden" in an `else` block). Above observations can be explained by this:

The `INFO: Take a new snapshot. Profile: 1 Main profile` log output is created first:

https://github.com/bit-team/backintime/blob/5cbffdf5d858231c370149f4cf9b967e0de66011/common/snapshots.py#L759-L765

Then the `takeSnapshot()` function is called which really tries to create a snapshot:

https://github.com/bit-team/backintime/blob/5cbffdf5d858231c370149f4cf9b967e0de66011/common/snapshots.py#L796

If no changes were reported by `rsync` and the config (profile) **does not enforce** taking a snapshot even if nothing has changed then you see the `INFO: Nothing changed, no new snapshot necessary` log output:

> The 2nd element of `params` is never set to `True`:
https://github.com/bit-team/backintime/blob/5cbffdf5d858231c370149f4cf9b967e0de66011/common/snapshots.py#L1284-L1286

> But that test does create a temporary folder and delete it after every run

Yes, this is the only way for BiT's `takeSnapshot` to check if anything has changed:
1. Create a temp folder as snapshot target (well hidden in an `if`):
    https://github.com/bit-team/backintime/blob/5cbffdf5d858231c370149f4cf9b967e0de66011/common/snapshots.py#L1185-L1186
2. Execute `rsync`
    https://github.com/bit-team/backintime/blob/5cbffdf5d858231c370149f4cf9b967e0de66011/common/snapshots.py#L1225-L1229
    The `user_data` argument contains the well-known `params` variable which stores the `rsync` result finally (error flag + changes flag).

3. If an error occurred during `rsync`ing or nothing has changed then the temp snapshot target folder is deleted again (`self.remove(new_snapshot)`):

    https://github.com/bit-team/backintime/blob/5cbffdf5d858231c370149f4cf9b967e0de66011/common/snapshots.py#L1274-L1286

buhtz:
I updated my initial comment. Read to review.

common/config.py
```diff
@@ -174,13 +181,16 @@ def __init__(self, config_path=None, data_path=None):
         else:
             self._LOCAL_CONFIG_PATH = os.path.abspath(config_path)
             self._LOCAL_CONFIG_FOLDER = os.path.dirname(self._LOCAL_CONFIG_PATH)
-        old_path = os.path.join(self._LOCAL_CONFIG_FOLDER, 'config2')
```
buhtz, line 177: Didn't I removed this in another PR?
Might be related to my messed up AboutDialog-PR where I assume to "solved" a merge conflict into the wrong direction.

Really sorry about that.

aryoda: Git is (a) be(a)st ;-)

buhtz:
Regarding to [your comment in PR #1269](https://github.com/bit-team/backintime/pull/1269#issuecomment-1266834094) the question is if this tests would make you sleep better? 😄 
Currently it is only a "local snapshot" without ssh involved. But the (un)mount reason signals (7 & 8) are fired also.

common/test/test_lint.py
```diff
@@ -77,6 +77,7 @@ def test_with_pylint(self):
             'E1101',  # no-member
             'W1401',  # anomalous-backslash-in-string (invalid escape sequence)
             'E0401',  # import-error
+            'I0021',  # useless-suppression
```
buhtz, line 80: See [pylint docu about `I0021`](https://pylint.readthedocs.io/en/latest/user_guide/messages/information/useless-suppression.html). It detects useless pylint-disable comments (e.g. `# pylint: disable=E0401`). I assume we will use such comments a lot in the next years. But they also become useless again very often because of heavy refactoring.

PR #1629, opened 2024-02-02
Fix: Removing exclusion patterns in 'Highly recommended' widget which are …

Prakul95:
…already excluded in manage profile

created a set for excluded list items and created a new list of recommended list which are not present in excluded

Would fix  #1620

PR #1665, opened 2024-03-12
Fix: "Highly recommended" exclusion patterns show missing patterns only

### LINKED ISSUE #1620: "Highly recommended" exclusion patterns in "Manage profiles" GUI dialog should show only missing patterns
## Current status

`rsync` and therefor BiT do allow to specify "exclude file patterns" for files that shall not be `rsync`ed:

Currently the GUI always shows all recommended (= default) exclusion file patterns:

image

## Feature request

It would be more helpful to show only those default exclusions patterns as "highly recommended" that are not already contained in the actual exclusion list.

## Relevant code

https://github.com/bit-team/backintime/blob/d583cf8a4b7de4271f897f07a94d0fbe07f9caff/qt/settingsdialog.py#L556-L561

https://github.com/bit-team/backintime/blob/d583cf8a4b7de4271f897f07a94d0fbe07f9caff/common/config.py#L98-L119

## Related to

- #1583
- #1618
- #1594
- #1382
- #1215
- #972

buhtz:
- I joined the two used labels and now use Qt's Rich Text feature to realize the bold font.
- I removed a second invisible column from the list widget. I was not able to find out why it is used. Maybe it was old code from a previous "solution"?

The result looks like this:
image

Or when all recommendations full filled:
image

Fix #1620
Close incomplete PR #1629

### DIFF
--- a/qt/settingsdialog.py
+++ b/qt/settingsdialog.py
@@ -570,6 +570,8 @@ class SettingsDialog(QDialog):
         label.setWordWrap(True)
         layout.addWidget(label)
 
+        print(f'\n{self.config.DEFAULT_EXCLUDE=}')  # DEBUG
+
         buttonsLayout = QHBoxLayout()
         layout.addLayout(buttonsLayout)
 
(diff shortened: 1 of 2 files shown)


qt/settingsdialog.py
```diff
@@ -561,14 +561,10 @@ def __init__(self, parent):
             .connect(self.excludeCustomSortOrder)
 
         layout.addWidget(self.listExclude)
-        self.listExcludeCount = 0
```
buhtz, line 564: This variable is related to the "second invisible column". Couldn't find a code location where it is used.

qt/settingsdialog.py
```diff
@@ -1788,18 +1805,14 @@ def addInclude(self, data):
 
         return item
 
-    def addExclude(self, pattern):
+    def _add_exclude_pattern(self, pattern):
         item = QTreeWidgetItem()
         item.setText(0, pattern)
         item.setData(0, Qt.ItemDataRole.UserRole, pattern)
-        self.listExcludeCount += 1
-        item.setText(1, str(self.listExcludeCount).zfill(6))
```
buhtz, line 1796: The second column (index 1).

aryoda (approved):

PR #1678, opened 2024-04-06
test/feat: Include F2FS root fs in test script

marfrit:
I am using F2FS on my Pinebook Pro as the root fs; backintime-cli will not build properly under these circumstances, since the test_tools.py script does not expect a F2FS root. This commit fixes this.

### DIFF
--- a/common/test/test_tools.py
+++ b/common/test/test_tools.py
@@ -432,7 +432,7 @@ class TestTools(generic.TestCase):
         self.assertRegex(tools.filesystem('/sys'), r'sys.*')
         self.assertRegex(
             tools.filesystem('/nonExistingFolder/foo/bar').lower(),
-            r'(:?ext[2-4]|xfs|zfs|jfs|raiserfs|btrfs|tmpfs|overlay|shiftfs)')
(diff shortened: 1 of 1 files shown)


buhtz:
Hello Markus,
Thanks for your contribution.
I don't have any experience with F2FS. I did some Wikpedia research about it. I can not find a answer to the question if it support hardlinks. Do you have any information or reference about it?

Beside this failing tests. Have you used Back In Time on your Pinebook? Does it work? Do you see any indications for problems?

Can you send the output of `backintime --diagnostics` please?

Kind
Christian

buhtz:
I propose to remove the whole test. See my PR #1687 . Will wait for feedback of for my team.

marfrit:
Hi, 

documentation I know - https://www.kernel.org/doc/Documentation/filesystems/f2fs.txt - it is designed to be a drop in replacement for ext4 for, e.g., eMMC devices. It supports hardlinks and I haven't had any issues (appart from backintime not installing properly ;)).

buhtz:
Thanks for the feedback.

And please take my apologize that I maybe close this PR without merging. Nevertheless, it's valuable PR as you can see in PR #1687 and three follow-up issues. Your PR triggered some other issues and improved code quality because of that.

PR #1681, opened 2024-04-08
feat: Debug output for scheduled jobs (--debug in crontab)

### LINKED ISSUE #1616: Add --debug checkbox to BiT GUI for cron jobs
Currently to create debug output in case of problems the user must manually edit the `crontab` file via

> Use sudo crontab -e and add --debug to the backintime job entries (move cursor after backintime and insert --debug - this is vi so press i to enter the edit mode, after editting press ESC then :wq to save the changes...

and this change will be overwritten again if the profile is edited in the GUI.

For better usability add a checkbox into the _General > Schedule_ block

   [x] Enable logging of debug messages

with a tooltip text

   `Write debug-level messages into the system log. Warning: Use only to diagnose problems since it creates a lot of output`

stcksmsh:
Added checkbox to allow debugging to `settings>general>schedule` in qt and the necessary code to common to make it work.

Fixes  #1616


qt/settingsdialog.py
```diff
@@ -482,6 +482,11 @@ def __init__(self, parent):
 
         self.comboSchedule.currentIndexChanged.connect(self.scheduleChanged)
 
+        self.cbDebugSchedule = QCheckBox(self)
+        self.cbDebugSchedule.setText(_('Enable logging of debug messages'))
+        self.cbDebugSchedule.setToolTip(('Write debug-level messages into the system log. Warning: Use only to diagnose problems since it creates a lot of output.'))
```
buhtz: Encapsulate the string within `_()` (underscore in the beginning). Then it gets into the translatation engine.

English is not my native language and I am not the best on it. Not sure but I suggest a modification of the string incl. correct line break before the 80th character:

    _('Writes debug-level messages to the system log. Warning: Only use this '
      'for error diagnosis, as it generates a large amount of output.')

buhtz:
Thanks a lot for this. Looks good.
In the team we have the rule to keep a PR of our own minimum 1 week open until we merge it without a review of someone else. But for your PR I would like to wait for aryoda's review because he had the original idea. So don't be worried if this PR is open for some weeks. He will reach out someday and he do not lose or ignore notification emails.

Please do not forget the changelog entry in the `CHANGES` file. Have a look at https://common-changelog.org/ if you want to learn more about writing changelogs. We don't have hard rules or guidelines in the BIT team about writing them. You are free to guess.

Your code looks good. I added only minor suggestions in the other comments.

When it comes to testing. There are two different "schools": London and Detroit (aka classic). None of them are right, good or wrong. It always depends. But for beginners they should try to follow the general rules offered by Detroit school. In that school a "unit" is not code or a code block it is "behavior".

Your PR do introduce a new behavior: The crontab line (generated by the Schedule settings) contain a `--debug` if the corresponding setting in the GUI has been done.

OK, let's cut the GUI here because it makes it to complex. You would create a unit test like this Pseudocode:

```
def MyTestSuite(unittest.TestCase):
    def test_crontab_contain_debug():
    
        # create a config instance with a minimal set of options
        conf = config.Config(simple_test_config)
        
        # Test the start conditions: No debug
        self.assertFalse(conf.debugSchedule())
        self.assertFalse(logger.DEBUG)
        
        # The "system under test" (sut)
        sut = conf.cronCmd(profile_id='1')
        self.assertIsNotIn('--debug', sut)
        
        # Invert the situation
        conf.setDebugSchedule(True)
        
        # Test start conditions like this might be pedantic. But my experience
        # tells me it is worth it especially in an old and complex code base
        # like BIT.
        self.assertTrue(conf.debugSchedule())
        
        sut = conf.cronCmd(profile_id='1')       
        self.assertIsIn('--debug', sut)

    def test_crontab_without_debug():
        conf = config.Config(simple_test_config)
        
        conf.setDebugSchedule(True)

        self.assertTrue(conf.debugSchedule())
        self.assertFalse(logger.DEBUG)

        sut = conf.cronCmd(profile_id='1')       
        self.assertIsNotIn('--debug', sut)

```

One problem with BIT is that the several code parts are not isolated enough from each other. This makes it often hard to nearly impossible to create a valuable unit tests without to much mocking/fakeing/tricking.

Have you ever heard of "test driven development" (TDD)? Your PR is a good example. The behavior I described is what you want. Before starting anything else just write the tests. While writing the test you becoming more clear about the details you need and you don't need. The test will fail in the beginning. When your tests is finished and you are clear about what your needs then you start to implement the code you need to make the test pass (green!).

common/config.py
```diff
@@ -958,6 +958,13 @@ def scheduleMode(self, profile_id = None):
     def setScheduleMode(self, value, profile_id = None):
         self.setProfileIntValue('schedule.mode', value, profile_id)
 
+    def debugSchedule(self, profile_id = None):
```
aryoda: I think `scheduleDebug` and `setScheduleDebug` would be more consistent to the other existing function names

aryoda:
@stcksmsh THX a lot for your contribution! I will review, build and test your code in detail next week I hope (I am running out of time ATM).

Meanwhile first observations:

- Would you mind adding the documentation for the new config property into our [man page](https://github.com/bit-team/backintime/blob/d33f35e10b4c3027b3b12881152c157eb1b2cc8c/common/man/C/backintime.1) too
- a [CHANGELOG](https://github.com/bit-team/backintime/blob/d33f35e10b4c3027b3b12881152c157eb1b2cc8c/CHANGES) is also required (as  @buhtz already mentioned above)

buhtz:
> * Would you mind adding the documentation for the new config property into our [man page](https://github.com/bit-team/backintime/blob/d33f35e10b4c3027b3b12881152c157eb1b2cc8c/common/man/C/backintime.1) too

That is not needed. The man page for the config file `backintime-config.1` is generated automatically based on the comments (starting with `#?`) used in the getter/setter methods in the config.py file. There is a [fancy script](https://github.com/bit-team/backintime/blob/dev/create-manpage-backintime-config.py) (by Germar I think) extracting the comments from the config.py and generate the man page for it.

**EDIT**: Ah I see you pointed to `backintime.1` instead of `backintime-config.1`. Maybe in the section about the `--debug` argument could be a hint about this new setting, yes.

aryoda:
> Maybe in the section about the `--debug` argument could be a hint about this new setting, yes.

This was a "late night" mistake of mine, I meant the `backintime-config.1` :smile:

common/test/test_backup.py
```diff
@@ -255,3 +255,39 @@ def test_takeSnapshot_exception_cleanup(self, takeSnapshot, sleep):
...
+    def test_crontab_contains_debug(self, sleep):
+        # create config object with default values
+        conf = config.Config('config')
+        logger.DEBUG = False
+
+        # assert start conditions: NO DEBUGGING
+        self.assertFalse(conf.debugSchedule())
+        self.assertFalse(logger.DEBUG)
+
+        # the "system under test" (sut)
```
buhtz: Line 268 - 271
I know it was part in my own pseudo code. But seeing it now in real I would suggest to remove this code block from this test. The reason is that this specific behavior is tested in the other test `test_crontab_without_debug()`.

buhtz:
Congrats! 🥳 🎈 New tests and they are green. Nice work. I added one comment to the code.

Beside of that I would suggest to move these two tests somewhere else because it IMHO to not really belong into that test suite.

I am not sure how to handle the `Config` instance in this case. Currently your code use a real config file inside the test folder.  And that instance is also used by all other tests. Never a good idea despite that nearly every other BIT test does it the same way. I would try to create an in-memory config file for this test.

I am not sure about a solution. Because `Config` is not able to read from a "file like" object (e.g. an in-memore `io.StringIO` instance) I would try to use a fake filesystem using PyFakeFS. Would it be OK for you if I would give a try and taking over from here?

stcksmsh:
Thanks for the feedback and for offering to handle the changes. Feel free to proceed with moving the tests and implementing the in-memory config file. Let me know if you need any assistance. Looking forward to the improvements!

buhtz:
My apologize again for blowing up this PR. Discovered a problem with the "tools" module while moving the new tests and then discovered something more.... 🤣 Shouldn't have driven it that far. **Kosta**: I am a bad example in this case. A PR shouldn't be that big and complex in the end.

Beside the whole PR I do requesting review by @bit-team for two specific aspects of this PR and won't merge it until a I get feedback about it:

1. In short: The debug mode of the GUI was inherited by the scheduled backup jobs. Previously if BIT GUI was started in debug mode (`--debug`) and the schedule settings are modified the result always was that the resulting crontab line contained a `--debug` argument (see `config.py` line 1759/1783). I removed that behavior because it feels not appropriated and find it also quit dangerous because the syslog is filled up without the knowledge of the users. With this PR we now have an explicit setting to achieve the desired behavior.
2. I out-commented 3 lines of code in the method `config.py::Config.cronCmd()` starting with line 1765. I see no value in it and can not imagine an appropriated use case. But just for sure would like to have your opinion about it.

Beside of this:
- Moved the new tests to their own test file and simplified their code.
- fix: some linter warnings/errors from "codespell" and "pylint".
- fix: PyLint test now fails on any error including pylint-usage errors.
- Based on experience in other projects and the PyLint statistic on BIT's code base (using `--reports=yes`) I used a code comment to introduce some new pylint rules. They are good candidates to be fixed in the near future without introducing to much code changes.

qt/settingsdialog.py
```diff
@@ -482,6 +482,14 @@ def __init__(self, parent):
 
         self.comboSchedule.currentIndexChanged.connect(self.scheduleChanged)
 
+        self.cbDebugSchedule = QCheckBox(self)
+        self.cbDebugSchedule.setText(_('Enable logging of debug messages'))
+        self.cbDebugSchedule.setToolTip(_(
+            'Writes debug-level messages into the system log. Caution: Only '
```
buhtz: Feel free to rephrase this tool tip.

buhtz (approved):
Approve on my site. But waiting for other team mates before merging.

aryoda:
FYI: I plan to build and review the PR end of this week (hopefully ;-)

CHANGES
```diff
@@ -1,13 +1,15 @@
 Back In Time
 
 Version 1.4.4-dev (development of upcoming release)
+* Breaking Change: Argument "--debug" in crontab not used anymore when GUI runs in debug mode.
```
aryoda: @buhtz Do you know what this CHANGE line means?
- What is "breaking"?
- How is the crontab `--debug` related to the BiT GUI `--debug` in the existing code (= not the code of this PR)?

qt/settingsdialog.py
aryoda: Done

common/config.py
aryoda: Done (just pushed a refactoring commit)

CHANGES
aryoda: I just found your comment and guess it is answering my question, right?

> In short: The debug mode of the GUI was inherited by the scheduled backup jobs. Previously if BIT GUI was started in debug mode (--debug) and the schedule settings are modified the result always was that the resulting crontab line contained a --debug argument (see config.py line 1759/1783). I removed that behavior because it feels not appropriated and find it also quit dangerous because the syslog is filled up without the knowledge of the users. With this PR we now have an explicit setting to achieve the desired behavior.

I would modify the CHANGES line to make it more clear what has changed, eg.:

Breaking change: GUI started with --debug does no longer add --debug to the crontab for scheduled profiles. Use the new "enable logging for debug messages" in the 'Schedule' section of 'Manage profiles' instead

buhtz:
- I wrapped the tool tip.
 - Separated it into two paragraphs to realize a line break before "Caution".
 - Replaced `!` with `.` because once a professional GUI translator told me one should not shout at or pressure users. That's roughly how I understood it. :smile: 

grafik

aryoda:
OK, this PR is now ready to be merged into `dev` (I'd suggest to wait a few "cooling-off" days before merging).

Thanks a lot to @stcksmsh for this great contribution and to @buhtz for the refinements - excellent team work :+1:

aryoda (approved):

CHANGES
aryoda: Done in https://github.com/bit-team/backintime/pull/1681/commits/333f09d354e88a04ed3f269f1cef36099931340e

buhtz (approved):

PR #1682, opened 2024-04-08
Fallback diff command

### LINKED ISSUE #1662: Compare snapshots: Don't use "false" as fallback diff command
https://github.com/bit-team/backintime/blob/ed282ad53a3bbbfbab079ee71ea8b13fb17703c1/qt/snapshotsdialog.py#L34-L42

Based on [report on bit-dev mailing list](https://mail.python.org/archives/list/bit-dev@python.org/message/NR4VSKGWOHB6NF4DINZZGI42I5SBNVXR/).

### Problem
An extern application is used to compare snapshots. If `meld` and `kompare` as such tools are not installed BIT is falling back to `false`. Be aware that `false` is not `False` and that `false` is a valid command in a *nix shell (see `$ manpage false`). This makes no sense from users perspective. The "Compare" button just does nothing in this case. No reaction. No response to the user.

BIT doesn't check for `false` but just checks for availability of the configured diff command.
https://github.com/bit-team/backintime/blob/ed282ad53a3bbbfbab079ee71ea8b13fb17703c1/qt/snapshotsdialog.py#L336-L340

image

### Solution
The diff command should be empty (or None) if meld or kompare are not available. And the Compare button should check for emptiness.
Also the error message could be improved. A simple "Command not found" is not user friendly and to technical.

stcksmsh:
Followed [buhtz's refacroting suggestions](https://github.com/bit-team/backintime/issues/1662#issuecomment-2041925731).

Changed initial `DIFF_CMD` and `DIFF_PARAMS` creation, changed `diffCmd` checking mechanism and added better error description

Fixes #1662

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,6 +1,7 @@
 Back In Time
 
 Version 1.4.4-dev (development of upcoming release)
+* Fix: Validation of diff command settings in compare snapshots dialog (#1662) (@stcksmsh Kosta Vukicevic)
 * Fix bug: Open symlinked folders in file view (#1476)
 * Fix bug: Respect dark mode using color roles (#1601)
 * Fix: "Highly recommended" exclusion pattern in "Manage Profile" dialog's "Exclude" tab show missings only (#1620)
--- a/qt/snapshotsdialog.py
+++ b/qt/snapshotsdialog.py
@@ -36,9 +36,9 @@ if tools.checkCommand('meld'):
 elif tools.checkCommand('kompare'):
     DIFF_CMD = 'kompare'
 else:
-    DIFF_CMD = ''
+    DIFF_CMD = None
 
-if DIFF_CMD == '':
+if DIFF_CMD is None:
     DIFF_PARAMS = ''
 else:
     DIFF_PARAMS = '%1 %2'
@@ -82,7 +82,7 @@ class DiffOptionsDialog(QDialog):
                 self, _('The command "{cmd}" is not valid.').format(cmd=diffCmd))
             return
 
-        # if it is empty, use the default
+        # if it is empty, use the default (here we use '' instead of None since if the user leaves the field empty, the value will be '')
         if diffCmd is '':
             diffCmd = DIFF_CMD
             diffParams = DIFF_PARAMS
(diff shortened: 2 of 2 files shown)


buhtz:
Thanks a lot for this contribution. I'll doing some manual testing now.
You can do my suggested modifications regularily on your local machine and then simply commit and push. After `git push` this PR is updated automatically.

**IMPORTANT**: As maintainer of the "original" repo I am also able to comit and push into your own branch. So please have an eye on that and "update" your local branch if I do modifications via `$ git fetch --prune --all && git pull`.

**EDIT2**: Now you see under this comment that I added a new commit.

buhtz:
Let me add two more wishes. It is up to you to decide if you want to handle them in this PR or if you prefer a separate PR. Maybe the latter is easier to handle.

1. The "Compare" button should be disabled when `CMD_DIFF is None` and of course enabled when it is set.
2. The Options dialog should check if the given diff command is valid (via `tools.checkCommand()`) when clicking on the "OK" button. If not valid the dialog should not save that command and should not close, except when pressing "Cancel".

stcksmsh:
I've made some changes, now the button is disabled when no valid command exists, and the user supplemented command is checked for validity before closing. Could you please look at [these two lines]( https://github.com/stcksmsh/backintime/commit/88ba687079367eff51a9fd7473feceb2222a2d20#diff-7078d6e6caa8c9d2ee5d39d0aeb267d5faed6ad86cd63ee03af94cd6561b3a52R375). Do you think there is a more elegant way around this button update?

buhtz:
> I've made some changes

It seems you did not see the "review comments" I made in the code? You can see them also in this discussion thread. The key point for me is to use `None` instead of an empty string (`''`) if `DIFF_CMD` is not set.

> please look at [these two lines](https://github.com/stcksmsh/backintime/commit/88ba687079367eff51a9fd7473feceb2222a2d20#diff-7078d6e6caa8c9d2ee5d39d0aeb267d5faed6ad86cd63ee03af94cd6561b3a52R375). Do you think there is a more elegant way around this button update?

Looks OK to me. It seems clear to the readers what does happen there.
A bit pedantic I would suggest to use [PEP8](https://peps.python.org/pep-0008/) conform variable names. This means not `CamelCase` but with underscores. In this case it would be `diff_cmd`. But this is not a hard rule in the BIT team. Most of the existing code in BIT do not follow PEP8 yet but I am working on it. ;)

buhtz:
Are you working on this or can I add some commits?

stcksmsh:
I'm currently swamped with university work, so I am not working on it. I also do not know where to go further, so commits would be appreciated.

buhtz:
> so commits would be appreciated.

OK, I am checking on it. I want to do some more manual testing. I am not sure if this will work in all cases.

buhtz (approved):
Let's wait some days to cool down.

Issue #1298, opened 2022-09-16
Refactor unittest: `test_sshCopyId()` runs only for Germar


buhtz:
This is about [`TestSshKey.test_sshCopyId()`](https://github.com/bit-team/backintime/blob/f1ec35fd039c3d25bad8495eb27c8b27dc71e581/common/test/test_sshtools.py#L244) in `common/test/test_sshtools.py`

https://github.com/bit-team/backintime/blob/f1ec35fd039c3d25bad8495eb27c8b27dc71e581/common/test/test_sshtools.py#L244

That is a nice skipping reason. 😄 This is the only skipped test left before all test go green.

If someone like to investigate that further and maybe find a good solution to solve this....

buhtz:
### Feature I found
I investigated that unittest and the feature/behavior behind it. Back In Time offers the feature to generated a new key pair and upload its public part to the remote server.

I was not aware that this is possible. I also assume that most users are also not aware and that is rarely used. Possible bugs are unknown because of less users.

The feature can be activated behind the plus button:
![image](https://github.com/bit-team/backintime/assets/11861496/0622ef2d-fb62-4558-8c64-cc2c59a78f1b)

### Discussion if we need that feature
I think we need discussion here.

I don't see a need for a feature like this. People how know how to setup SSH server/client and a key pair don't need BIT frontend for that. They do it on the shell or on a SSH-related GUI (e.g. putty on Windows). Or they still have setup their passwordless SSH login in the past. No one would die if this feature isn't in BIT anymore.

The plus-button could be replaced with a link to an tutorial how to setup passwordless-SSH-login.

Looking it the other way the GUI should be improved to make this feature more prominent to the users.

### Bug ?
There is also maybe a "bug". The plus button is disabled when creating a new SSH snapshot profile. It is disabled because there is always the key file pre selected. This happens on my system all the time. I assume that the existing `id_rsa` file is selected automatically if it exists on a system.

https://github.com/bit-team/backintime/blob/848655ae6917e0c2a9f4565941d30fa9b0fedda3/qt/settingsdialog.py#L254-L259

So I assume this feature is only active for people who don't have a key-pair "installed" yet.

### My vote and conclusion
In the end I see a lot of work for a feature that no one needs, that is not in the scope of BIT and where much better tools exist. I vote for removing the whole feature.

Germar:
[...]

PR #1688, opened 2024-04-14
Implement SSH Proxy Host

cgrinham:
Implemented options for using a proxy/bastion host with SSH backups.
First contribution so please let me know if you are happy with the style, I've tried to maintain the existing style but balance it with pep8 where possible. I think I caught all the places that are required for this but this is my first dive into this codebase so I wouldn't be surprised if I missed some things.

I separated out the ssh-copy-id command generation into its own function so I can run tests on it. I was thinking of doing the same for the ssh default args - let me know what you think.

image

This is my first time working with any desktop UI, more than open to suggestions of how it should work. I tried to be consistent with the rest of the dialog, however I think a checkbox or similar might be good for enabling these options.

Additionally, when doing a snapshot I found that because the `backup` method unmounts the mount at the end, the snapshot list no longer works and shows only "Now". If I comment the unmount out, it works as expected and shows the existing and new snapshots in the list. Is this a known issue?

### DIFF
--- a/common/test/test_sshtools.py
+++ b/common/test/test_sshtools.py
@@ -17,6 +17,7 @@
 
 import os
 import sys
+import random
 import subprocess
 import stat
 import shutil
@@ -399,43 +400,60 @@ class TestStartSshAgent(generic.SSHTestCase):
             self.ssh.startSshAgent()
 
 
-class SSHCopyIDTestCase(unittest.TestCase):
-
-    def test_ssh_copy_id_command_default_port(self):
+class SSHCopyID(unittest.TestCase):
+    def test_complete_command(self):
+        """Complete generated command"""
         command = sshtools.sshCopyIdCommand(
             generic.PRIV_KEY_FILE,
             'user',
             'non_existing_host',
-        )
-        self.assertEqual(
-            command,
-            ['ssh-copy-id', '-i', generic.PRIV_KEY_FILE, '-p', '22', 'user@non_existing_host']
-        )
-
-    def test_ssh_copy_id_command_custom_port(self):
-        command = sshtools.sshCopyIdCommand(
-            generic.PRIV_KEY_FILE,
-            'user',
-            'non_existing_host',
-            port='23'
-        )
-        self.assertEqual(
-            command,
-            ['ssh-copy-id', '-i', generic.PRIV_KEY_FILE, '-p', '23', 'user@non_existing_host']
-        )
-
-    def test_ssh_copy_id_command_with_proxy(self):
-        command = sshtools.sshCopyIdCommand(
-            generic.PRIV_KEY_FILE,
-            'user',
-            'non_existing_host',
-            proxy_user='non_existing_proxy_user',
-            proxy_host='non_existing_proxy_host',
-            proxy_port='24',
         )
         self.assertEqual(
             command,
             [
-                'ssh-copy-id', '-i', generic.PRIV_KEY_FILE, '-p', '22', '-o',
-                'ProxyJump=non_existing_proxy_user@non_existing_proxy_host:24', 'user@non_existing_host']
+                'ssh-copy-id',
+                '-i',
+                generic.PRIV_KEY_FILE,
+                '-p',
+                '22',
+                'user@non_existing_host'
+            ]
+        )
+
+    def test_default_port(self):
+        """Default port used"""
+        command = sshtools.sshCopyIdCommand(
+            generic.PRIV_KEY_FILE,
+            'user',
+            'non_existing_host',
+        )
+        self.assertEqual(command[4], '22')
+
+    def test_custom_port(self):
+        """Custom (random) port"""
+        custom_port = str(random.randint(23, 128))
+
+        sut = sshtools.sshCopyIdCommand(
+            generic.PRIV_KEY_FILE,
+            'user',
+            'non_existing_host',
+            port=custom_port
+        )
+        self.assertEqual(sut[4], custom_port)
+
+    def test_ssh_copy_id_command_with_proxy(self):
+        """Used proxy"""
+        proxy_user = 'non_existing_proxy_user'
+        proxy_host = 'non_existing_proxy_host'
+
+        sut = sshtools.sshCopyIdCommand(
+            generic.PRIV_KEY_FILE,
+            'user',
+            'non_existing_host',
+            proxy_user=proxy_user,
+            proxy_host=proxy_host
+        )
+
+        self.assertTrue(sut[6].startswith(
+            'ProxyJump=non_existing_proxy_user@non_existing_proxy_host')
         )
--- a/common/test/test_sshtools.py
+++ b/common/test/test_sshtools.py
@@ -35,8 +35,10 @@ import sshtools
 import tools
 from exceptions import MountException
 
+SKIP_MESSAGE_SSH = 'Skip as this test requires a local ssh server, public ' \
+                   'and private keys installed'
 
-@unittest.skipIf(not generic.LOCAL_SSH, 'Skip as this test requires a local ssh server, public and private keys installed')
+@unittest.skipIf(not generic.LOCAL_SSH, SKIP_MESSAGE_SSH)
 class TestSSH(generic.SSHTestCase):
     # running this test requires that user has public / private key pair created and ssh server running
 
@@ -336,7 +338,7 @@ class TestSshKey(generic.TestCaseCfg):
                 if os.path.exists(knownHostsSic):
                     shutil.copyfile(knownHostsSic, knownHosts)
 
-@unittest.skipIf(not generic.LOCAL_SSH, 'Skip as this test requires a local ssh server, public and private keys installed')
+@unittest.skipIf(not generic.LOCAL_SSH, SKIP_MESSAGE_SSH)
 class TestStartSshAgent(generic.SSHTestCase):
     # running this test requires that user has public / private key pair created and ssh server running
     SOCK = 'SSH_AUTH_SOCK'
@@ -399,7 +401,6 @@ class TestStartSshAgent(generic.SSHTestCase):
         with self.assertRaises(MountException):
             self.ssh.startSshAgent()
 
-
 class SSHCopyID(unittest.TestCase):
     def test_complete_command(self):
         """Complete generated command"""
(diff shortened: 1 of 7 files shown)


buhtz:
Dear Christie,

Thank you very much for your contribution on behalf of the team and myself. I am pleased with the style, the unit tests, and your thoughtful refactoring. I appreciate your contribution.

Our team currently comprises three members, along with the former maintainer in a supporting role. However, apart from myself, all others are temporary unavailable for various reasons. Therefore, it may take some time before we can thoroughly discuss and decide on all the details of your PR.

First and foremost, I would like to gain a better understanding of the use case. I have not previously utilized an SSH proxy. Is it similar to a "jump host"? Additionally, could you share details of your real setup where this feature proves useful to you? Understanding your experience would greatly assist me in visualizing the use case myself.

The issue regarding "unmount" and the snapshot list requires investigation. Does this problem occur exclusively when using the proxy, or does it also manifest with "regular" SSH hosts?

We are dedicated to enhancing the codebase. Despite its current non-compliance with PEP8 standards, our goal remains to align with PEP8 guidelines, acknowledging its 15-year legacy and our position as the third generation of maintainers.

I will consider how best to enhance the dialogue, perhaps by visually disabling the SSH proxy section and enabling it through a checkbox. Ultimately, a redesign of the entire dialogue is necessary, but the current codebase is not conducive to such changes. Therefore, I intend to implement the solution in a simple manner and postpone the redesign.

Best regards,
Christian

EDIT: Note to myself: Have a look at #1298

cgrinham:
Hi Christian, thanks for your reply!
Yes this is exactly that - jump host. Wasn't sure which term to go with.
My set up is probably a bit unusual which is why my screenshot maybe doesn't explain it well.
I have a server at home (ProxyServer) and another at a relative's (BackupServer). BackupServer has a reverse tunnel to ProxyServer. I am backing up my laptop to BackupServer via ProxyServer. This way I don't have to open ports on the router at my relative's house.

A use case I have also seen would be to limit the attack vector of a backup server by eg. accepting only requests from a certain IP address (the proxy server).

A simpler example would be: 
image

> The issue regarding "unmount" and the snapshot list requires investigation. Does this problem occur exclusively when using the proxy, or does it also manifest with "regular" SSH hosts?
I tested with and without my changes and the issue persisted on both.

I would be very keen to help with some refactoring work - I can identify some small places to start to show what I'd do then would be great to get some feedback about whether people are happy with my contributions. Very keen to add more tests also.

buhtz:
I am taking it from here, working on your branch and will come up with some modifications in the next days.

Sidenote about translation and localization:

    QLabel(_('Port') + ':', self)

In most cases it makes more sense to make punctuations and other signs part of the string to translate. This make sense for right-to-left (RTL) languages and bidirectional-reading (BIDI) environments.

cgrinham:
@buhtz okay cool thanks
re: translation, I just copied it from further up the code but will keep that in mind and good to know

buhtz:
Hello Christie,
I have a side question. Do you somehow modified the file mode (file permissions) of the files in the repo? Can you please `cd` into your local repos folder and show me the output of `git config --list --show-scope --show-origin | grep filemode`. And the output of `ls -l updatecopyright.sh` please. Thanks.

buhtz:
This is how it looks like and behave now.
Peek 2024-04-21 08-49

The label "SSH Proxy (optional)" need to be modified. I would propose to use "SSH Proxy" (the term proxy is used in the man page, too) and add a tooltip to the whole widget explaining the purpose, use alternative term "Jump host" and reference the command switches used (e.g. `ssh -J`).

# Needs discussion
I also vote to finally remove [`TestSshKey.test_sshCopyId()`](https://github.com/bit-team/backintime/blob/f1ec35fd039c3d25bad8495eb27c8b27dc71e581/common/test/test_sshtools.py#L244) (#1298). A system test, runs only on Germars machine. There is also not much value in it but has lot of maintenance costs. It simply tests an extern tool which has not much value.
# Open problems
The thing with the lost snapshot list needs further investigation.

cgrinham:
> Hello Christie, I have a side question. Do you somehow modified the file mode (file permissions) of the files in the repo? Can you please `cd` into your local repos folder and show me the output of `git config --list --show-scope --show-origin | grep filemode`. And the output of `ls -l updatecopyright.sh` please. Thanks.

Not knowingly!
here is the output:
```
local	file:.git/config	core.filemode=true
```

```
-rwxrwxr-x 1 christie christie 344 Apr  1 20:31 updatecopyright.sh
```

cgrinham:
> The label "SSH Proxy (optional)" need to be modified. I would propose to use "SSH Proxy" (the term proxy is used in the man page, too) and add a tooltip to the whole widget explaining the purpose, use alternative term "Jump host" and reference the command switches used (e.g. `ssh -J`).


New UI looks great - I agree with the checkbox the "(optional)" is no longer needed.

> # Needs discussion
> I also vote to finally remove [`TestSshKey.test_sshCopyId()`](https://github.com/bit-team/backintime/blob/f1ec35fd039c3d25bad8495eb27c8b27dc71e581/common/test/test_sshtools.py#L244) (#1298). A system test, runs only on Germars machine. There is also not much value in it but has lot of maintenance costs. It simply tests an extern tool which has not much value.

I would agree - these tests are also quite hard to read as a newcomer to the codebase

common/test/test_sshtools.py
```diff
@@ -396,3 +398,62 @@ def test_startSshAgentMissing(self, mockWhich):
...
+
+        sut = sshtools.sshCopyIdCommand(
+            generic.PRIV_KEY_FILE,
+            'user',
+            'non_existing_host',
+            proxy_user=proxy_user,
+            proxy_host=proxy_host
+        )
+
+        self.assertTrue(sut[6].startswith(
```
cgrinham: @buhtz your update to this test means that we are no longer testing the inclusion of the port for the proxy

buhtz: Yes, this is correct and on purpose. The test about "including the default port when not specified" (the behavior to test) is done in `def `test_default_port()`.

cgrinham: @buhtz this is different - the proxy part of the command also has its own port value that would be useful to test

buhtz:
Here is the new visual with cleaned label and a tool tip. Any re-wording suggestions?
image

cgrinham:
> Here is the new visual with cleaned label and a tool tip. Any re-wording suggestions? image

Looks good to me

common/test/test_sshtools.py
buhtz: Yes, I was definitive to radical here. 😈

buhtz: This test brings me to a question: sshCopyIdCommand() do not use a default port value. What should be the expected behavior?

```
def sshCopyIdCommand(
    pubkey,
    user,
    host,
    port='22',
    proxy_user=None,
    proxy_host=None,
    proxy_port=None,
    cipher=None
):
```

I would argument to give `proxy_port` the default value `'22'` because it is the same for `port`. Does it make sense?

cgrinham: the ssh-copy-id command will default to 22 if you don't supply a port, so I don't think it is necessary?

buhtz:
> Additionally, when doing a snapshot I found that because the `backup` method unmounts the mount at the end, the snapshot list no longer works and shows only "Now". If I comment the unmount out, it works as expected and shows the existing and new snapshots in the list. Is this a known issue?

I need to setup a jump host scenario in a VM to investigate this further.
 - Does this problem occur exclusively when using the proxy, or does it also manifest with "regular" SSH hosts?
 - Can you reproduce the problem? Please run BIT with `--debug` argument and show me the terminal output.
 - If the time line do not dissapear without using an SSH proxy then I hypothesize that `qt/app.py::FillTimeLineThread()` and `common/snapshots.py::iterSnapshots()` are somehow involved. They do not fail if nothing is mounted. They are simply silent. When a regular (without jump host) SSH snapshots was taken there might be a re-mount somewhere hidden in the code we do not know know about? Maybe the debug output will give us a hint about it.

**EDIT**: The failing TravisCI is an issue on their own servers and not related to this PR.

cgrinham:
> * Does this problem occur exclusively when using the proxy, or does it also manifest with "regular" SSH hosts?

This happened with any SSH host.

Will try get you the output later today - I tested on the main branch and I saw the same behaviour

buhtz:
If you are able to reproduce the problem with the snapshot timeline on "main" branch independent from this PR then I would suggest you to report it in a separate issue.

common/test/test_sshtools.py
buhtz: The command on the shell will default to 22 but not the ssCopyIdCommand() function. I used `None` as default port.
The question is who should "inject" the default port? Should BIT do it (if `proxy_host is None`) or should BIT not using any port if it is not set and let the shell command ssh-copy-id decide about default port?

cgrinham: If I didn't set a port in BIT, then I would expect the default port to be decided by ssh-copy-id, not by BIT, as ssh-copy-id is likely the authority on what the defaults of its args should be.

buhtz: Shouldn't it then be the same for the regular `port=` argument? Any reasons against this behavior?

buhtz: Damn... Now I brock BIT. :rofl:
I tried to leave the (proxy)port field empty to allow the SSH command itself to decide about the default port. But having an empty port value in the config break so many other parts of BIT. The code is not robust enough for this. I could fix it but it is to much and to risky. I will role back for this PR and explicit use "22" as default port specified by BIT. So BIT will always use a port value.

cgrinham: ha!
Yes seems like it is probably better to be consistent with the BIT codebase than try to introduce a new way of handling it

buhtz:
> I tested on the main branch and I saw the same behaviour

Do you tested it on the branch named `main` or on the primary/default branch of the repository named `dev`? Please report back at #1705 where I track this specific issue.

I will merge the PR because the bug you describe is not reproducible. I setup a jumps host and it worked fine. Because the bug is not critical I decide to merge this PR but keep the separate issue #1705 open for a while.

cgrinham:
> > I tested on the main branch and I saw the same behaviour
> 
> Do you tested it on the branch named `main` or on the primary/default branch of the repository named `dev`? Please report back at #1705 where I track this specific issue.
> 
> I will merge the PR because the bug you describe is not reproducible. I setup a jumps host and it worked fine. Because the bug is not critical I decide to merge this PR but keep the separate issue #1705 open for a while.

I tested it on the dev branch - haven't tried the main branch.

Okay I'll try get some time to reproduce it this week.

Thanks for merging!

PR #1696, opened 2024-04-21
Refactor: Replace Config.user() by getpass.getuser()

### LINKED ISSUE #1694: Remove: Config.user()
Remove method `common/config.py::Config.user()` without replacement.

Use [`getpass.getuser()`](https://docs.python.org/3/library/getpass.html#getpass.getuser) directly. It is Python in-build, the recommended way, portable and does use `pwd` module as a fallback.

buhtz:
Close #1694

### DIFF
--- a/common/test/test_backintime.py
+++ b/common/test/test_backintime.py
@@ -155,7 +155,9 @@ under certain conditions; type `backintime --license' for details.
 
         line_contains_to_exclude = [
             "Gtk-WARNING",
-            "qt.qpa.plugin: Could not find the Qt platform plugin"
+            "qt.qpa.plugin: Could not find the Qt platform plugin",
+            'qt.dbus.integration: Could not connect "org.freedesktop.IBus" '
+            'to globalEngineChanged(QString)',
         ]
 
         # remove lines via startswith()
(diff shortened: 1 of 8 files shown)


qt/settingsdialog.py
```diff
@@ -2373,9 +2373,14 @@ def searchConfig(self, path):
 
                 try:
                     cfg = config.Config(cfgPath)
+
                     if cfg.isConfigured():
                         return cfg
-                except:
+
+                except Exception as exc:
```
buhtz, line 2380: Catching all exception but doing nothing about it (`pass`). Can not imagine how much bugs this may hide or trigger.

common/test/test_backintime.py
```diff
@@ -155,7 +155,9 @@ def test_local_snapshot_is_successful(self):
 
         line_contains_to_exclude = [
             "Gtk-WARNING",
-            "qt.qpa.plugin: Could not find the Qt platform plugin"
+            "qt.qpa.plugin: Could not find the Qt platform plugin",
+            'qt.dbus.integration: Could not connect "org.freedesktop.IBus" '
```
buhtz, line 159: This error message appeared in the output of my machine.
IBus is related to the setup of input method (switching between German and Japanese keyboard).

PR #1700, opened 2024-04-23
doc(README): Remove known problem about "apt-key keyring" / Copyright adjustments

buhtz:
Removed known problem about "apt-key keyring" because not relevant anymore. Related issue is closed.
Also some minor adjustments.

About "Copyright" I modified in the README.md:
I did some intensive research about that topic related to another project. What I understood and learned.
 - Copyright is given always and implicit even if it is not written down somewhere. See [Berne Convention](https://en.wikipedia.org/wiki/Berne_Convention).
 - No need to increment the Copyright year. Copyright counts minimum 50 years **after the death** of the author ([source](https://matija.suklje.name/how-and-why-to-properly-write-copyright-statements-in-your-code#why-not-bump-the-year-on-change)). Depending on the country that law is used, it is longer than 50 years (usually 70). Dan started BIT round about 1997 (emtiu knows better) and to my knowledge he still is alive. So we talk about year 2080 or later.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -187,6 +187,24 @@ Alternatively use `make test-v` for a more verbose output. The `make` system
 will use `pytest` as test runner if available otherwise Python's own `unittest`
 module.
 
+## SSH
+
+Some tests require an available SSH server. They get skipped if this is not the
+case. The goal is to log into the SSH server on your local computer via
+`ssh localhost` without using a password:
+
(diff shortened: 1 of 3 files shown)


aryoda:
> Removed known problem about "apt-key keyring" because not relevant anymore. Related issue is closed.

Perhaps we move the the problem description and link to the solution (= issue comment) to our FAQ since other user may still have the problem (even though the issue is closed for us)...

aryoda (approved):

PR #1702, opened 2024-04-23
Replace wrapLines() with Python's textwrap.wrap() #1699

### LINKED ISSUE #1699: Replace wrapLines() with Python's textwrap.wrap()
# Introduction
Welcome to the project, if you pick up this Issue because of the "GOOD FIRST USE" label. You will be mentored through the process if you want. First the Issue is explained. At the end you will find some guidance for first contributors. Please do not hesitate to ask questions. Your solution don't need to be perfect.

# Problem
BIT has it's own function `logger.py::wrapLines()` but Python offer the same functionality with [`textwrap.wrap()`](https://docs.python.org/3/library/textwrap.html#textwrap.wrap).

# Solution
Replace that function with [`textwrap.wrap()`](https://docs.python.org/3/library/textwrap.html#textwrap.wrap). Check via `git grep` where it is used in the code. It is not much. Easy to replace.

Argument `new_line_indicator` is used in `logger.py::_do_syslog()`. Use `textwrap.wrap(..., subsequent_indent='CONTINUE: ')` instead.

# Your next steps
1. If this is your first contribution in this project please introduce your self and tell us about your skills, wishes and plans.
2. Read the existing [contributors documentation](https://github.com/bit-team/backintime/blob/dev/CONTRIBUTING.md).
3. We can develop the next steps in the further discussion. Don't hesitate to ask.

# Related code
https://github.com/bit-team/backintime/blob/5aa0746bf36bb49d95969e558ca4fdde38dfaa86/common/logger.py#L149-L184

Hunter-Builds:
Let me know if this will do.
Should I add it to CHANGES?

Close #1699

### DIFF
--- a/common/logger.py
+++ b/common/logger.py
@@ -19,7 +19,6 @@ import syslog
 import os
 import sys
 import atexit
-import textwrap
 
 import bcolors
 
@@ -58,9 +57,8 @@ def closelog():
 
 
 def _do_syslog(message: str, level: int) -> str:
-    for line in textwrap.wrap(message):
-        syslog.syslog(level, '{}{}: {}'.format(
-            SYSLOG_MESSAGE_PREFIX, _level_names[level], line))
+    syslog.syslog(level, '{}{}: {}'.format(
+        SYSLOG_MESSAGE_PREFIX, _level_names[level], message))
 
 
 def error(msg, parent=None, traceDepth=0):
--- a/common/logger.py
+++ b/common/logger.py
@@ -19,6 +19,7 @@ import syslog
 import os
 import sys
 import atexit
+import textwrap
 
 import bcolors
 
@@ -57,7 +58,7 @@ def closelog():
 
 
 def _do_syslog(message: str, level: int) -> str:
-    for line in wrapLine(message):
+    for line in textwrap.wrap(message):
         syslog.syslog(level, '{}{}: {}'.format(
             SYSLOG_MESSAGE_PREFIX, _level_names[level], line))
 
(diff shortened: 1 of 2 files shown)


buhtz:
Hello Tom,
Thank you for your contribution. I appreciate it. I will review it.

Would you mind to introduce yourself a bit? How did you find that issue and the project?

Regarding to [Common Changelog](https://common-changelog.org/) no need for a changelog entry because behavior of BIT is not modified.

Best,
Christian

buhtz:
OK, now I had a closer look to the modifications.

I realized that the original `logger.wrapLine()` defaults to a line width of **950** characters.

After some research I assuming that this is related to limit of some syslog implementations from the **past**. Today usual syslog implementations do not have a limit. The limit depends on the network (UDP package size for example). Some old rsyslogd implementations do have a limit at 2k (some at 8k) characters. If syslog reach the limit a line is truncated without indicating an error.

Python's syslog do not mention any limit to the length of message.

Checking my syslog (`$ journalctl --no-pager -a | grep backintime | less`) I really can find some entries that where wrapped by BIT (search for `CONTINUE:` string). They are related to debug output and long commands (e.g. rsync itself). I see no need to wrap.

In short: The limit is not relevant anymore and maybe never was.

So I propose to just remove the wrapping from the `_do_syslog()` function and just log the message at it is.

buhtz:
> Would you like to remove the `_do syslog()` wrapping here or should it
> be a separate pull request?

Just do ist in this PR.Thank You.

buhtz (approved):
Great work. Thank you.
I will merge it in some days. A little cool-off-periode for PRs is team policy.

PR #1709, opened 2024-05-02
refactor!: Remove "filesystem_mounts" and "snapshot_version" field in backup "info" file

### LINKED ISSUE #1684: Remove dead code related to "filesystem_mounts" and "snapshot_version" field in backup "info" file
https://github.com/bit-team/backintime/blob/d33f35e10b4c3027b3b12881152c157eb1b2cc8c/common/snapshots.py#L1140

The field "filesystem_mounts" (and also "snapshot_version") is written into the "info" file of each backup. It is never used anywhere (checked via grep).

 - Remove that field.
 - Remove `tools.filesystemMountInfo()` because it is used only once at this location.
 - Remove the related unit tests.

My assumption is that this was once added for debug reasons.

**EDIT**: The same goes for the field "snapshot_version". Never used anywhere.

buhtz:
Each snapshot contains an "info" file. This PR remove the two fields "filesystem_mounts" and "snapshot_version" from it because they are not used.

Additionally related code and unit tests are removed or modified.

There is no strict need but I created a changelog entry if it might impacts someone.

Close #1684

### DIFF
--- a/common/snapshots.py
+++ b/common/snapshots.py
@@ -1117,27 +1117,32 @@ class Snapshots:
                         f' command was {cmd}. Also see the previous '
                         'WARNING message for a more details.', parent=self)
 
-    def backupInfo(self, sid):
+    def _backup_info_file(self, sid):
         """
-        Save infos about the snapshot into the 'info' file.
+        Save infos about the snapshot into the 'info' file. The result is
+        stored in 'sid.info' also.
 
         Args:
-            sid (SID):  snapshot that should get an info file
+            sid (SID): Snapshot that should get the info file.
         """
-        logger.info("Create info file", self)
+        logger.debug(
+            f'Create info file for snapshot "{sid.displayName}".', self)
+
         machine = self.config.host()
         user = getpass.getuser()
         profile_id = self.config.currentProfile()
+
         i = configfile.ConfigFile()
-        i.setIntValue('snapshot_version', self.SNAPSHOT_VERSION)
         i.setStrValue('snapshot_date', sid.withoutTag)
         i.setStrValue('snapshot_machine', machine)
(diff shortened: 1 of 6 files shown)


common/snapshots.py
```diff
@@ -1117,27 +1117,32 @@ def backupConfig(self, sid):
...
         """
-        Save infos about the snapshot into the 'info' file.
+        Save infos about the snapshot into the 'info' file. The result is
+        stored in 'sid.info' also.
 
         Args:
-            sid (SID):  snapshot that should get an info file
+            sid (SID): Snapshot that should get the info file.
         """
-        logger.info("Create info file", self)
```
buhtz, line 1127: IMHO BIT is to verbose in INFO mode and also give meaningless output.

common/snapshots.py
```diff
@@ -1117,27 +1117,32 @@ def backupConfig(self, sid):
...
         i = configfile.ConfigFile()
-        i.setIntValue('snapshot_version', self.SNAPSHOT_VERSION)
         i.setStrValue('snapshot_date', sid.withoutTag)
         i.setStrValue('snapshot_machine', machine)
         i.setStrValue('snapshot_user', user)
         i.setIntValue('snapshot_profile_id', profile_id)
         i.setIntValue('snapshot_tag', sid.tag)
-        i.setListValue('user', ('int:uid', 'str:name'), list(self.userCache.items()))
-        i.setListValue('group', ('int:gid', 'str:name'), list(self.groupCache.items()))
-        i.setStrValue('filesystem_mounts', json.dumps(tools.filesystemMountInfo()))
```
buhtz, line 1140: This call to `tools.filesystemMountInfo()` was my inital trigger to work on this. Discovered this function and wondered for what it might be useful.

Issue #1585, opened 2023-12-13
"Backup folders" list should reflect the selected snapshot


buhtz:
# Problem description
![grafik](https://github.com/bit-team/backintime/assets/11861496/9441c28e-c2db-4368-a85f-76b76ee4bd91)

Look on the right column with four folders. One of them is named `tab-cloud`.
Look in the middle column at "Backup folders". The `tab-cloud` is missing in that list.

The `tab-cloud` was included in the past. But I removed it from the Include list.

The problem is that "Backup folders" do reflect the current state of the  Include list. But it should reflect the Include list at that point the snapshot was made.

# How should it be
Regarded to the screenshot the folder `tab-cloud` should appear in the middle column ("Backup folders") because it exists in the snapshot. It does not matter that it does not exist in the real filesystem anymore.

# What need to be done
Technically it is not complex or hard to understand but for beginning contributors this is an Issue causing a lot of work. First of all the code need be investigated. Please describe a possible approach before writing to much code.

emtiu:
Actually, I think this is rather important. It might lead users to think that they have lost data, because a folder that was included in the backup is not displayed.

In general, backintime is not great at handling Profiles with frequent changes to the Include/Exclude list. I think this should receive attention.

PR #1707, opened 2024-04-28
doc: Contributing info with section about our Pull Request treatment

buhtz:
I added a section about what happens after a pull request is opened. e.g. our 1-week-rule. Additionally did some minor fixes to the docu.

I would like to combine and add the "About team & roadmap" text I send you via email into this PR. Thast why I set it do "Draft". So I'll wait for your feedback.

**EDIT**: ~~I also would like to somehow add information about the team and a broad roadmap to this PR. But I am not sure where to place it. I have a [draft of that information text](https://github.com/buhtz/spielwiese/blob/main/team_roadmap.md) in a separate repo.~~

PR #1714, opened 2024-05-05
fix: "Backup folders" list should reflect the selected snapshot

rafaelhdr:
issue reference: https://github.com/bit-team/backintime/issues/1585

* List places based on snapshot listing it's folders
* On select snapshot, it updates places again

I hope this fulfill the bug mentioned but let me know if I should improve anything.

Kind
Rafael


buhtz:
Hello Rafael,
Thank you for your contribution and taking the time to improve Back In Time.
On behalf of the team, I appreciate it.

We will review your PR as soon as possible.

Best,
Christian

buhtz:
Hello Rafael,
I took the advantage doing some refactoring. But your solution looks nice and smooth.

There might be a bug and I am not sure how to fix. Maybe you have an idea. Please try these steps to reproduce:
1. Make sure to have two profiles in BIT.
2. Select one specific snapshot in the timeline.
3. Change the profile via the combo box in the main toolbar to the second profile.

Then you could see a `FileNotFoundError` inside `updatePlaces()`. Do you?
Technically the reasons is that updatePlaces() is called before `self.sid = sid` is called but the previous snapshot profile is unmounted. That is why the backup pathes can't be find anymore. To my understanding it is a matter of interfering events. In my logic the places-list should be updated every time the timeline is updated.

I tried several things but got stuck. Maybe you can find a good solution?

rafaelhdr:
Hello @buhtz,

Thank you very much for the feedback and modifications provided.
I am very new to BIT, but I think I understood what you meant.

I don't know how to solve it, but I will take a look on it.

buhtz:
> I am very new to BIT, but I think I understood what you meant.

😄 Members of the current maintenance team also not the founders of BIT. This code is foreign for us, too. Are you a regular user of BIT or you just picked up the Issue?

> I don't know how to solve it, but I will take a look on it.

I know the code is not structured and isolated enough. So if you have any idea in that direction... But don't let this PR become to big. Restructuring code via introducing new modules, classes etc should be done in a separate PR.

rafaelhdr:
Gotcha. Thanks anyway :)

I just picked the issue... I am not a regular user. I wanted to practice Python aside of the job, and found bit in some website that shares projects in need of contributions.

Thank you for being so kind. I will let you know. First I need some time to get more familiar with the project, but I will let you know for sure if I have more suggestions :)

rafaelhdr:
@buhtz what do you think of this?

```
            if not os.path.isdir(self.sid.pathBackup(base)):
                # Folder not mounted. We can skip for the next updatePlaces()
                return
```

I realized it calls the `updatePlaces` twice when we change the profile via combo box (three times after my change).

With this change, it would ignore if the folder doesn't exists (as it is expected to have another `updatePlaces()`)

Let me know what you think. And it is ok if you think we could make it in a different way.

buhtz (approved):

buhtz:
> @buhtz what do you think of this?

Looks OK to me. Thank you for fixing it.

> Let me know what you think. And it is ok if you think we could make it in a different way.

In a perfect world I better would like to avoid the useless call of updatePlaces(). But the code is not prepared for "elegant" things like this. Currently your fix is OK.

rafaelhdr:
Yeah. I totally agree with you. Hopefully it will be better in the future.

Now it is all good from my side, right? It is up to the team to merge it later...

Sorry for bothering, but how is the normal flow here for PRs?

buhtz:
> Now it is all good from my side, right?

Very good. Right!

> It is up to the team to merge it later...
> Sorry for bothering, but how is the normal flow here for PRs?

No need to be sorry. 😃 There is another PR (#1707) adding this information to the contribution info. Maybe you want to review it because you are the target group? Do you miss information in it?

Yes, the PR now is finished and ready to merge. We usually wait 1 week until merging to give the other maintainers the opportunity to review and interfere.

rafaelhdr:
I appreciated. It seems good, and answer my questions from here :)

PR #1733, opened 2024-05-28
build: Remove debian folder

### LINKED ISSUE #1548: Remove debian-folder
The `debian` folder is used by Germar to create deb-files distributed via a PPA.

I propose to migrate that folder somewhere else. I would suggest to migrate it to Launchpad itself because Launchpad also host the related PPA.

That folder is specific to a bunch of GNU Linux distribution and therefor irrelevant for another bunch of distros (e.g. Arch based). This is no problem but natural. But it is not related to an upstream repository which should always be independent from a specific GNU Linux distribution or alternative distribution ways (Flatpak, Snap, ...). Things like this should maintained elsewhere.

Official Debian itself don't use this `debian` folder. Debian to have its [own repository](https://salsa.debian.org/jmw/pkg-backintime) maintaining such a folder.

I'll contact Germar via mail.

buhtz:
As discussed in #1548 the debian folder and related helper scripts are removed.

Currently they are conserved in a separate branch https://github.com/bit-team/backintime/tree/fix/1548_separate_debian_ppa . There is a new script in that branch temporary cloning the latest BIT repo to use it for deb-build. Myself I will also make a backup of that branch in my private archive in case the branch get lost.

The idea was to make the debian folder and the related PPA living in its own repository. Because the intention was to reduce maintenance burden and spare resources of the upstream project, I suggest not to create this repo in `bit-team`. Otherwise users would think that the `bit-team` maintain the PPA and answer PPA related support questions.

Close #1548


emtiu:
Are the Debian maintainers and the PPA creators already aware of this intended change?

buhtz:
> Are the Debian maintainers and the PPA creators already aware of this intended change?

PPA maintainers are informed several times.
Debian maintainers are not relevant because they don't use that folder. They maintain [their own debian folder](https://salsa.debian.org/jmw/pkg-backintime/-/tree/debian/debian?ref_type=heads) at there Salsa (GitLab) instance.

aryoda (approved):

Issue #1679, opened 2024-04-06
Make  EXTREMELY CLEAR that the Auto-remove options take priority over all the "Keep one snapshot for/per" options / Rename "Smart remove"


TiagoTiago:
I assumed I could always count on the older snapshots sticking around. Then yesterday I fucked up and `mount --bind` a huge folder from another drive inside one of the folders that gets regularly backuped, and today I found out I just got the single backup made yesterday, no versioning for anything left :(

I hope nothing important got wiped; but I do worry since there have been times I only discovered some days or even weeks later that I something got messed up and I managed to get it restored from the version from a few snapshots ago....

ps: I am already beating myself for not keeping up with 3-2-1, no need to pour any more salt on that wound

TiagoTiago:
Or even better, let the user choose whether any of the "one per" is sacred and survives even running out of space.

TiagoTiago:
Ugh, and while it was creating today's backup with that folder excluded, it eliminated yesterday's, I guess because it checked whether should erase whole snapshots before  noticing the option of just erasing the excluded folder from all snapshots... \*sighs\*


edit: Hm, wait, where is that option? I could swear it was there before.... Maybe it was a manually added option in the Expert tab got lost on on upgrade or let it slip when migrating to a new hardware years ago and never noticed? O r am I just not looking in the right place? Or is it expected to do that by default and it wouldn't show as an option?

edit2: Ok, I'm not insane, `--delete-excluded` is a real thing and it does look like it's present in the parameters BiT is passing to rsync.... Now the question is; with the versioning trickery BiT does, does that option do anything?

buhtz:
Dear Tiago,
Thanks for this report. You  have to help me a bit with understanding.

Please try to explain the problem in more details and steps. I did not understand your problem description.

Also please describe the situation and steps that triggered that problem.

And please provide some more basic information. Run `backintime --diagnostics` and post the complete output here. If this command do not work on your system with your version of BIT then please give us the name and version of your operating system, the version of BIT and where you installed BIT from.

Best,
Christian

TiagoTiago:
> Dear Tiago, Thanks for this report. You have to help me a bit with understanding.
> 
> Please try to explain the problem in more details and steps. I did not understand your problem description.
> 
> Also please describe the situation and steps that triggered that problem.
> 
[...]

PR #1756, opened 2024-06-07
fix: Global flock fallback to single-user mode if not enough permissions

### LINKED ISSUE #1743: [Errno 13] Permission denied: '/run/lock/backintime.lock' (on Arch)
Hi everyone,

I run the AUR package `backintime-git` on an up-to-date EndeavourOS installation. Since #1697 was merged, I unfortunately can not run my backups (which usually run automatically) as a user anymore (and I don't use backintime as/for root) as the moved lock file can not be written to the `/run/lock` folder:

```
$ backintime backup

Back In Time
Version: 1.4.4-dev.e6f7f4f3

Back In Time comes with ABSOLUTELY NO WARRANTY.
This is free software, and you are welcome to redistribute it
under certain conditions; type `backintime --license' for details.

  File "/usr/share/backintime/common/backintime.py", line 1209, in <module>
    startApp()
  File "/usr/share/backintime/common/backintime.py", line 531, in startApp
    args.func(args)
  File "/usr/share/backintime/common/backintime.py", line 783, in backup
    ret = takeSnapshot(cfg, force)
          ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/share/backintime/common/backintime.py", line 98, in takeSnapshot
    ret = snapshots.Snapshots(cfg).backup(force)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/share/backintime/common/snapshots.py", line 785, in backup
    with flock.GlobalFlock():
  File "/usr/share/backintime/common/flock.py", line 46, in __enter__
    self._flock_handle = self._file_path.open(mode)
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/pathlib.py", line 1013, in open
    return io.open(self, mode, buffering, encoding, errors, newline)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
PermissionError: [Errno 13] Permission denied: '/run/lock/backintime.lock'
```

Permissions for the /run/lock folder:
```
$ ls -l /run | grep lock
drwxr-xr-x    - root      31 Mai 10:05 lock
```

The `/etc/tmpfiles.d` folder is empty, `/usr/lib/tmpfiles.d/legacy.conf` contains the line
```
d /run/lock 0755 root root -
```
which matches the permissions of the created folder and also matches systemd's standard behaviour according to their [git logs](https://cgit.freedesktop.org/systemd/systemd/tree/tmpfiles.d/legacy.conf). To me it seems that this folder is actually not intended to be written to by the user in the default configuration (and only contains mostly device-specific stuff for me). Instead, user processes mostly create their lock files either in `/tmp` (which was done before and lead to #1697) or in `/run/user/UID`. However, using this folder (or folders for multi-user setups) for backintime as well likely clashes with the intention of having a global lock file for all users simultaneously.

I am not sure how to solve this problem in the best way. As a start, one could probably add some additional fallbacks to the lockfile locations in `flock.py` in case `/run/lock` is unwritable (including the old `/tmp` or `/run/user/UID`), however for a true global lock this likely needs some more work as mentioned (maybe some kind of daemon the backintime processes are communicating with).

Another thing I want to note: I only noticed the missing permissions for the lock file after triggering a manual backup (not `backup-job`). It might be a good idea to throw a desktop notification in case the lock couldn't be gathered (like when some data has no read permissions) such that the user can act accordingly.

Thanks a lot and kind regards

For completion, here's the output for `backintime --diagnostics`:
<details>

```
QSettings::value: Empty key passed
{
    "backintime": {
        "name": "Back In Time",
        "version": "1.4.4-dev.e6f7f4f3",
        "running-as-root": false,
        "latest-config-version": 6,
        "keyring-supported": true
    },
    "host-setup": {
        "OS": {
            "/etc/os-release": "EndeavourOS",
            "/etc/arch-release": "\n",
            "/etc/lsb-release": "DISTRIB_ID=\"EndeavourOS\"\nDISTRIB_RELEASE=\"rolling\"\nDISTRIB_DESCRIPTION=\"EndeavourOS Linux\"\nDISTRIB_CODENAME=\"rolling\"\n"
        },
        "platform": "Linux-6.9.2-zen1-1-zen-x86_64-with-glibc2.39",
        "system": "Linux #1 ZEN SMP PREEMPT_DYNAMIC Sun, 26 May 2024 01:30:09 +0000",
        "display-system": "wayland",
        "locale": "de_DE, UTF-8",
        "PATH": "/home/UsernameReplaced/.pip-apps/bin:/home/UsernameReplaced/.local/bin:/usr/lib/jvm/default/bin:/usr/bin/site_perl:/usr/bin/vendor_perl:/usr/bin/core_perl:/usr/local/bin:/usr/local/sbin:/sbin:/usr/sbin:/bin:/usr/bin:/home/UsernameReplaced/.zgen/unixorn/bitbucket-git-helpers.plugin.zsh/___/bin:/home/UsernameReplaced/.zgen/skx/sysadmin-util/___:/home/UsernameReplaced/.zgen/StackExchange/blackbox/___/bin:/opt/rocm/bin:/home/UsernameReplaced/.zgen/horosgrisa/zsh-dropbox/___/bin",
    },
    "python-setup": {
        "python": "3.12.3 main Apr 23 2024 09:16:07 CPython GCC 13.2.1 20240417",
        "sys.path": [
            "/usr/lib/python312.zip",
            "/usr/lib/python3.12",
            "/usr/lib/python3.12/site-packages"
        ],
        "qt": {
            "Version": "PyQt 6.7.0 / Qt 6.7.1",
            "Theme": "BeautySolar",
            "Theme Search Paths": [
                "/usr/share/icons",
                ":/icons"
            ],
            "Fallback Theme": "breeze",
            "Fallback Search Paths": []
        }
    },
    "external-programs": {
        "rsync": {
            "version": "3.3.0",
            "protocol": "31.0",
            "capabilities": {
                "file_bits": 64,
                "inum_bits": 64,
                "timestamp_bits": 64,
                "long_int_bits": 64,
                "socketpairs": true,
                "symlinks": true,
                "symtimes": true,
                "hardlinks": true,
                "hardlink_specials": true,
                "hardlink_symlinks": true,
                "IPv6": true,
                "atimes": true,
                "batchfiles": true,
                "inplace": true,
                "append": true,
                "ACLs": true,
                "xattrs": true,
                "secluded_args": "optional",
                "iconv": true,
                "prealloc": true,
                "stop_at": true,
                "crtimes": false
            },
            "optimizations": {
                "SIMD_roll": true,
                "asm_roll": false,
                "openssl_crypto": true,
                "asm_MD5": false
            },
            "checksum_list": [
                "xxh128",
                "xxh3",
                "xxh64",
                "md5",
                "md4",
                "sha1",
                "none"
            ],
            "compress_list": [
                "zstd",
                "lz4",
                "zlibx",
                "zlib",
                "none"
            ],
            "daemon_auth_list": [
                "sha512",
                "sha256",
                "sha1",
                "md5",
                "md4"
            ]
        },
        "ssh": "OpenSSH_9.7p1, OpenSSL 3.3.0 9 Apr 2024",
        "sshfs": "3.7.3",
        "encfs": "(no encfs)",
        "shell": "/bin/zsh",
        "shell-version": "zsh 5.9 (x86_64-pc-linux-gnu)"
    }
}
```

</details>

### LINKED ISSUE #1751: Global flock active by default (but shouldn't)
**Context**: Bug was introduce by me with PR #1697 providing a new GlobalFlock context manager.

**Problem**: The check of the config value `global.use_flock` is gone. So all user will have an activate global flock no matter of the configuration.

Before PR #1697 that global flock was disabled by default config. That config value was queried in a method named `flockExclusive()`.
https://github.com/bit-team/backintime/blob/4d7e920afe18e767a8999cb0752b972b86d5772e/common/snapshots.py#L784

**Solution**:
My proposed solution is to add a new argument to `GlobalFlock()` context manager to disable it based on the config value `global.use_flock`.
That is not elegant but there is a structure problem with the code in snapshots.py using that context manager. The code block handled by the context manager has hundred of lines of code. Because of that I can not do something like this:

```
if cfg.globalFlock():
    with flock.GlobalFlock():
        do_the_magic()
else:
    do_the_magic()
```

I'll combine the solution together with Issue #1743

buhtz:
The global flock context manager (introduced in #1679) now fall back to single-user mode if not enough permissions available to create an flock file accessible by other users and root. The multi-user mode will be activate later until BIT is started as root for the first time. It will create an flock file accessible by all users. That behavior is relevant for Arch-based systems but not relevant on Debian-based systems because regular users have enough permissions to /run/lock. See the docstring for technical details.

It also fix a bug (introduced with #1679) ignoring the config value of `global.use_flock`.

Additionally I modified and fixed the Sphinx config to make all necessary information about this behavior available in the generated HTML documentation.

Fix #1743 
Fix #1751


common/flock.py
```diff
@@ -10,12 +10,16 @@
 from pathlib import Path
 import logger
 
+logger.DEBUG = True  # REMOVE BEFORE MERGE
```
buhtz: TODO
Need this for manual testing.

willemw12:
How to test this? Is see "INFO: Lock" and "INFO: Unlock". Is that it?

buhtz:
Hello Willem,
thanks for the reply and your efforts to test it. Not having an exception/crash is nice to hear. 😄 

> How to test this? Is see "INFO: Lock" and "INFO: Unlock". Is that it?

You should make sure to activate _Run only one snapshot at a time_ in the Options tab of the Manage profiles (or Profile Settings) dialog? This need to be done for your users BIT profile and for your roots BIT profile.

Run BIT in terminal with `--debug` option. If BIT doesn't crash, great! 🤣 On Arch running as user, you should find a message like this in the terminal output.

```
DEBUG: [common/flock.py:67 GlobalFlock._log] Set flock /run/user/1000/backintime.flock by PID 1234.
...
DEBUG: [common/flock.py:67 GlobalFlock._log] Release flock /run/user/1000/backintime.flock by PID 1234.
```

If running as root the path is different.

```
DEBUG: [common/flock.py:67 GlobalFlock._log] Set flock /run/lock/backintime.flock by PID 1234.
...
DEBUG: [common/flock.py:67 GlobalFlock._log] Release flock /run/lock/backintime.flock by PID 1234.
```

When running again as user, *after running it as root*, the path should also be /run/lock.

willemw12:
Checked it on Arch in Distrobox, as a non-root user. Also does not crash. (Running as root would require more setup.)

buhtz:
I see no strong need for an additional review. Despite myself, two users tested the solution and reported no problems. Because of the time resources I hope it is OK for others if I merge this in the upcoming weekend.

Issue #1760, opened 2024-06-08
Individual files showing in "Backup Folders" section of places section of GUI.


fallingrock:
Individual files showing in "Backup Folders" section of places section of GUI with the 'file list' showing  'This folder doesn't exist in the current selected snapshot'.

This appears to be related to a change made by @rafaelhdr for #1585 

![image](https://github.com/bit-team/backintime/assets/4029821/622bd83d-311e-477f-a5ea-352c84f6c499)

```
$ backintime --diagnostics
WARNING: Failed to connect to Udev serviceHelper daemon via D-Bus: org.freedesktop.DBus.Error.Spawn.ChildExited
WARNING: D-Bus message: Launch helper exited with unknown return code 1
WARNING: Udev-based profiles cannot be changed or checked due to Udev serviceHelper connection failure
{
    "backintime": {
        "name": "Back In Time",
        "version": "1.4.4-dev.ba1ff787",
        "running-as-root": false,
        "latest-config-version": 6,
        "local-config-file": "/home/UsernameReplaced/.config/backintime/config",
        "local-config-file-found": true,
        "global-config-file": "/etc/backintime/config",
        "global-config-file-found": false,
        "started-from": "/usr/share/backintime/common",
        "user-callback": "/home/UsernameReplaced/.config/backintime/user-callback",
        "keyring-supported": true
    },
    "host-setup": {
        "OS": {
            "/etc/os-release": "Ubuntu 22.04.4 LTS",
            "/etc/lsb-release": "DISTRIB_ID=Ubuntu\nDISTRIB_RELEASE=22.04\nDISTRIB_CODENAME=jammy\nDISTRIB_DESCRIPTION=\"Ubuntu 22.04.4 LTS\"\n",
            "/etc/debian_version": "bookworm/sid\n"
        },
        "platform": "Linux-6.5.0-35-generic-x86_64-with-glibc2.35",
        "system": "Linux #35~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Tue May  7 09:00:52 UTC 2",
        "display-system": "x11",
        "locale": "en_US, UTF-8",
        "PATH": "/home/UsernameReplaced/.local/bin:/home/UsernameReplaced/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin:/snap/bin",
        "RSYNC_OLD_ARGS": "(not set)",
[...]

PR #1762, opened 2024-06-10
Only show directories in places window #1760

fallingrock:
Only show directories in places window #1760


CHANGES
```diff
@@ -1,6 +1,7 @@
 Back In Time
 
 Version 1.4.4-dev (development of upcoming release)
+* Fix: Only show directories in places window (#1760)
```
buhtz:
```suggestion
```
The bug you do fix here was never released. So there is not need to mention it in the Changelog.

CHANGES
```diff
@@ -921,4 +922,4 @@ Version 0.5.1 (2008-10-10)
 * changed some texts
 
 Version 0.5 (2008-10-03)
-* This is the first release.
+* This is the first release.
```
buhtz:
```suggestion
* This is the first release.

```

qt/app.py
```diff
@@ -2049,4 +2049,4 @@ def debugTrace():
     cfg.PLUGIN_MANAGER.appExit()
     appInstance.exitApplication()
 
-    logger.closelog()  # must be last line (log until BiT "dies" ;-)
+    logger.closelog()  # must be last line (log until BiT "dies" ;-)
```
buhtz:
```suggestion
    logger.closelog()  # must be last line (log until BiT "dies" ;-)

```

buhtz (changes requested):
Hello David,
On behalf of the team, thank you for your contribution and taking
the time to improve Back In Time. We appreciate it.

Your solution is quit obvious. Haven't seen this my self. :D

I added suggestions about minor modifications. We will [merge it after a creative break](https://github.com/bit-team/backintime/blob/dev/CONTRIBUTING.md#what-happens-after-you-opened-a-pull-request-PR).

Best,
Christian

qt/app.py
buhtz: See https://pylint.pycqa.org/en/latest/user_guide/messages/convention/missing-final-newline.html

fallingrock: Intellij is adding that.  I'll adjust my settings for this project.

buhtz:
Thanks a lot.
I canceled the TravisCI builds on purpose because to save credits.

buhtz (approved):

PR #1783, opened 2024-07-06
Fix/1755 Enable lint checks that already pass successfully

feman323:
There are several lint checks disabled, even though they have passed successfully without requiring any changes


buhtz:
Thank you for this.
Did you also tried to activate that rules on `qt/test/test_lint.py`?

feman323:
I just checked. All those rules are already active in qt/test/test_lint.py

buhtz (approved):

PR #1787, opened 2024-07-09
fix(build): Remove debian/copyright from configure script (#1548)

buhtz:
Another issue related to debian-folder removal (#1548). ~I will investigate why Travis didn't catch it.~ Travis didn't catch it because `make install` is not run there. I tried it but ran into several problems that might be related to the docker-like environment running on Travis. But I am not sure.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -61,13 +61,13 @@ script:
   - cd common
   - ./configure  --python=python3
   - make unittest-v
-  - make install
+  - sudo make install
   - cd ..
   - cd qt
   - ./configure --python=python3
   - make
   - pytest
-  - make install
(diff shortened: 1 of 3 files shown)


SimonCao1207:
Hi, I've just set up the installation and got stuck at "sudo make install" due to "install --mode=644 ../debian/copyright $(DEST)/share/doc/backintime-common". Should this line be removed in Makefile ? Thanks

buhtz:
> Should this line be removed in Makefile ?

This is exactly what his Pull Request here does. Please click on "Files changed" to see the modifications this Pull Request will bring into the repository. The modifications are not yet merged but will be soon.

PR #1789, opened 2024-07-10
Enable lint checks W0123 and W0707

feman323:


common/mount.py
```diff
@@ -604,11 +604,11 @@ def _umount(self):
         try:
             subprocess.check_call(['fusermount', '-u', self.currentMountpoint])
 
-        except subprocess.CalledProcessError:
+        except subprocess.CalledProcessError as e:
```
buhtz: One letter variable names will violate other rules.
I would suggest `err` or `exc`.

buhtz (changes requested):
Thank you for this PR. I just request some minor modifications.

common/mount.py
feman323: Makes sense. I renamed it in mount.py and in tools.py

buhtz (approved):

Issue #1755, opened 2024-06-06
Enable PyLint rules of your own choice and fix the errors


buhtz:
# Introduction
Welcome to the project, if you pick up this Issue because of the "GOOD FIRST USE" label. You will be mentored through the process if you want. First the Issue is explained. At the end you will find some guidance for first contributors. Please do not hesitate to ask questions. Your solution don't need to be perfect.

Be aware that this is a **meta** Issue not describing one specific problem and task, but a bunch of similar problems you can choose from.

It is a fine issue for beginners. But participation or contribution is not recommended unless you are familiar with Back In Time and are an active user. This project is not intended for learning basic Python skills or enhancing your GitHub profile.

# Problem
Code analyzing tools like _PyLint_ and others show multiple of 1.000 errors (~4.500 in default configuration) on the current code base of _Back In Time_. They can not be fixed all at once. Back In Time currently use `pylint` in two unit test (see `common/test/test_lint.py` and `qt/test/test_lint.py`) with a very small selection of enabled error codes.

In the code of the two files [`common/test/test_lint.py`](https://github.com/bit-team/backintime/blob/dev/common/test/test_lint.py) and [`qt/test/test_lint.py`](https://github.com/bit-team/backintime/blob/dev/qt/test/test_lint.py) you will find a list of disabled error codes you can start with:

```
# ...in common/test/test_lint.py
# 'W0612',  # unused-variable
# 'W0237',  # arguments-renamed
# 'W0221',  # arguments-differ
# 'W0603',  # global-statement
# ...in qt/test/test_lint.py
# 'R0201',  # no-self-use
# 'R0202',  # no-classmethod-decorator
# 'R0203',  # no-staticmethod-decorator
# 'W0123',  # eval-used
# 'W0237',  # arguments-renamed
# 'W0221',  # arguments-differ
# 'W0404',  # reimported
# 'W4902',  # deprecated-method
# 'W4904',  # deprecated-class
# 'W0603',  # global-statement
# 'W0614',  # unused-wildcard-import
# 'W0611',  # unused-import
# 'W0612',  # unused-variable
# 'W0707',  # raise-missing-from
```

**Be aware** that there are two files named `test_lint.py`, one in folder `common/test` and one in `qt/test`. They should be identical, especially the list of enabled/disabled error codes.

# Solution
[...]

PR #1790, opened 2024-07-11
Worked on fixing issue W0612 (unused-variable) in "common"

dnerever:
Fixed the occurrence of unused variables through multiple files within `/common`

Related to #1755


common/config.py
```diff
@@ -1603,11 +1603,10 @@ def olderThan(self, time, value, unit):
...
             firstDay = now.date() - datetime.timedelta(days = now.date().day + 1)
-            for i in range(value - 1):
-                if firstDay.month == 1:
-                    firstDay = firstDay.replace(month = 12, year = firstDay.year - 1)
-                else:
-                    firstDay = firstDay.replace(month = firstDay.month - 1)
+            if firstDay.month == 1:
+                firstDay = firstDay.replace(month = 12, year = firstDay.year - 1)
+            else:
+                firstDay = firstDay.replace(month = firstDay.month - 1)
```
buhtz: What is the rational behind this? The for loop is gone.

I assume the iteration is needed but the iteration counter not. In this case someone should write

    for _ in range(values-1):
        foobar()

The `_` is an accepted variable name in the Python world for "unnamed" variables.

But this method seems to be very important. We could break a lot if something is wrong here.
I will have a look if there is a good unit test for this method and if not I would write one. Or do you want to write a unit test?

common/schedule.py
```diff
@@ -164,7 +164,7 @@ def is_cron_running():
 
     with subprocess.Popen(['ps', '-eo', 'comm'], stdout=subprocess.PIPE) as ps:
         try:
-            grep = subprocess.run(
```
buhtz, line 178: Great. This is my own code introduced not long ago. ;)

Absolutely right this modificiation.

common/sshtools.py
```diff
@@ -568,17 +568,16 @@ def benchmarkCipher(self, size=40):
 
             print('%s%s:%s' % (bcolors.BOLD, cipher, bcolors.ENDC))
 
-            for i in range(2):
```
buhtz: The whole method is unclear to me what its purpose is.
This for-loop makes no sense to me. But I assume there is one.

If you don't have an idea I would recommend to reset that modification and we remember that in a new issue.

common/sshtools.py
```diff
@@ -835,7 +834,7 @@ def maxArg():
                                         stdout=subprocess.PIPE,
                                         stderr=subprocess.PIPE,
                                         universal_newlines=True)
-                out, err = proc.communicate()
```
buhtz, line 843: You can keep that modification. That is ok.

buhtz:
Hello Keith,
Thank you for that PR. You see it can get complicated. I added some comments. Let me know how you want to proceed.

Because of its complexity and possible impact I schedule that PR until the upcoming release (see milestone).

common/config.py
buhtz: I introduced #1791 . So just reset this modification and leave it as it is. We will take care of it another time.

dnerever:
Hey Christian,
Can you direct me to where you added your comments. 

I do see how it can get complicated so delaying it makes sense. Any suggestions for ensuring my future PRs avoid introducing complexity? I'm going now to take a look at the other opens issues in the repo.

Thanks again for all the help!

buhtz:
You are not the one making it complicated. The old and smelly Back In Time code does. I just tried to pointing out that even simple modifications can become complicated in old code base like Back In Time has.

About the comments. I added 4 comments directly to the diff-like output. This means a comment is connected directly to a line in the code. I marked them with yellow circles in this screenshot. Just scroll up.

image

buhtz (approved):

buhtz:
To prevent you from extra work and wasting time, I just want to mention that someone else will work on the same rule but in "qt" folder.

https://github.com/bit-team/backintime/issues/1755#issuecomment-2271453055

EDIT: By the way I realized that you did not activate the W0612 rule in `common/test/test_lint.py` in this PR. Is there are reason? If all W0612 errors in "common" are fixed you can activate it.

buhtz (changes requested):
Please see my last comment.

buhtz:
I see why you didn't enable W0612 by default. There are to many errors.

buhtz (approved):

PR #1797, opened 2024-07-15
Update FAQ.md: Segfault-on-Exit bug fixed

emtiu:
Updating the FAQ to reflect the Segfault-on-Exit error being (finally!) fixed :)


buhtz (approved):

buhtz:
By the way: It might be more "secure" if you would work on your own forked repository? :thinking:  In my case it would be, because I often screw up things (with git). :rofl:

emtiu:
Yes, you're right – then there will be a smaller chance that I accidentally commit something to the main repo. Thanks for the advice :)

PR #1816, opened 2024-07-27
build!: Minimum Python version 3.9 required

### LINKED ISSUE #1731: Drop Python3.8 support this year
Regarding to [python.org](https://www.python.org/downloads/) the Python version 3.8 will lose its "security" support phase in October this year.

Because of that we should drop the support for Python 3.8 after the upcoming release.

buhtz:
Close #1731


daveTheOldCoder:
What's the reason for this change? I'm asking because my python3 version is Python 3.8.10.

Does the latest version of BiT not work with Python 3.8?

I suppose I could upgrade it, but I would need to find a PPA or use a manual install.

Issue #781, opened 2017-06-18
backintime root crontab doesn't run (missinng line-feed 0x0A on last line bug)


dtschmitz:
Thank you for all the great work you do.

I have reinstalled Arch and cronie with backintime, enabled cronie and started the service from systemctl, status:

```
[root@ThinkPad backintime]# systemctl status cronie
● cronie.service - Periodic Command Scheduler
   Loaded: loaded (/usr/lib/systemd/system/cronie.service; enabled; vendor prese
   Active: active (running) since Sun 2017-06-18 10:15:42 EDT; 10min ago
 Main PID: 5091 (crond)
    Tasks: 1 (limit: 4915)
   CGroup: /system.slice/cronie.service
           └─5091 /usr/bin/crond -n

Jun 18 10:15:42 ThinkPad systemd[1]: Started Periodic Command Scheduler.
Jun 18 10:15:42 ThinkPad crond[5091]: (CRON) INFO (Syslog will be used instead o
Jun 18 10:15:42 ThinkPad crond[509[root@ThinkPad backintime]# crontab -l
  30 8 *   *   *   /usr/bin/nice -n19 /usr/bin/ionice -c2 -n7 /usr/bin/backintime backup-job > /root/.local/share/backintime/`echo $(date +"%Y%m%d%H%M").cron.log` 2>&1
1]: (CRON) INFO (RANDOM_DELAY will be scaled w
Jun 18 10:15:42 ThinkPad crond[5091]: (CRON) INFO (running with inotify support)
Jun 18 10:15:42 ThinkPad crond[5091]: (CRON) INFO (@reboot jobs will be run at c
Jun 18 10:18:01 ThinkPad crond[5091]: (root) RELOAD (/var/spool/cron/root)

```

The root crontab:
```
[root@ThinkPad backintime]# crontab -l
  30 8 *   *   *   /usr/bin/nice -n19 /usr/bin/ionice -c2 -n7 /usr/bin/backintime backup-job > /root/.local/share/backintime/`echo $(date +"%Y%m%d%H%M").cron.log` 2>&1
```

If I try to run with a specific time to test (above last test @ 8:30am) the cron doesn't run.
Yet, if I copy paste the command to the root command line, it does run.  
I have no issue with backintime -- has always run like a Swiss Watch.

What am I missing?  Any help would be appreciated.
Kindest Regards,
Dietrich

P.S. I am running  Antergos (Arch Linux) rolling release

Germar:
[...]

PR #1827, opened 2024-08-04
[1.5.2] fix: Crontab end with newline

buhtz:
Make sure crontab is always written with a newline. Some cron implementations (e.g. cronie) would give an error otherwise.
Was fixed in #781 and reintroduced with 6d77e319cd9c15f54dd60acc7879b938535d317c (refactoring).
Additionally fixing behavior of accidentally adding empty lines to crontab.

I verified this does not influence regular vixie cron used by Debian systems.


common/schedule.py
```diff
@@ -78,6 +85,10 @@ def write_crontab(lines):
     """
     content = '\n'.join(lines)
 
+    # Crontab need to end with a newline
```
emtiu:
```suggestion
    # Crontab needs to end with a newline
```

emtiu (approved):

PR #1835, opened 2024-08-08
Fix: Correct global flock fallbacks (#1834)

### LINKED ISSUE #1834: Automated backups failing due to KeyException in flock.py
Back In Time has failed to run scheduled backups on my desktop since I updated to 1.5.0 due to changes in `common/flock.py`. Manually started backups run correctly, including those started with the same command that Back In Time uses in its crontab entry, despite other crobtab entries running successfully, which indicates an environmental variable issue. The issue occurs when trying to create a usable lock file:
- `/run/lock/` is owned by root and has the permissions `drwxr-xr-x`, and a lock file cannot be created here. While I see that other systems allow group/world to write here, my attempt to update the permissions on this directory are being reverted by a package I have not identified.
- `XDG_RUNTIME_DIR` is set to `/run/user/1000` when run in an interactive terminal, but is unset when run by my cron package (`cronie`). This is accessed without a fallback or error handling (`os.environ['XDG_RUNTIME_DIR']`), which causes an unhandled `KeyException` that crashes Back In Time without further log messages.
- Had Back In Time continued past this point, `XDG_CACHE_HOME` is not set in cron nor in an interactive terminal, and the fallback is incorrectly set to `~/cache` instead of `~/.cache` despite comments suggesting otherwise, and will not find a valid folder.

Here are my `journalctl` logs for BackInTime for an automatic backup through cron, followed by a manually triggered backup for comparison:
https://0x0.st/XOCh.txt
https://0x0.st/XVrZ.txt

I have tested a fix for these issues, along with a suggestion in the same file from pylint, and will be submitting a pull request shortly.

I installed Back In Time through the AUR package at https://aur.archlinux.org/pkgbase/backintime

```
[tim@Tim-Desktop-5 ~]$ backintime --diagnostics
{
    "backintime": {
        "name": "Back In Time",
        "version": "1.5.2",
        "running-as-root": false,
        "latest-config-version": 6,
        "keyring-supported": false
    },
    "host-setup": {
        "OS": {
            "/etc/os-release": "Arch Linux",
            "/etc/arch-release": "\n",
            "/etc/lsb-release": "DISTRIB_ID=\"Arch\"\nDISTRIB_RELEASE=\"rolling\"\nDISTRIB_DESCRIPTION=\"Arch Linux\"\n"
        },
        "platform": "Linux-6.10.3-zen1-2-zen-x86_64-with-glibc2.40",
        "system": "Linux #1 ZEN SMP PREEMPT_DYNAMIC Tue, 06 Aug 2024 07:47:21 +0000",
        "display-system": "x11",
        "locale": "en_US, UTF-8",
        "PATH": "/home/UsernameReplaced/bin:/home/UsernameReplaced/.local/bin:/home/UsernameReplaced/bin:/home/UsernameReplaced/.local/bin:/usr/local/bin:/usr/bin:/bin:/usr/local/sbin:/opt/android-sdk/emulator:/opt/android-sdk/cmdline-tools/latest/bin:/opt/android-sdk/platform-tools:/opt/android-sdk/tools:/opt/android-sdk/tools/bin:/home/UsernameReplaced/.local/share/flatpak/exports/bin:/var/lib/flatpak/exports/bin:/usr/lib/jvm/default/bin:/usr/bin/site_perl:/usr/bin/vendor_perl:/usr/bin/core_perl:/opt/rocm/bin:/var/lib/snapd/snap/bin",
    },
    "python-setup": {
        "python": "3.12.4 main Jun  7 2024 06:33:07 CPython GCC 14.1.1 20240522",
        "python-executable": "/usr/bin/python",
        "sys.path": [
            "/usr/lib/python312.zip",
            "/usr/lib/python3.12",
            "/usr/lib/python3.12/site-packages"
        ],
        "qt": {
            "Version": "PyQt 6.7.1 / Qt 6.7.2",
            "Theme": "matefaenzadark",
            "Theme Search Paths": [
                "/home/UsernameReplaced/.local/share/flatpak/exports/share/icons",
                "/usr/share/icons",
                "/var/lib/snapd/desktop/icons",
                ":/icons"
            ],
            "Fallback Theme": "hicolor",
            "Fallback Search Paths": [
                "/usr/share/pixmaps"
            ]
        }
    },
    "external-programs": {
        "rsync": {
            "version": "3.3.0",
            "protocol": "31.0",
            "capabilities": "file_bits: 64; inum_bits: 64; timestamp_bits: 64; long_int_bits: 64; socketpairs: True; symlinks: True; symtimes: True; hardlinks: True; hardlink_specials: True; hardlink_symlinks: True; IPv6: True; atimes: True; batchfiles: True; inplace: True; append: True; ACLs: True; xattrs: True; secluded_args: optional; iconv: True; prealloc: True; stop_at: True; crtimes: False",
            "optimizations": "SIMD_roll: True; asm_roll: False; openssl_crypto: True; asm_MD5: False",
            "checksum_list": "xxh128, xxh3, xxh64, md5, md4, sha1, none",
            "compress_list": "zstd, lz4, zlibx, zlib, none",
            "daemon_auth_list": "sha512, sha256, sha1, md5, md4"
        },
        "ssh": "OpenSSH_9.8p1, OpenSSL 3.3.1 4 Jun 2024",
        "sshfs": "3.7.3",
        "encfs": "(no encfs)",
        "shell": "/bin/bash",
        "shell-version": "GNU bash, version 5.2.32(1)-release (x86_64-pc-linux-gnu)"
    }
}
```

NickNackGus:
- Prevents a crash if `XDG_RUNTIME_DIR` is not set
- Default to `~/.cache` instead of `~/cache`
- return None explicitly instead of implicitly (pylint R1710 inconsistent return statements)

Fix #1834


buhtz:
Hello Timothy,
On behalf of the team, a big thank you for reporting the bug, investigating it and also contribute fix. Thank you for taking the time to improve Back In Time. We appreciate it.

Your PR looks awesome on a first view. I will review it as soon as possible. I will do some more testing. I need to investigate if this affects Debian, too. I am assuming not but if so it would block the current version 1.5.2 arriving in Debian and the upcoming Ubuntu release (Debian Import Freeze is in some days on 15th August).

Best,
Christian

buhtz:
I can confirm that this will fix #1834

CHANGES
```diff
@@ -1,5 +1,6 @@
 Back In Time
 Version 1.5.3-dev (development of upcoming release)
+* Fix: Correct global flock fallbacks (#1834)
```
buhtz: You can also put your name as author behind that line. Using nickname or full name is up to you. Like this:

    * Fix: Correct global flock fallbacks (#1834) (Timothy Southwick)

buhtz (changes requested):
Can you please fix the other "return" statements in __init__() and make them return None explicit. Thank you.

NickNackGus:
Done - though `None` isn't strictly required in all cases. pylint was only complaining about it where a method returned non-`None` in some cases, and didn't specify a return value in others.

buhtz (approved):

buhtz:
Currently all PRs are blocked until Ubunuts "Debian Import Freeze" on 15th August.

Until then maybe one of my team mates can also have a look on this PR.

common/flock.py
```diff
@@ -84,23 +84,26 @@ def __init__(self,
...
-            return
+            return None
 
         # Try user specific file lock
         # e.g. /run/user/<UID>
-        self._file_path = Path(os.environ['XDG_RUNTIME_DIR']) / filename
+        self._file_path = Path(
+            os.environ.get('XDG_RUNTIME_DIR',
+                           f'/run/user/{os.getuid()}')
+        ) / filename
```
emtiu, line 94: I assume that changing `return` into `return None` is just a question of style, and has no functional effect?

Otherwise, this looks sensible, _**if**_ we can assume that `/run/user/{os.getuid()}` is available on every distro these days.

NickNackGus: `return None` is the suggested alternative to `return` where a method has a return statement with a return value specified, both in my testing and the documentation for the pylint warning [here](https://pylint.pycqa.org/en/latest/user_guide/messages/refactor/inconsistent-return-statements.html). A quick test shows that methods that either don't specify return values or that do not use a return statement return `None` by default.

According to [the library reference](https://docs.python.org/3/library/os.html#os.getuid), `os.getuid()` is available in all Python versions on Unix-based systems, but is a stub on Emscripten and WASI.

NickNackGus:
One thing this PR doesn't address - if there's an exception around here in the code, there's no further information in the logs to help debug. I'm not sure how best to catch and log such issues so that they're easier to debug in the future. Putting a try/catch for anything that inherits from `Exception` would work, but also add a warning in pylint for catching too general an exception, and would need to be after the logging code was loaded at any rate.

common/flock.py
buhtz: Yes __init__() always return None. Here we just make it explicit instead of implicit. Python Zen rule #2.

About `/run/user/{os.getuid()}`: Ever folder that is used as flock is tested first (see `_can_use_file()`). Otherwise the next alternative folder will be tried and in the end an exception should be raised.

buhtz:
I don't see a need for more Exception handling. The method `_can_use_file()` does most of it and also catch a broad `Exception` in the end.
Catch exception you do expect. Don't catch everything. It was right that my code crashed because it was crap and not robust enough.

BIT has a lot edge cases I often don't have in mind. Because of that PR I [extended the release process docu about what to consider when doing manual tests](https://github.com/buhtz/backintime/blob/doc/installtesting/common/doc-dev/BiT_release_process.md#manual-testing---recommendations).

PR #1840, opened 2024-08-13
chore: Fix appstream meta data & clarify license as "GPL-2.0-or-later"

buhtz:
The v1.5.2 release on Debian brought up some AppStream related warnings and errors. Nothing serious.

- Fixed a missing screenshot. Accidentally removed it from weblate before.
- Reduced length of "summary".
- Add "Back In Time Team" as "developers".
- More precise license info: Replaced `GPL-2.0` with `GPL-2.0-or-later`. Have done some research about that topic. The v2 text does not distinguish between "-only" or "-or-later". That need to be specified elsewhere in a project. If not specified "-or-later" can be assumed by default. To my knowledge there is no clear rule or statement about that. I am also not aware of any court ruling clarifying that situation. I was not able to establish contact to the former BIT founder Oprea Dan to clearify that. @Germar do you have an opinion or information on that topic and can clarify the licence to use for your parts of the code?

EDIT: Asked myself why I didn't get this errors earlier because I was using the appstrim linter on my system. But it seems the linter on Debian 12 is quit old compared to the one used in Debians build system.

### DIFF
--- a/qt/io.github.bit_team.back_in_time.gui.metainfo.xml
+++ b/qt/io.github.bit_team.back_in_time.gui.metainfo.xml
@@ -6,15 +6,17 @@
     Validate (lint) on shell:
         appstreamcli validate *xml
 -->
-
-
 <component type="desktop-application">
   <id>io.github.bit_team.back_in_time.gui</id>
   <name>Back In Time</name>
-  <summary>Backup tool for GNU Linux desktop using rsync with its hard-links feature to save storage space</summary>
+  <!--
+      Length limit for "summary" field is unclear but could be 90.
+      See https://github.com/ximion/appstream/issues/648
+  -->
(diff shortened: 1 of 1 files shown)


daviewales:
I am neither a lawyer, nor an expert on GPL licenses. But my feeling would be that because the original licenses state 'GPL-2.0', not 'GPL-2.0 or later', that the project is 'GPL-2.0 only'. I tend to agree with your thinking in another PR comment: https://github.com/bit-team/backintime/pull/1841#issuecomment-2299134487

buhtz:
> my feeling would be that because the original licenses state 'GPL-2.0', not 'GPL-2.0 or later', that the project is 'GPL-2.0 only'.

I updated [my comment](https://github.com/bit-team/backintime/pull/1841#issuecomment-2299134487) about that topic.

PR #1841, opened 2024-08-13
refactor: Daemon class clarify its license

buhtz:
Refactoring of the Daemon class in tools.py and separating it into its own file.

The licence situation was not clear but I contacted the original author. Details in the module docstring. The code was taken from a blog post of 2007 licence as CC BY-SA-3.0 which is not compatible with GPLv2. But the author clarified to me that the license was targeting the blog post and not the code in it.

### DIFF
--- a/common/tools.py
+++ b/common/tools.py
@@ -1,6 +1,8 @@
 # SPDX-FileCopyrightText: © 2008-2022 Oprea Dan
 # SPDX-FileCopyrightText: © 2008-2022 Bart de Koning
 # SPDX-FileCopyrightText: © 2008-2022 Richard Bailey
+# SPDX-FileCopyrightText: © 2008-2022 Germar Reitze
+# SPDX-FileCopyrightText: © 2008-2022 Taylor Raack
 #
 # SPDX-License-Identifier: GPL-2.0-or-later
 #
--- a/common/tools.py
+++ b/common/tools.py
@@ -22,10 +22,8 @@ import locale
 import gettext
 import hashlib
 import ipaddress
-import atexit
 from datetime import datetime
 from packaging.version import Version
-from time import sleep
 import logger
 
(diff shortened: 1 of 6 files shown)


buhtz:
I am still insecure about how to handle interpret the GPLv2 and its suffixes. See section 9 of the original license text.
My insecurity might depend on my non-native English. In concurrence in what I said before now it seems to me that "GPLv2" is by default "-only". In case of BIT where "-or-later" never was mentioned explicit somewhere we might interpret it is "-only".

https://github.com/bit-team/backintime/blob/366ef003d7049ae460d43386f920285453d0ad9d/LICENSE#L237-L248

Any thoughts @bit-team ?

**EDIT**: As stated per email I changed my interpretation. The original GPLv2 license text has no "default suffix". This is supported by some comments I got in several forums and also be the [REUSE.software](https://reuse.software) tool `reuse`. Download booth licence versions via `$ reuse download GPL-2.0-only` and ``$ reuse download GPL-2.0-or-later` and then just `diff` the text. There is no difference.
The difference is made by the licence notice in the header of each individual file. And these are quite clear that it is allowed to use them with a later version of GPL, too. So IMHO it is `GPL-2.0-or-later`.

common/tools.py
```diff
@@ -1,22 +1,14 @@
-"""Collection of helper functions not fitting to other modules.
-"""
-# Back In Time
-# Copyright (C) 2008-2022 Oprea Dan, Bart de Koning, Richard Bailey,
-# Germar Reitze, Taylor Raack
```
daviewales, line 5: Should 'Germar Reitze' and 'Taylor Raack' still be credited in the new version of the file?

buhtz: See! That's why I need you! ❤️ 🤣 
Have missed that name because of the line break. Stupid me.

PR #1842, opened 2024-08-14
refactor: "General" tab and its "Schedule" section

buhtz:
Refactoring Manage profile dialog with separating the General tab code into its own method and create a Schedule widget.

EDIT: I use its own file for the schedule widget because it makes the diff easier to read and understand.

EDIT2: The intention of the PR and refactoring the Manage Profiles dialog is not only about just beautify the code. I have some Issues in mind that might be easier to solve when this code is better organised; e.g. the GoCrypt implementation and every issue around the Config class.


qt/schedulewidget.py
```diff
@@ -0,0 +1,289 @@
...
+
+        # Weekday
+        self._combo_weekday = self._weekday_combobox()
+        self._rowidx_weekday = _create_form_entry(
+            _('Weekday:'), self._combo_weekday)
+
+        # Time (formerly known as Hour)
+        self._combo_time = self._time_combobox()
+        self._rowidx_time = _create_form_entry(
+            _('Time:'), self._combo_time)
```
buhtz, line 70: Modified the string `Hour:` into `Time:` because it fits IMHO better to values like `['0:00', '01:00', ..., '23:00']`.

Issue #1674, opened 2024-03-31
Synology DSM: rsync/ssh is working manually, but backintime refuses to use the same path


Miq1:
I am using Ubuntu 22.04 LTS with backintime 1.2.1, so there is no ``--diagnostics`` yet. There seems to be no backport of a more recent backintime for Jammy.

I am trying to use backintime to back up to a remote folder on my NAS (Synology DS216j with version 7.1.1). While that is working with the folder mounted locally, but with issues like excess copying of files that are not changed, I cannot get the rsync/ssh variety to work.

I can manually rsync with ssh to the folder ``/volume1/NimbusBackup/`` from the console, but setting up a backintime profile ends with this error: 
```
sshfs -o ServerAliveInterval=240 -o LogLevel=Error -o IdentityFile=/home/micha/.ssh/id_rsa -p 22 -o idmap=user -o cache_dir_timeout=2 -o cache_stat_timeout=2 micha@192.168.178.21:/volume1/NimbusBackup/ /home/micha/.local/share/backintime/mnt/D77AA3BC/mountpoint kann nicht eingebunden werden

micha@192.168.178.21:/volume1/NimbusBackup/: No such file or directory

```
Since I am using the exact same path when manually using rsync, I have no clue what may be wrong.

buhtz:
Hello Michael,
Thanks for your report.

I don't understand your problem description. Can you be a bit more specific about it please.

"Synology" is a good indicator. Please check this two FAQ entries:
- [How to use Synology DSM 5 with BIT over SSH](https://github.com/bit-team/backintime/blob/dev/FAQ.md#how-to-use-synology-dsm-5-with-bit-over-ssh)
- [How to use Synology DSM 6 with BIT over SSH](https://github.com/bit-team/backintime/blob/dev/FAQ.md#how-to-use-synology-dsm-6-with-bit-over-ssh)

Best,
Christian

Miq1:
Sorry for being unclear - I am German and my English may not be as good as I would hope for... Second I have not used Linux for many years and only recently converted from Windows to Ubuntu, so I am lacking knowledge about the nooks and crannies of course.

I sort of followed the FAQ, hence I was able to set up the Synology NAS to accept rsync requests over ssh for my user without requiring a password any more.

I proved it was working by ``rsync -vvv -e ssh micha@192.168.178.21:/volume1/NimbusBackup/``, which gave me
```
[...]

PR #1843, opened 2024-08-18
doc: Add Synology DSM version 7 to the FAQs

Silver-Saucepan:
Add instructions on how to configure Synology devices running DSM7 for use with SSH-mode BIT profiles


aryoda:
@Silver-Saucepan Thanks a lot for your contribution!

Your PR is related to  #1674 (where I wanted to write a FAQ entry too) and does save me a lot of time now :-)

Would it be possible to check if Synology's Audio Station shows a wrong "used disk" size for the backup folder (since it seems to follow the links too instead of ignoring them = effectivly overcounting the used disk size...)

Silver-Saucepan:
@aryoda Yes, when Synology's File Station calculates the size of a folder it does not seem to be inode-number-aware, so each hard link within the folder is treated as a separate file, even if it links to a file who's size it has already counted, and so its size is added to the total. Thus the calculated size of a BIT snapshots folder can be many times its actual space usage on disk.

You might know this already but, the GNU ```du``` command does seem to be inode-number-aware and so can be used to get the actual space usage, counting additional hard links to a file it has already encountered as zero size.  e.g. if the snapshots are in a folder called ```1``` then ```du -hs 1/``` shows the actual space used.  However, if you use ```du``` on an individual snapshot folder (e.g. ```du -hs 1/20240819-061501-243``` it includes the size of 'files' that are actually hard-links to files in other snapshots.

This should probably be described in the DSM 7 section of the FAQs, and DSM 6 probably has the same behaviour but I can't test it to be sure.

buhtz:
Would it be OK if I merge this?

Silver-Saucepan:
It's OK by me. I tested it on a NAS running DSM 7.2.1

buhtz:
> It's OK by me. I tested it on a NAS running DSM 7.2.1

Thank you very much. This helps a lot.

Issue #1516, opened 2023-08-31
After importing a previous config from a snapshot, the snapshots list is populated with duplicates


emtiu:
When testing for #1508, I noticed a bug in the current dev code that I had previously encountered in the wild.

After importing a previous config from a snapshot, the snapshots list is populated with duplicates. They disappear after the GUI has been restarted once.

Steps to reproduce:
1. Start the GUI with no configuration.
2. Choose "Yes" to _restore a previous configuration_.
3. Select a snapshot and click "Restore".
4. In the "Settings" dialog that opens now, click "OK" without making changes.
5. Observe how **every snapshot is listed twice:**

![Screenshot_20230831_120907](https://github.com/bit-team/backintime/assets/2159933/b84762fd-3e46-42d9-9560-83b5167fe0bd)

After quitting and restarting the GUI once, the snapshots are listed as expected, without duplicates:

![Screenshot_20230831_120837](https://github.com/bit-team/backintime/assets/2159933/f560f114-aeb7-4f83-9750-dd13377c493f)

I remember that this bug was quite a heart-stopper when I encountered it in the wild. I thought there was something wrong with the snapshots.

buhtz:
I played around with debug code and `inspect` module. I followed some trails and ideas but I am still not sure. Qt's signal/slot mechanic seems to be involved somehow and the `FillTimeLineThread` class. But finally I don't get it.

Anyway, it is planned to refactor that code. So I am assuming the problem will be fade away somehow.

buhtz:
The timeline widget was refactored and nearly rewritten in 1.6.1. I didn't experienced the issue myself anymore.

Feel free to reopen the issue.

buhtz:
Issue is still present. Fixed it in PR #1850. Released with version `2.0.0`.

Issue #254, opened 2015-10-11
[enhancement] Text based UI or CLI for config file creation/editing


Germar:
Back In Time is a great piece of software :-)
I've managed to use it also on my embedded NAS, which carries a Debian distribution. I just installed the backintime-common package, because GUIs are not supported in there. Unfortunately, the only very uncomfortable thing is that I had to "prepare" a configuration file on my desktop PC using the KDE interface and then copy it on the NAS. In fact, some configuration options in the config file are not so easy to understand (for example: a unit of "20" may represent days or gigabytes, if I remember well...), so a direct editing of the config file, apart from small tunings, is not an easy task. Also, BIT version differences between environments might make it even more difficult, because you may not know exactly which options are available/supported.

So, I think a good enhancement would be to provide a text-based UI or a new CLI program that allows you to create/edit the configuration file, so that the current backintime executable can then run the defined profiles as usual.

As an alternative, a good documentation of the different configuration options and syntax (which considers BIT version differences) would be valuable.

Germar:
(by germar)
Please take a look at 'man backintime-config'. This explains all config keys. You can find examples in /usr/share/doc/backintime/examples/.

A CLI would be a nice feature and there was already a branch in https://code.launchpad.net/~ryan-c-ahearn/backintime/command_line_interface
But unfortunately this branch was never merged into main dev branch. By now it is completely outdated and would need a lot of work to comply with main branch.

I'm currently working on getting rid of two GUIs (Gnome/Gtk+ and KDE) to make further development easier with only one Qt4 GUI. But if someone would like to overhaul the CLI code we could merge this into 1.1 series.

Germar:
(by mauromol)
Sorry for not knowing about the "man backintime-config", it is certainly useful meanwhile!

emtiu:
When is comes to the question TUI vs. CLI, I think a CLI would have an additional use in scripting. There may be situation where you want to modify your BackInTime configs by script (e.g. to change the include/exclude lists on several profiles at once).

So, ideally we would have both CLI commands and an ncurses-style TUI. But if there should be only one of those, it should be CLI.

buhtz:
In short: A config file in human readable TOML format would vanish the need of having a CLI/TUI to edit the BIT config file.

I setup backintime on a headless server for the first time. So I needed to edit the config file.
I got the idea that editing that file might be more easier if its syntax would be easier to read.
In the Python world today TOML is state of the art. It looks a bit more similar to Windows INI.

[...]

PR #1850, opened 2024-08-24
refactor!: Demonstrator: New config management class (and some more stuff)

### LINKED ISSUE #1923: New configuration managment with "class Konfig" in version 1.6.1
While preparing the 1.5.3 release (PR #1922) I realized that the helper script "create-manpage-backintime-config.py" is broken. I can tell by the diff that some config entries are lost in the man page. The code is hard to read and I don't want to waste time on it anymore.

We do have a pending PR (#1850) offering a new config class including auto generating the man page.

The plan is to migrate to that new class in version ~~1.6.1~~ 2.0 or 2.1.

buhtz:
Blocked by
- #1897
- #1855
- #1864
- #1886 
- #1887 
- #1865 
- #2157
- #2167
- #2156
- #2493

That PR is intended to get merged but also it is more a demonstrator instead of a modification to the current BIT behavior. The code technically is not "connected" to regular BIT and merging it won't change BIT behavior.

The objective was to replace the configuration management logic currently realized in the `Config` class. The current implementation is inefficient and blown up. Additionally that class does a lot more then just managing the configuration.

The PR demonstrate the new solution. Some details are not finished and need to be solved. But I would like to discuss the approach as early as possible before finishing it.

1. See the `__main__` in `common/konfig.py` file for a short example.
1. I choose `Konfig` as name for the new class because of lack of a better alternative. I am open for suggestions. On a long run we will "delete" the whole `Config` class with separating its behavior in several other classes. Once when `Config` is gone we can rename `Konfig` into `Config`.
2. There is only one instance of that class (singleton).
3. Pythons in build [`configparser`](https://docs.python.org/3/library/configparser.html) now is used to read and write the config file. Because that package reads INI-like files I used a little trick because BIT's config file is not an INI file. After reading the raw file into memory I simply add `[BIT]` before the first line and then read it with `configparser`.
4. Profile specific configuration parameters now separated into their own class `Profile` internally dealing with `Konfig`.
5. The creation of the `backintime-config (1)` man page still is done with a script. The script does not parse the konfig.py file anymore but use `inspect` to extract the relevant information (from public properties with docstrings).

How to integrate it into BIT? Not in this PR but later. I would use `Config` in a wrapper-like manner using `Konfig` internally. This is a lot of extra work. But this step would make it smooth and less-risky transforming the codebase into using `Konfig`. And also the needed PRs can be smaller. This way we can carefully decouple the codebase step by step from the `Config` class monolith.

## Side note to myself
`ConfigParser` can't handle lines like this:

    `qt.diff.params=%1 %2`

Problem is solved.

Close #1923

## Changes made
- *Major change*: Replaced `Config` with `Konfig`. `Config` still exists but is using `Konfig` in the back.
- Add `reuse` to linter unit tests
- Add `codespell` to linter unit tests
- Solve duplication of backup timestamps in timeline widget (#1516)
- Increased `ruff` version.
- Satsifiy all linters
- Disabled and skipped several unit tests due to incomplete migration and refactoring (#2461)


common/konfig.py
```diff
@@ -0,0 +1,447 @@
...
+
+class Profile:
+    """Manages access to profile-specific configuration data."""
+    _DEFAULT_VALUES = {
+        'snapshots.mode': 'local',
+        'snapshots.ssh.port': 22,
+        'snapshots.ssh.cipher': 'default',
+        'snapshots.ssh.user': getpass.getuser(),
+        'snapshots.ssh.private_key_file':
+            str(Path('~') / '.ssh' / 'id_rsa'),
```
daviewales, line 46: Is your intention that `~` would be expanded? Or that it remains as literal `~`?

`Path('~')` yields `PosixPath('~')`, in which case you might as well just do:

    'snapshots.ssh.private_key_file': '~/.ssh/id_rsa'

On the other, hand, if you want to expand the home directory to an absolute path, you probably want this instead:

        'snapshots.ssh.private_key_file': str(Path.home() / '.ssh' / 'id_rsa')

I'm curious why you use separate strings for each path component, but I'm guessing that's just a style preference?

common/konfig.py
```diff
@@ -0,0 +1,447 @@
...
+
+    _DEFAULT_VALUES = {
+        'global.hash_collision': 0,
+        'global.language': '',
+        'global.use_flock': False,
+    }
+
+    _DEFAULT_SECTION = '[bit]'
+
+    def __init__(self, config_path: Path = None):
```
daviewales: Should this be:
```suggestion
    def __init__(self, config_path: Optional[Path] = None):
```

common/konfig.py
```diff
@@ -0,0 +1,447 @@
...
+        if not config_path:
+            xdg_config = os.environ.get('XDG_CONFIG_HOME',
+                                        os.environ['HOME'] + '/.config')
```
daviewales: Could also do:
```suggestion
            xdg_config = os.environ.get('XDG_CONFIG_HOME',
                                        Path.home() / '.config')
```

common/konfig.py
```diff
@@ -0,0 +1,447 @@
...
+
+        logger.debug(f'Config path used: {self._path}')
+
+        self.load()
+
+        # Names and IDs of profiles
+        # EXtract all relevant lines of format 'profile*.name=*'
+        name_items = filter(
+            lambda val:
+                val[0].startswith('profile') and val[0].endswith('name'),
```
daviewales: Should it be `'.name'` rather than `'name'` to avoid picking up other keys which happen to end with `'name'`? (Hypothetical: `'profile.basename'`)
```suggestion
                val[0].startswith('profile') and val[0].endswith('.name'),
```

common/konfig.py
```diff
@@ -0,0 +1,447 @@
...
+            self._conf.items()
+        )
+        self._profiles = {
```
daviewales, line 1174: Looking at my own `~/.config/backintime/config`, I see that I have a bunch of `profile1...` items, but none are `profile1*.name`. Is this a new config item? I assume my profile is called `profile1`, but it wouldn't be detected here.

common/konfig.py
```diff
@@ -0,0 +1,447 @@
...
+
+            yield path_or_buffer
+
+            try:
+                # regular file: close it
+                path_or_buffer.close()
+
+            except AttributeError:
+                # in-memory file: "cursor" back to first byte
+                path_or_buffer.seek(0)
```
daviewales: Should this be inside a `finally:` block? I'm wondering what happens if an exception is raised before we start the second `try:` block?
```suggestion
            finally:
                try:
                    # regular file: close it
                    path_or_buffer.close()

                except AttributeError:
                    # in-memory file: "cursor" back to first byte
                    path_or_buffer.seek(0)
```

common/konfig.py
```diff
@@ -0,0 +1,447 @@
...
+
+        raise NotImplementedError('Prevent overwritting real config data.')
+
+        buffer = StringIO()
+        self._config_parser.write(buffer)
+        buffer.seek(0)
+
+        with self._path.open('w', encoding='utf-8') as handle:
+            # Write to file without section header
+            handle.write(''.join(buffer.readlines()[1:]))
```
daviewales: Could you just throw away the unwanted first line by advancing the stream position with `readline`? Then you can just `read()` the rest of the buffer directly, without needing to convert it to a list, then join it back together again.

```suggestion
            # Discard unwanted first line
            buffer.readline()
            handle.write(buffer.read())
```

create-manpage-backintime-config.py
```diff
@@ -46,365 +37,385 @@
...
+
+    content += groff_paragraph_break()
+    content += "Arguments don't need to be quoted. All characters are " \
+               "allowed except '='.\n"
+
+    content += groff_paragraph_break()
+    content += "Run 'backintime check-config' to verify the configfile, " \
+               "create the snapshot folder and crontab entries.\n"
+
+    content += groff_section('POSSIBLE KEYWORDS')
```
daviewales: Could perhaps join these strings together another way?

Either:

```suggestion
    content = (
        f'.TH backintime-config 1 "{stamp}" '
        + f'"version {ver}" "USER COMMANDS"\n'

        + groff_section('NAME')
        + 'config \\- Back In Time configuration file.\n'

        + groff_section('SYNOPSIS')
        + '~/.config/backintime/config\n'
        + groff_linebreak()
        + '/etc/backintime/config\n'

        + groff_section('DESCRIPTION')
        + 'Back In Time was developed as pure GUI program and so most '
          'functions are only usable with '
        + groff_bold('backintime-qt')
        + '. But it is possible to use Back In Time e.g. on a '
          'headless server. You have to create the configuration file '
          '(~/.config/backintime/config) manually. Look inside '
          '/usr/share/doc/backintime\\-common/examples/ for examples.\n'

        + groff_paragraph_break()
        + 'The configuration file has the following format:\n'
        + groff_linebreak()
        + 'keyword=arguments\n'

        + groff_paragraph_break()
        + "Arguments don't need to be quoted. All characters are "
          "allowed except '='.\n"

        + groff_paragraph_break()
        + "Run 'backintime check-config' to verify the configfile, "
          "create the snapshot folder and crontab entries.\n"

        + groff_section('POSSIBLE KEYWORDS')
    )
```

Or even:

```suggestion
    content = ''.join([
        f'.TH backintime-config 1 "{stamp}" '
        f'"version {ver}" "USER COMMANDS"\n',

        groff_section('NAME'),
        'config \\- Back In Time configuration file.\n',

        groff_section('SYNOPSIS'),
        '~/.config/backintime/config\n',
        groff_linebreak(),
        '/etc/backintime/config\n',

        groff_section('DESCRIPTION'),
        'Back In Time was developed as pure GUI program and so most '
        'functions are only usable with ',
        groff_bold('backintime-qt'),
        '. But it is possible to use Back In Time e.g. on a '
        'headless server. You have to create the configuration file '
        '(~/.config/backintime/config) manually. Look inside '
        '/usr/share/doc/backintime\\-common/examples/ for examples.\n',

        groff_paragraph_break(),
        'The configuration file has the following format:\n',
        groff_linebreak(),
        'keyword=arguments\n',

        groff_paragraph_break(),
        "Arguments don't need to be quoted. All characters are "
        "allowed except '='.\n",

        groff_paragraph_break(),
        "Run 'backintime check-config' to verify the configfile, "
        "create the snapshot folder and crontab entries.\n",

        groff_section('POSSIBLE KEYWORDS'),
    ])
```

Note also that you don't need the `\` line continuation.
Python automatically joins consecutive strings.

e.g.:

``` python
example = (
    "This becomes a single "
    "string in Python"
)
```

daviewales (commented):
Hi @buhtz, I've had a look over this.
I don't have any big suggestions, but I commented on a few small things.
There are a few style tweaks (particularly the suggestion for `create-manpage` at the end), so let me know if you'd prefer not to get these kind of style comments.

common/konfig.py
buhtz: Good point. Learned something. Needed to read something about until I understood what `Optional[]` here means. Thanks.

common/konfig.py
buhtz: The intention is not to get expanded. Because the `~` is used in the string written to the config file. That is how a users would do it in some cases. Otherwies I would have written `Path.home()`.

The separate strings (ohne string per path component/folder) is to my knowledge the regular way how the `pathlib` package is intended to be used. This is platform independence. The package decide itself how the folder/path separator is written in the current file system.

buhtz:
> so let me know if you'd prefer not to get these kind of style comments.

Oh no, I love it. 👍

common/konfig.py
buhtz: Profile 1 is always `_('Main profile')`. The name is implicit. We also have an Issue about it to "fix" this behavior.
My current workaround is to simple add that name (as default) while loading the config. See around line 356 in `Konfig.load()`.

common/konfig.py
buhtz: Mhm... I am not sure but see no need for it.
To my understanding of context managers that block you enclosded in "finally" will be called every time no matter what happens. This is because I do use that context manager with `with...` statement. The code after the `yield` line is executed under all circumstance when the `with` block is left.

I might be wrong.

The intention of this creepy context manager is to make it possible to write unit tests using config files from RAM and not from the file system.

common/konfig.py
daviewales: Pathlib is platform independent, but you don't need to separate the path out to make it work. The following will be correctly converted to a Windows path on Windows:

``` python
Path('~/example/path')
```

You can test to see what would happen using `pathlib.PureWindowsPath`:

``` python
p = str(pathlib.PureWindowsPath('~/example/path'))
print(p)
# ~\example\path
```

As you can see, a path containing forward slashes is correctly converted to a valid Windows path automatically.

By using `Path('~/example/path')`, you will automatically get forward slashes on non-Windows, and back-slashes on Windows.

So, given the code above, the following is exactly equivalent:

``` python
'snapshots.ssh.private_key_file': str(Path('~/.ssh/id_rsa')),
```

On Linux, this will become:

```
'snapshots.ssh.private_key_file': '~/.ssh/id_rsa',
```

While on Windows it will become:

```
'snapshots.ssh.private_key_file': r'~\.ssh\id_rsa',
```

I'm not sure how this config is used, but you might even get away with keeping it as a Path object without converting it to string.

common/konfig.py
daviewales: I had never encountered this magic context manager until I saw it here. There was a line in the [docs](https://docs.python.org/3/library/contextlib.html#contextlib.contextmanager) which made me concerned about what would happen in the case of an error:

> If an unhandled exception occurs in the block, it is _reraised inside the generator_ at the point where the yield occurred. Thus, you can use a [try](https://docs.python.org/3/reference/compound_stmts.html#try)…[except](https://docs.python.org/3/reference/compound_stmts.html#except)…[finally](https://docs.python.org/3/reference/compound_stmts.html#finally) statement to trap the error (if any), or ensure that some cleanup takes place.

Here is a test:

``` python
@contextmanager
def manager():
    try:
        print("The game begins...")
    except Error as err:
        print("Oh no, an error!")
        print(err)

    yield "thing"

    try:
        print("The game ends.")
    except Error as err:
        print("Oh no, another error!")
        print(err)

with manager() as m:
    print("Let's break maths:", 1/0)
```

If you run this, you will see that the cleanup code is never run.
It never prints `"The game ends."`:

``` python
The game begins...
---------------------------------------------------------------------------
ZeroDivisionError                         Traceback (most recent call last)
Cell In[52], line 2
      1 with manager() as m:
----> 2     print("Let's break maths:", 1/0)

ZeroDivisionError: division by zero
```

So, I think we need the `finally:` block.

Here's another example which ensures that `yield "thing"` is safely wrapped in a `try: finally:`:

``` python
@contextmanager
def manager():
    try:
        try:
            print("The game begins...")
        except Error as err:
            print("Oh no, an error!")
            print(err)

        yield "thing"

    finally:
        try:
            print("The game ends.")
        except Error as err:
            print("Oh no, another error!")
            print(err)
```

If you run this, you will see that the cleanup runs successfully, and `"The game ends."` is successfully printed before the error is raised:

``` python
The game begins...
The game ends.
---------------------------------------------------------------------------
ZeroDivisionError                         Traceback (most recent call last)
Cell In[54], line 2
      1 with manager() as m:
----> 2     print("Let's break maths:", 1/0)

ZeroDivisionError: division by zero
```

daviewales:
Regarding your side note about configparser not handling lines like the following:

```
`qt.diff.params=%1 %2`
```

How hard would it be to write a custom parser?
I assume there is already an existing custom parser which you don't like?

From a cursory inspection of my config file, it looks like a key-value file separated by the first `=` sign from the left.

So, a naive parser could simply do something like:

``` python
from pathlib import Path
from pprint import pprint

config_file = Path.home() / '.config/backintime/config'
with open(config_file) as file:
    config = {}
    for line in file:
        k, v = line.split('=', 1)
        config[k.strip()] = v.strip()

pprint(config)
```

I have probably missed some subtle details about how the config file works.
But if it's not too much more complicated than this, is it worth just writing the code to process it rather than trying to force it to work with configparser?

On the other hand, if your goal is to ultimately migrate the config file to use a standard format, then I understand wanting to make it work with configparser.

common/konfig.py
buhtz: Mhm... reminds me of something I read in the past but never understood. 🤣 Now I have real world illustration. 🎉 Thank you for this.

Anyway, I don't think the context manager is a good solution at all. There is a rule about testing: Never modify productive code just because of a unit test. One alterative to the context manager would be a virtual file system via `pyfakefs` while running the test suite. I really like PyFakeFS for specific cases but in our case it would make things complicated without much benefit.

I might remove the file handling logic from `Konfig` class. It should be something like this:

    fp = Path.home() / 'somewhere' / 'configfile'
    with fp.open('r', encoding='utf-8') as handle:
        config = Konfig(handle)

        # or ...
        config = Konfig()
        config.load(handle)

This way it is more easier using in-memory files in a unit test:

    def test_foobar():
        buffer = StringIO('filed=value')
        config = Konfig()
        config.load(handle)

common/konfig.py
buhtz: Technically you are right. There is no difference. But I would like to have the syntax also platform independent and not only the behavior.

It is just a matter of habit and style. There is no strict technical reason for it.

I might be to pedantic about it because I am also not aware of any linter rule checking for something like this. 😆

buhtz:
> How hard would it be to write a custom parser?

We still have a custom parser, but I want to get rid of that part of the code and don't want to keep it just because of something like this. Our problem with `%` is [not new](https://stackoverflow.com/q/64256563/4865723). I just haven't had the time to investigate that further.

In a long run I want to move from that INI-like format to TOML. One, because it is the current Python common default config format in nearly all use cases. Second, it is easier to read and to manage. Having a human readable config file format would also make it less urgent (or needless) to have a CLI/TUI interface (#254) for users running BIT on a headless server without X.

common/konfig.py
daviewales: Happy to respect a style choice.

common/konfig.py
daviewales: I think that's a good call.

daviewales:
Hmm... Looks like you can just use:

``` python
configparser.ConfigParser(interpolation=None)
```

So sounds like it should be possible to make it work.

common/konfig.py
```diff
@@ -324,46 +317,21 @@ def profile_names(self) -> list[str]:
...
-            logger.debug(f'Configuration read from "{self._path}".')
+        # raw content
+        content = buffer.read()
 
         # Add section header to make it a real INI file
         self._config_parser.read_string(f'{self._DEFAULT_SECTION}\n{content}')
 
         # The one and only main section
-        self._conf = self._config_parser['bit']
+        self._conf = self._config_parser[self._DEFAULT_SECTION[1:-1]]
```
daviewales: Do you mean `[1:]`?

``` python
a = [1, 2, 3, 4]
a[1:-1]
# [2, 3]
a[1:]
# [2, 3, 4]
```

buhtz: No. I just needed to cut of the square brackets from the string.

```
>>> _DEFAULT_SECTION = '[bit]'
>>> _DEFAULT_SECTION[1:-1]
'bit'
```

buhtz:
OK, I know implemented all config-fields from Config class in Konfig class as properties including man page content.

buhtz:
TODO: Add Konfig is singleton unit test

buhtz:
Nearly after two years I would say this PR is finished and ready to merge. What a big mess, but I saw no other way. :rofl: :hankey:

Issue #1003, opened 2019-05-21
RuntimeError: reentrant call inside <_io.BufferedWriter name=''>


likg:
With backintime (1.2.0) when you try to backup whole system (/ mount point) backup task fails with the following error:
```bash
Back In Time
Version: 1.2.0

Back In Time comes with ABSOLUTELY NO WARRANTY.
This is free software, and you are welcome to redistribute it
under certain conditions; type `backintime --license' for details.

INFO: Lock
INFO: Take a new snapshot. Profile: 1 Main profile
INFO: Call rsync to take the snapshot
Traceback (most recent call last):
  File "/usr/share/backintime/common/backintime.py", line 1165, in <module>
    startApp()
  File "/usr/share/backintime/common/backintime.py", line 517, in startApp
    args.func(args)
  File "/usr/share/backintime/common/backintime.py", line 739, in backup
    ret = takeSnapshot(cfg, force)
  File "/usr/share/backintime/common/backintime.py", line 94, in takeSnapshot
    ret = snapshots.Snapshots(cfg).backup(force)
  File "/usr/share/backintime/common/snapshots.py", line 692, in backup
    ret_val, ret_error = self.takeSnapshot(sid, now, include_folders)
  File "/usr/share/backintime/common/snapshots.py", line 1036, in takeSnapshot
    proc.run()
  File "/usr/share/backintime/common/tools.py", line 1981, in run
    self.callback(line, self.user_data)
  File "/usr/share/backintime/common/snapshots.py", line 797, in rsyncCallback
    self.setTakeSnapshotMessage(0, _('Take snapshot') + " (rsync: %s)" % line)
  File "/usr/share/backintime/common/snapshots.py", line 143, in setTakeSnapshotMessage
    self.snapshotLog.append('[I] '  + message, 3)
  File "/usr/share/backintime/common/snapshotlog.py", line 195, in append
    self.logFile.write(msg + '\n')
  File "/usr/share/backintime/common/tools.py", line 1513, in handler
    self.callback()
  File "/usr/share/backintime/common/snapshotlog.py", line 203, in flush
    self.logFile.flush()
RuntimeError: reentrant call inside <_io.BufferedWriter name='/root/.local/share/backintime/takesnapshot_.log'>
```
Simple workaround in this case is to add backintime log file (e.g. `/root/.local/share/backintime/takesnapshot_.log`) to exclude list.

Please review this case, maybe log file should be excluded (or handled in a special way), to avoid such issue.

[...]

PR #1856, opened 2024-09-01
chore(doc): Improve docu about FIFO and Alarm class

buhtz:
While I tried to understand `tools.Alarm` and `password_ipc.FIFO` I added some more comments and improved doc strings.

Happens in preparation of PR #1850 .

### DIFF
--- a/common/password_ipc.py
+++ b/common/password_ipc.py
@@ -58,12 +58,20 @@ class FIFO(object):
         """
         # sys.stdout.write('read fifo\n')
         if not self.isFifo():
+            # TODO raise an Exception or write to stderr
             sys.exit(1)
 
         self.alarm.start(timeout)
 
         with open(self.fifo, 'r') as handle:
+            # Will wait until data is available,
+            # or an exception (e.g. exception.Timeout) is raised.
             ret = handle.read()
+
+        # If the alarm timeout ends but read() received not data, a
+        # excpetion.Timeout will be raised at this point.
+        # The exception will be catched far away in
+        # password.py::Password_Cache.run().
 
         self.alarm.stop()
 
--- a/common/password_ipc.py
+++ b/common/password_ipc.py
@@ -1,19 +1,10 @@
-#    Copyright (C) 2012-2022 Germar Reitze
+# SPDX-FileCopyrightText: © 2012-2022 Germar Reitze
 #
-#    This program is free software; you can redistribute it and/or modify
-#    it under the terms of the GNU General Public License as published by
-#    the Free Software Foundation; either version 2 of the License, or
-#    (at your option) any later version.
+# SPDX-License-Identifier: GPL-2.0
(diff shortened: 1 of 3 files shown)


common/password_ipc.py
```diff
@@ -24,88 +15,118 @@
...
+        # sys.stdout.write('read fifo\n')
         if not self.isFifo():
+            # TODO raise an Exception or write to stderr
             sys.exit(1)
+
         self.alarm.start(timeout)
-        with open(self.fifo, 'r') as fifo:
-            ret = fifo.read()
+
+        with open(self.fifo, 'r') as handle:
```
daviewales, line 66: I feel that `fifo` is more descriptive than `handle`.

common/tools.py
```diff
@@ -1981,23 +1981,25 @@ def checkEqual(self, path):
...
+    """Establish a callback function that is called after a timeout using
+    SIGALRM signal.
 
-    The implementation uses a SIGALRM signal so
-    do not call code in the callback that does not support multi-threading
-    (reentrance) or you may cause non-deterministic "random" RTEs.
+    If no callback is specified a `exception.Timeout` will be raised instead.
+    The implementation uses a SIGALRM signal. Attention: Do not call code in
+    the callback that does not support multi-threading (reentrance) or you may
+    cause non-deterministic "random" RTEs (RTE??? race conditions?).
```
daviewales: Guess: Run Time Error?

daviewales: (Confirmed: see error message in https://github.com/bit-team/backintime/issues/1003)

common/tools.py
```diff
@@ -2006,49 +2008,53 @@ def __init__(self, callback = None, overwrite = True):
...
             return
+
         try:
             # Warning: This code may cause non-deterministic RTEs
             #          if the handler function calls code that does
             #          not support reentrance (see e.g. issue #1003).
             signal.signal(signal.SIGALRM, self.handler)
             signal.alarm(timeout)
         except ValueError:
+            # Why???
```
daviewales, line 2030: ValueErrors can be raised from [signal.signal](https://docs.python.org/3/library/signal.html#signal.signal). Doesn't explain _why_ we suppress the errors.

buhtz: Maybe remove that in the future. I just didn't want to modify to much in one step. Who know what problem this can cause!?

common/password_ipc.py
buhtz: My first intention was to separate (visually) "self.fifo" from "fifo". So I was looking for another name. For me "handle" is common when it comes to file handlers in with-open-blocks. Technically "handle" is not a file but just a file descriptor or cursor in that file.

common/tools.py
daviewales: Yes, I'd leave it for now.

PR #1857, opened 2024-09-01
refactor!: Automigration of (very) old config files (version 4 or lower) not supported anymore

buhtz:
Config files with version 4 or lower not longer migrated to the latest version 6. That versions not longer in the wild. And I can remove some code because of it.

### DIFF
--- a/common/config.py
+++ b/common/config.py
@@ -56,13 +56,13 @@ from exceptions import PermissionDeniedByPolicy, \
                        InvalidChar, \
                        InvalidCmd, \
                        LimitExceeded
-import version
 
 
 class Config(configfile.ConfigFileWithProfiles):
     APP_NAME = 'Back In Time'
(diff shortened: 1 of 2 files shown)


common/config.py
```diff
@@ -199,38 +189,14 @@ def __init__(self, config_path=None, data_path=None):
...
-                    self.setExclude(old_values, profile_id)
-
-                    #remove keys
-                    self.removeProfileKey('snapshots.include_folders', profile_id)
-                    self.removeProfileKey('snapshots.exclude_patterns', profile_id)
+                logger.error(
+                    'The config file version is 4 or lower. This config was '
+                    'made with a version of Back In Time that is out dated. '
+                    'Because of that upgrading config to the current version '
+                    'is not possible.',
```
daviewales: Would be good to specify the latest version which supports v4 config upgrade.

PR #1859, opened 2024-09-02
fix: Reset cached SSH password when changed

### LINKED ISSUE #1852: backintime-askpass and ssh-add enter infinite loop when cached SSH key password does not match current password
<details>
<summary><code>backintime --diagnostics</code></summary>

``` sh
{
    "backintime": {
        "name": "Back In Time",
        "version": "1.5.3-dev.366ef003",
        "running-as-root": false,
        "latest-config-version": 6,
        "keyring-supported": true
    },
    "host-setup": {
        "OS": {
            "/etc/os-release": "Linux Mint 22",
            "/etc/lsb-release": "DISTRIB_ID=LinuxMint\nDISTRIB_RELEASE=22\nDISTRIB_CODENAME=wilma\nDISTRIB_DESCRIPTION=\"Linux Mint 22 Wilma\"\n",
            "/etc/debian_version": "trixie/sid\n"
        },
        "platform": "Linux-6.8.0-40-generic-x86_64-with-glibc2.39",
        "system": "Linux #40-Ubuntu SMP PREEMPT_DYNAMIC Fri Jul  5 10:34:03 UTC 2024",
        "display-system": "x11",
        "locale": "en_AU, UTF-8",
        "PATH": "/home/UsernameReplaced/.bun/bin:/home/UsernameReplaced/.cargo/bin:/home/UsernameReplaced/.local/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin:/home/UsernameReplaced/.local/bin:/home/UsernameReplaced/.local/bin:/home/UsernameReplaced/.cargo/bin",
    },
    "python-setup": {
        "python": "3.12.3 main Jul 31 2024 17:43:48 CPython GCC 13.2.0",
        "sys.path": [
            "/usr/lib/python312.zip",
            "/usr/lib/python3.12",
            "/usr/local/lib/python3.12/dist-packages",
            "/usr/lib/python3/dist-packages",
            "/usr/lib/python3.12/dist-packages"
        ],
        "qt": {
            "Version": "PyQt 6.6.1 / Qt 6.4.2",
            "Theme": "Mint-Y-Sand",
            "Theme Search Paths": [
                "/usr/share/cinnamon/icons",
                "/usr/share/icons",
                ":/icons"
            ],
            "Fallback Theme": "",
            "Fallback Search Paths": []
        }
    },
    "external-programs": {
        "rsync": {
            "version": "3.2.7",
            "protocol": "31.0",
            "capabilities": "file_bits: 64; inum_bits: 64; timestamp_bits: 64; long_int_bits: 64; socketpairs: True; symlinks: True; symtimes: True; hardlinks: True; hardlink_specials: True; hardlink_symlinks: True; IPv6: True; atimes: True; batchfiles: True; inplace: True; append: True; ACLs: True; xattrs: True; secluded_args: optional; iconv: True; prealloc: True; stop_at: True; crtimes: False",
            "optimizations": "SIMD_roll: True; asm_roll: False; openssl_crypto: True; asm_MD5: False",
            "checksum_list": "xxh128, xxh3, xxh64, md5, md4, sha1, none",
            "compress_list": "zstd, lz4, zlibx, zlib, none",
            "daemon_auth_list": "sha512, sha256, sha1, md5, md4"
        },
        "ssh": "OpenSSH_9.6p1 Ubuntu-3ubuntu13.4, OpenSSL 3.0.13 30 Jan 2024",
        "sshfs": "3.7.3",
        "encfs": "1.9.5",
        "shell": "/bin/bash",
        "shell-version": "GNU bash, version 5.2.21(1)-release (x86_64-pc-linux-gnu)"
    }
}
```

</details>

This bug follows from discussion in https://github.com/bit-team/backintime/pull/1839#issuecomment-2308382426.

In summary, `backintime-askpass` is used (abused?) to cache the SSH key password for password-protected SSH keys.
To reproduce the problem, create a test SSH key and protect it with a password. Then add the key to the `ssh-agent` with `ssh-add` using the commands below, which will use `backintime-askpass` to get the password. Then remove the key from `ssh-agent`, change the key password, and add it to `ssh-agent` again. Instead of prompting you to enter the new password, it will silently provide the old password to `ssh-add`, which will fail, then ask `backintime-askpass` to provide the password again. This continues in a loop indefinitely.

## Reproduce

``` sh
# create test key `id_rsa_test` with password `test`
PASSWORD=testing
KEY=~/.ssh/id_rsa_test
ssh-keygen -t rsa -b 4096 -N "$PASSWORD" -f "$KEY"

# Add the key to ssh-agent, using the `backintime-askpass` tool to get the password
SSH_ASKPASS=backintime-askpass SSH_ASKPASS_REQUIRE=prefer ASKPASS_PROFILE_ID=1 ASKPASS_MODE=ssh ssh-add "$KEY"

# Remove the key from the `ssh-agent`
ssh-add -d "$KEY"

# Change the password
OLD_PASSWORD="$PASSWORD"
PASSWORD=testing2
ssh-keygen -p -P "$OLD_PASSWORD" -N "$PASSWORD" -f "$KEY"

# Add the key to ssh-agent again, using the `backintime-askpass` tool to get the password
SSH_ASKPASS=backintime-askpass SSH_ASKPASS_REQUIRE=prefer ASKPASS_PROFILE_ID=1 ASKPASS_MODE=ssh ssh-add "$KEY"
```

Note: The above assumes that `backintime-askpass` is in a clean state.
If you have already used `backintime-askpass` once, then you will reproduce the issue the first time you run `ssh-add`, and don't need to proceed to the second run. You will know that you have reproduced the issue when `ssh-add` hangs, and starts to use lots of CPU.

## Next steps

- Think about whether it is a good idea to use `backintime-askpass` to cache SSH passwords. Should we instead use the system login keychain to store the password across reboots and the built-in `ssh-agent` caching to store it for the duration of a login session? (I don't know enough about the use-cases to confidently answer this question yet.)
- If we _do_ need to keep `backintime-askpass`, we could tweak it to prevent it entering the infinite loop:
    - A simple solution would be for `backintime-askpass` to quickly check if the cached password can unlock the key, before providing the password to `ssh-add`. This can be done by piping the password to `ssh-keygen -y -f "$KEY"`, and checking the exit status as described in this [SO answer](https://stackoverflow.com/a/23666831/). This is simple enough that it might make sense to just apply this fix, and evaluate the first question later.

daviewales:
Before, the cached SSH password would be used without checking if it was valid. Now, we check if it's valid before using it. If it's not valid, we ask the user to enter a new password before proceeding.

Fixes #1852


common/sshtools.py
```diff
@@ -412,6 +413,37 @@ def unlockSshAgent(self, force=False):
                     env['ASKPASS_TEMP'] = thread.temp_file
                     thread.start()
 
+                from pprint import pprint
+                with open('env.test', 'w') as file:
+                    pprint(env, stream=file)
+
```
daviewales: TODO: Remove this debugging code.

common/sshtools.py
```diff
@@ -412,6 +413,37 @@ def unlockSshAgent(self, force=False):
...
+                        env=env | {'SSH_ASKPASS_REQUIRE': 'prefer'}
+                )
+
+                # if backintime-askpass supplied an invalid cached password
+                if proc.returncode > 0:
+                    pw = password.Password()
+
+                    # TODO: Figure out correct way to get pw_id
+                    pw_id = 1
+
```
daviewales: TODO: What does `pw_id` mean? How do we know what the correct value is?

daviewales: I need to do some more digging here. `qt/settingsdialog.py` talks about `password_1` and `password_2`, and this seems directly related to `pw_id = 1` and `pw_id = 2`. I'm wondering if one is the password for the SSH key, and the other is the encryption password?

buhtz: It depends on the profile mode. Using "SSH encrypted" you will get this in the Generals tab.
image
"SSH private key" is "password_1" and "Encryption" is "password_2". You can see this relation in the beginning of `Config.saveProfile()`.

daviewales: I've added a comment explaining where this comes from in c3bd0a132.

buhtz:
Do you think the PR is finished on your side? In this case I would re-label it as "RequestToReview" and see if we can get some more feedback from others. btw: You should click on "Update branch" and resolve merge conflicts if they appear.

daviewales:
I don't think I can add labels, but yes, I think it's ready for review.

buhtz:
Once again, a big thank you for your contribution!  

My [request for a review on the mailing list](https://mail.python.org/archives/list/bit-dev@python.org/thread/T3SW5HUKPFBR2BRTRRCBDDLXYJGBETYS/) did not receive any response.  
I will merge it soon without really being able to assess the implications.  
I don't know enough about the functionality in this area and don't have the resources to delve into it.

A new release is not planned in the near future, so it shouldn't be a big problem.

PR #1860, opened 2024-09-03
Feature: Warn if destination directory is formatted as ntfs

### LINKED ISSUE #1854: Invalid argument (22) (NTFS target drive)
I was helping a co-worker setup BIT on his laptop. When he ran it the first time, he got a 'Invalid argument (22)' error.

Turns out his backup drive was formatted as NTFS.

I'm thinking it would be a good idea to add a check, on the destination directory, to ensure it's a compatible file system.

The question is, what constitutes a compatible file system?

I think the error 22 comes from trying to use certain characters in file names.

fallingrock:
Warn user if destination directory is formatted as ntfs to avoid "Unsupported operation" errors.

Fix #1854

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,6 +1,7 @@
 Back In Time
 
 Version 1.5.3-dev (development of upcoming release)
+* Feature: Warn if destination directory is formatted as NTFS
 * Breaking Change: Minimal Python version 3.9 required (#1731)
 * Fix: Prevent duplicates in Exclude/Include list of Manage Profiles dialog
 * Fix: Fix Qt segmentation fault when canceling out of unconfigured BiT (#1095) (Derek Veit @DerekVeit)
--- a/CHANGES
+++ b/CHANGES
@@ -1,7 +1,7 @@
 Back In Time
 
 Version 1.5.3-dev (development of upcoming release)
-* Feature: Warn if destination directory is formatted as NTFS
+* Feature: Warn if destination directory is formatted as NTFS (#1854) (David Gibbs @fallingrock)
 * Breaking Change: Minimal Python version 3.9 required (#1731)
 * Fix: Prevent duplicates in Exclude/Include list of Manage Profiles dialog
 * Fix: Fix Qt segmentation fault when canceling out of unconfigured BiT (#1095) (Derek Veit @DerekVeit)
--- a/common/config.py
+++ b/common/config.py
@@ -133,6 +133,9 @@ class Config(configfile.ConfigFileWithProfiles):
     ENCODE = encfstools.Bounce()
     PLUGIN_MANAGER = pluginmanager.PluginManager()
 
+    NTFS_FILESYSTEM_WARNING = _('The destination filesystem for {path} is formatted with NTFS, '
+                                'which has known incompatibilities with Unix-style filesystems.')
+
     def __init__(self, config_path=None, data_path=None):
         """Back In Time configuration (and much more then this).
 
@@ -495,10 +498,7 @@ class Config(configfile.ConfigFileWithProfiles):
             return False
 
         elif fs.startswith('ntfs'):
-            self.notifyError(
-                _('Destination filesystem for {path} is formatted with NTFS '
-                  'which has known incompatibilities with Unix style file systems.')
-                .format(path=value))
+            self.notifyError(self.NTFS_FILESYSTEM_WARNING.format(path=value))
 
         elif fs == 'cifs' and not self.copyLinks():
             self.notifyError(_(
--- a/common/config.py
+++ b/common/config.py
@@ -494,6 +494,12 @@ class Config(configfile.ConfigFileWithProfiles):
 
             return False
 
+        elif fs.startswith('ntfs'):
+            self.notifyError(
+                _('Destination filesystem for {path} is formatted with NTFS '
+                  'which has known incompatibilities with Unix style file systems.')
+                .format(path=value))
+
         elif fs == 'cifs' and not self.copyLinks():
             self.notifyError(_(
                 'Destination filesystem for {path} is an SMB-mounted share. '
--- a/qt/settingsdialog.py
+++ b/qt/settingsdialog.py
@@ -1947,10 +1947,11 @@ class SettingsDialog(QDialog):
 
                 self.config.removeProfileKey('snapshots.path.uuid')
 
-            if tools.filesystem(path).startswith('ntfs'):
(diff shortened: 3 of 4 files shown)


qt/settingsdialog.py
```diff
@@ -15,6 +15,7 @@
 import copy
 import re
 import getpass
+import psutil
```
buhtz: I wouldn't like to add a new dependency for that use case.

Why not using the shell tools?

    >>> r = subprocess.run(['df', '--output=fstype', '.'], stdout=subprocess.PIPE, text=True)
    >>> r.stdout.splitlines()[1]
    'ext4'

psutils also only read from the /proc filesystem.

**EDIT**:
BIT still hast this "feature". See https://github.com/bit-team/backintime/blob/8c7afef81ae79be491a848400261eabdc3e90c8c/common/tools.py#L1450

fallingrock:
Hmmm ... build failed because psutil wasn't available.

Can I add psutil to the travis.yml file to fix this?

buhtz (changes requested):
I wonder if this will help in all cases.

The file system depends on the way how the ntfs volume is mounted.
How was the volume mounted in case of the issue you reported? Would it catch that case?

CHANGES
```diff
@@ -1,6 +1,7 @@
 Back In Time
 
 Version 1.5.3-dev (development of upcoming release)
+* Feature: Warn if destination directory is formatted as NTFS
```
buhtz:
```suggestion
* Feature: Warn if destination directory is formatted as NTFS (#1854) (David Gibbs @fallingrock)
```

qt/settingsdialog.py
fallingrock: Wasn't aware of the filesystem method ... I'll check it out.

qt/settingsdialog.py
```diff
@@ -1947,6 +1947,14 @@ def btnSnapshotsPathClicked(self):
 
                 self.config.removeProfileKey('snapshots.path.uuid')
 
+            if tools.filesystem(path).startswith('ntfs'):
+                text = _('Warning: {0} is formatted as NTFS which has known incompatabilities '
+                         'with unix style file systems.\n\n'
+                         'Are you sure you want to use it as your backup destination?')
+                question = text.format(path)
```
buhtz: Imagine our translators using the web-frontend at translators.codeberg.org. Some of them have no coding background. They also need more context. Because of that it is important to give placeholders good names. The details are [hidden in our docs](https://github.com/bit-team/backintime/blob/dev/common/doc-dev/2_localization.md#instructions-for-the-translation-process).

```suggestion
                text = _('Warning: {path} is formatted as NTFS which has known incompatibilities '
                         'with Unix style file systems.'
                         'Are you sure you want to use it as your backup destination?')
                question = text.format(path=path)
```

I also removed the two `\n` because of our translators. I am not sure if, so please checkout if the dialog box does wrap the text by itself. And while you are checking it please create a screenshot of message box and attach it to the PR. If the text is not wrapped, please simply make two translatable strings like this:

    text = _('foo') + '\n\n' + _('bar')

Beside the text, the translators will also get a screenshot of the message box. So they will understand.

It is damn hard to get translators and keep them motivated. So I try to make it as easy as possible for them. 😄

buhtz (changes requested):

buhtz:
The message seems a bit unspecific for some users. Are there cases where NTFS volumes do work? So why are the users free to use NTFS with BIT? Shouldn't we just deny such destination volumes?

fallingrock:
> The message seems a bit unspecific for some users. Are there cases where NTFS volumes do work? So why are the users free to use NTFS with BIT? Shouldn't we just deny such destination volumes?

I think it depends on if the user has filenames containing certain characters.

The following reserved characters:   
 \> (less than)
 \< (greater than)   
 : (colon)   
 " (double quote)   
 / (forward slash)   
 \ (backslash)   
 | (vertical bar or pipe)   
 ? (question mark)   
 \* (asterisk)

fallingrock:
Screenshot of new error messages
image
image

buhtz:
> I think it depends on if the user has filenames containing certain characters.
> The following reserved characters: ...

Interesting. In this case my alternative suggested approach just doing an rsync test run and check after that if hardlinks appear won't have caught such a situation. Your approach seems better.

Would you like to add an entry in FAQ.md about that NTFS details? We later can move that FAQ entry somewhere else into the docu.

common/config.py
```diff
@@ -494,6 +494,12 @@ def setSnapshotsPath(self, value, profile_id = None, mode = None):
 
             return False
 
+        elif fs.startswith('ntfs'):
+            self.notifyError(
+                _('Destination filesystem for {path} is formatted with NTFS '
+                  'which has known incompatibilities with Unix style file systems.')
```
buhtz: I am not skilled with English words. I just asked the "stochastic parrot" and modified the string a bit. So feel free to modify it either.

```suggestion
                _('The destination filesystem for {path} is formatted with NTFS, '
                  'which has known incompatibilities with Unix-style filesystems.')
```

qt/settingsdialog.py
```diff
@@ -1947,6 +1947,15 @@ def btnSnapshotsPathClicked(self):
 
                 self.config.removeProfileKey('snapshots.path.uuid')
 
+            fs = tools.filesystem(path)
+            if fs.startswith('ntfs'):
+                text = _('Warning: Destination filesystem for {path} is formatted with NTFS '
+                         'which has known incompatibilities with Unix style file systems.\n'
+                         'Are you sure you want to use it as your backup destination?')
```
buhtz: Beside avoiding escape characters like `\n` in the translation string I also try to avoid referencing user in person via "you".

```suggestion
                text = '\n'.join([
                       _('Warning: The destination filesystem for {path} is formatted with NTFS, '
                         'which has known incompatibilities with Unix-style filesystems.'),                             
                       _('Is this the backup destination to be used?')
                })
```

buhtz (changes requested):
The PR looks good. Thank you very much for your efforts. A FAQ entry would be the cherry on top.

I would like to wait for review from the other team mates: emtiu and/or aryoda. They are not quit available currently so it can take some time.

common/config.py
fallingrock: I'm no english major either (GIBBeriSh is my family language :) ) ... I was just matching the existing message pattern in config.py.

buhtz (approved):

PR #1863, opened 2024-09-05
chore: Debug output and translation recommendations

buhtz:
My intention was to reduce the debug output at all. But I wasn't sure about all the Qt and Keyring related stuff. So in the end I didn't change so much.
Additionally I improved the contribution docu about how to deal with translatable strings.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -81,6 +81,7 @@ request is accepted.
   `unittest` rather than `pytest`. If you know the difference please try follow
   the _Classical (aka Detroit) school_ instead of _London (aka mockist)
   school_.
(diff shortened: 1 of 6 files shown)


daviewales:
Looks OK.

buhtz:
Thank you for the review.
If you are interested in pimping up your profile, you an use the "Review Changes" button in the "Files changed" section/tab.
image

This might influence this graphic in your profile.
image

Or maybe you don't see that button because of permission settings?

daviewales:
Thanks. I usually use that when I'm going to make a bunch of suggestions, and want to send them all at once. To be honest my profile is not a consideration! I didn't have any suggestions for this one, but wanted to let you know I'd looked.

buhtz:
> To be honest my profile is not a consideration!

Heartwarming. 😄 Sometimes I have to deal with profile-pimpers.

PR #1865, opened 2024-09-06
refactor: Separate code about Manage profile dialog

buhtz:
Blocked by
 - #1892
 - #1886

About he PR
- I gave the Manage Profiles (aka Settings) dialog its own module (folder `manageprofiles`).
- The code related to the Generals tab (the first tab in the dialog) is separated into its own file.
- I did not touch the other tabs code to prevent blowing up the PR.

IMHO a code review makes not much sense because there was to much code that is modified. [Testing](https://github.com/bit-team/backintime/blob/dev/common/doc-dev/BiT_release_process.md#manual-testing---recommendations) the dialog and its features (e.g. in a VM) would make more sense and has more value.

PR #1867, opened 2024-09-08
Fix/1864setsnapshotspath

### LINKED ISSUE #1864: Blocking/Refactor: Make Config.setSnapshotsPath() dumb
This is about `common/config.py::Config.setSnapshotsPath()`.
https://github.com/bit-team/backintime/blob/8c7afef81ae79be491a848400261eabdc3e90c8c/common/config.py#L440

That method is to complex and does to many things for a "simple" config setter. We need to find a solution to decouple/isolate that code somehow.

The reason is I do have two big PRs that are blocked because of that one method. One is PR #1850 (replace `class Config` with a new one) and the other is PR #1865 (heavy refactoring of the Manage Profiles dialog).

buhtz:
Complex refactoring and code movement. :rescue_worker_helmet:  Behavior of BIT shouldn't be modified. :fearful: 

This PR fix refactoring issue #1864 and prepares for a new config management class (PR #1850) and refactoring SettingsDialog code (PR #1865). The problem this PR solves was that the method `Config.setSnapshotsPaths()` has to much logic in it not belonging in that method nor into that class. I moved and "parked" the code in the `tools.py` module.

The current solution give me the ability to easier decouple the code in `SettingsDialog.saveProfile()` in context of PR #1865.

Of course I modified several other things but tried to keep that PR as simple as possible to reduce the risk of introducing bugs.

Fix #1864

### DIFF
--- a/common/tools.py
+++ b/common/tools.py
@@ -369,6 +369,160 @@ def get_native_language_and_completeness(language_code):
 
     return (name, completeness)
 
+# |---------------------------------------|
+# | Snapshot handling                     |
+# |                                       |
+# | Candidates for refactoring and moving |
+# | into better suited moduls/classes     |
+# |---------------------------------------|
+
+
+def validate_snapshots_path(path, cfg, profile_id=None):
+    """Check if the given path is valide for being a snapshot path.
+
+    It is checked if it is a folder, if it is writeable, if the filesystem is
+    supported and several other things.
+
+    Args:
+        path: The path to validate as a snapshot path.
+        cfg: The config instance.
+        profile_id: Related snapshots profile id.
+
+    Returns:
+        (bool): `False` or `True`
+    """
+    if not isinstance(path, pathlib.Path):
+        path = pathlib.Path(path)
+
+    if profile_id is None:
+        profile_id = cfg.currentProfile()
+
+    mode = cfg.snapshotsMode(profile_id)
+
+    if not path.is_dir():
+        cfg.notifyError(_('Invalid option. {path} is not a folder.')
+                        .format(path=path))
+        return False
+
+    # build full path
+    # <path>/backintime/<host>/<user>/<profile_id>
+    host, user, profile = cfg.hostUserProfile(profile_id)
+    full_path = path / 'backintime' / host / user / profile
+
+    # create full_path
+    try:
+        full_path.mkdir(mode=0o777, parents=True, exist_ok=True)
+
+    except PermissionError:
+        cfg.notifyError('\n'.join([
+            _('Creation of following folder failed:'),
+            str(full_path),
+            _(f'Write access may be restricted.')]))
+        return False
+
+    # Test filesystem
+    rc, msg = is_filesystem_valid(
+        full_path, path, mode, cfg.copyLinks())
+    if msg:
+        cfg.notifyError(msg)
+    if rc is False:
+        return False
+
+    # Test write access for the folder
+    rc, msg = is_writeable(full_path)
+    if msg:
+        cfg.notifyError(msg)
+    if rc is False:
+        return False
+
+    return True
+
+
+def is_filesystem_valid(full_path, msg_path, mode, copy_links):
+    """
+    Args:
+        full_path: The path to validate.
+        msg_path: The path used for display in error messages.
+        mode: Snapshot profile mode.
+        copy_links: Snapshot profiles copy links setting.
+
+    Returns:
+        (bool, str): A boolean value indicating success or failure and a
+            msg string.
+
+    """
+    fs = filesystem(full_path if isinstance(full_path, str) else str(full_path))
+
+    msg = None
+
+    # DEV NOTE
+    # notifyError() results in a messagebox
+    # Might be a candidate for a simple error-event-handler
+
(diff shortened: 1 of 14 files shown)


common/tools.py
```diff
@@ -369,6 +370,160 @@ def get_native_language_and_completeness(language_code):
...
+    Args:
+        folder: The folder to check.
+
+    Returns:
+        (bool, str): A boolean value indicating success or failure and a
+            msg string.
+    """
+
+    if not isinstance(folder, pathlib.Path):
+        folder = pathlib.Path(folder)
```
daviewales: You can skip the check, as `Path(Path('example'))` will work. (Pathlib does its own `isinstance` check.)

```suggestion
    folder = pathlib.Path(folder)
```

Not sure which is strictly better. Depends on whether you're optimizing to minimise lines of code in BiT source, or instructions executed by Python.

The original code with the `isinstance` check is _slightly_ less work for Python, but this suggestion is less code for us.
See the pathlib initialisation source:
https://github.com/python/cpython/blob/77bb14622172c1a5d7d712bfe7050e7277d7ebb0/Lib/pathlib.py#L358

common/tools.py
```diff
@@ -369,6 +370,160 @@ def get_native_language_and_completeness(language_code):
...
+
+def is_writeable(folder):
+    """Test write access for the folder.
+
+    Args:
+        folder: The folder to check.
+
+    Returns:
+        (bool, str): A boolean value indicating success or failure and a
+            msg string.
```
daviewales, line 515: Would it make more sense to just return `True` or `False`, and let the caller determine the message? The caller has everything they need to construct the message emitted by the exception below, and we don't even return a message for success.

buhtz: You are right. This is kind of ugly. But I tried to be pragmatic as possible without modifying to much.
All these nofiyError() handler code in BIT is another big topic. I would realize this with a pythonic Observer-Pattern-like approach. I have done this before in other applications, too.

But for this PR I would keep it as it is and touch in another time. That code itself will someday move into another class or module. It is just parked currently.

buhtz:
Slightly modified to totally decouple the `Config` class from the parked code. Of course that "solution" is not permanent.

common/tools.py
```diff
@@ -378,61 +379,64 @@ def get_native_language_and_completeness(language_code):
...
     supported and several other things.
 
+    Dev note  (buhtz, 2024-09): That code is a good candidate to get moved
+        into a class or module.
+
     Args:
         path: The path to validate as a snapshot path.
-        cfg: The config instance.
-        profile_id: Related snapshots profile id.
+        host_user_profile: I three item list.
```
daviewales: Perhaps some more detail about what the three items are?

common/tools.py
```diff
@@ -378,61 +379,64 @@ def get_native_language_and_completeness(language_code):
...
+        error_handler(_('Invalid option. {path} is not a folder.')
+                      .format(path=path))
         return False
 
     # build full path
     # <path>/backintime/<host>/<user>/<profile_id>
-    host, user, profile = cfg.hostUserProfile(profile_id)
-    full_path = path / 'backintime' / host / user / profile
+    host_user_profile = pathlib.Path(os.sep.join(host_user_profile))
+    full_path = path / 'backintime' / host_user_profile
```
daviewales: Could just use `*` unpacking:

``` python
    host_user_profile = pathlib.Path(*host_user_profile)
    full_path = path / 'backintime' / host_user_profile
```

Alternatively, could use `os.sep.join`, but don't need `pathlib.Path`:

``` python
    host_user_profile = os.sep.join(host_user_profile)
    full_path = path / 'backintime' / host_user_profile
```

In fact, pathlib is smart enough that you could just do:

``` python
    full_path = path / 'backintime' / '/'.join(host_user_profile)
```

But the slickest option might be:

``` python
    full_path = pathlib.Path(path, 'backintime', *host_user_profile)
```

common/tools.py
```diff
@@ -378,61 +379,64 @@ def get_native_language_and_completeness(language_code):
...
+        mode: The profiles mode.
+        copy_links: The copy links value.
+        error_handler: Handle function receiving error messages.
 
-    Returns:
-        (bool): `False` or `True`
+    Returns: Success (`True`) or failure (`False`).
     """
     if not isinstance(path, pathlib.Path):
         path = pathlib.Path(path)
```
daviewales: Another check we can skip.

common/tools.py
```diff
@@ -369,6 +371,162 @@ def get_native_language_and_completeness(language_code):
...
+        return False
+
+    # build full path
+    # <path>/backintime/<host>/<user>/<profile_id>
+    host_user_profile = pathlib.Path(os.sep.join(host_user_profile))
+    full_path = path / 'backintime' / host_user_profile
+
+    # create full_path
+    try:
+        full_path.mkdir(mode=0o777, parents=True, exist_ok=True)
```
daviewales, line 432: Should creating the directory happen in a different function?
Or should this function be renamed to reflect the fact that it creates a directory?

buhtz:
Solved merge conflicts caused by merging PR #1860 (Warn about NTFS).

Issue #1575, opened 2023-12-03
Migrate to Python build-system (PEP 517 & PEP 621)


buhtz:
This is a meta issue documenting the migration of BITs build-system which currently do use `make`.

The folder structure will migrate to the [`src` layout](https://packaging.python.org/en/latest/tutorials/packaging-projects) and the projects metadata will get stored in a [`pyproject.toml` file](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html) (see  [PEP 621 - Storing project metadata in `pyproject.tom`](https://peps.python.org/pep-0621)). Based on that the build-system will migrate to `setuptools` or something similar (e.g. [`hatch`](https://hatch.pypa.io)) and the `make` build-system will be eliminated.

BIT has some special needs that need to be taken into account.
 - The migration need to be done all at once. I see no slow and secure step-by-step approach.
 - It do use two components, the CLI `common` and the GUI `qt`.
 - It do use GNU gettext for localization.
 - It need to run as root (currently via `pkexec`) on users request.
 - The project is that much complex and old that we expect to discover further issues.
 - serviceHelper.py DBUS daemon to modify Udev rules
 - polkit/pkexec stuff to run bit as root
 - systray maybe
 - password cache
 - icons, logos, po/mo files
 - HTML documentation
 - maybe use `importlib.resources`

Currently I do experiments in smaller examples (e.g. [`tech-demo-python-packaging`](https://codeberg.org/buhtz/tech-demo-python-packaging) and [`bit_demo`](https://codeberg.org/buhtz/bit_demo)) to figuring out problems and solutions. Some decision's need to be made in the maintenance team. But the time has not yet come. There is no schedule or idea when to "execute" this migration.

**EDIT**: My initial idea was to do the migrate late in the future after we cleaned up the BIT code and solved some Issues. But as more as I work with BIT I get the idea to migrate as soon as possible. It will solve some Issues and will save resources. The latter is because when modifying code before the migration, the risk is high that we need to touch that code again after the migration again.

# Blocked-by
- #2424 
- and several more

buhtz:
I modified the priority and milestone to increase urgency.

I can not modify the codebase without being covered by unit tests (in their strict meaning). But I will not waste resources in improving the test suite while the make file system is used. This is because I have to re-arrange the test suite again after migration to a Python build-system.

In the end: The migration is necessary to start everything else.

[...]

PR #1868, opened 2024-09-09
chore: Enable full linting on specific files

buhtz:
Introduce full linting without any exceptions for a specific (and small) list of files. This increase the pressure on our code quality.

- The two files `common/test/test_lint.py` and `qt/test/test_lint.py` now have a module variable `full_test_files` containing a list of py file names that will get the full linter test battery.
- All other source files are only receiving the PyLint run with an exclusive and minimal list of rules only (`test050_pylint_exclusive_ruleset()`).
- Full linting means that (minimum) the default rule set of each linter is used without excluding any rules.
- Beside previously used PyLint now a chain of linters is used. The unit test method names have numbers to fix the execution order of the linters. 1) `ruff` very fast 2) `pycodestyle` quit fast 3) `pylint` slow but pedantic.
- The reason for having a chain of linters is they do have exceptional rules and capabilities.
- Much files are modified in this PR. I tried to fix some linter errors in them. I wasn't successful on all of them.

I do use a similar setup on two of my own smaller projects.
The motivation behind it was that I will introduce some new files (from splitting existing files) in the next weeks and I want to have them standard conform from the start.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,11 @@
+# SPDX-FileCopyrightText: © 2008 Back In Time Team
+#
+# SPDX-License-Identifier: CC0 (public domain)
+#
+# This file is part of the program "Back In Time" which is released under GNU
+# General Public License v2 (GPLv2).
+# See file LICENSE or go to <https://www.gnu.org/licenses/#GPL>.
+#
 # TravisCI (https://travis-ci.org) configuration file
 os: linux
  
@@ -42,7 +50,7 @@ jobs:
 
 install:
   - pip install -U pip
-  - pip install pylint pyfakefs keyring
+  - pip install pylint ruff pycodestyle pyfakefs keyring
   - pip install pyqt6 dbus-python
   # add ssh public / private key pair to ensure user can start ssh session to localhost for tests
   - ssh-keygen -b 2048 -t rsa -f /home/travis/.ssh/id_rsa -N ""
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,10 @@
 # SPDX-FileCopyrightText: © 2008 Back In Time Team
 #
-# SPDX-License-Identifier: CC0 (public domain)
+# SPDX-License-Identifier: CC0-1.0
 #
-# This file is part of the program "Back In Time" which is released under GNU
-# General Public License v2 (GPLv2).
-# See file LICENSE or go to <https://www.gnu.org/licenses/#GPL>.
+# This file is released under Creative Commons Zero 1.0 (CC0-1.0) and part of
+# the program "Back In Time". The program as a whole is released under GNU
+# General Public License v2 or any later version (GPL-2.0-or-later).
 #
 # TravisCI (https://travis-ci.org) configuration file
 os: linux
--- a/common/test/test_lint.py
+++ b/common/test/test_lint.py
@@ -1,22 +1,104 @@
+# SPDX-FileCopyrightText: © 2023 Christian BUHTZ <c.buhtz@posteo.jp>
+#
+# SPDX-License-Identifier: GPL-2.0-or-later
+#
+# This file is part of the program "Back In Time" which is released under GNU
+# General Public License v2 (GPLv2).
+# See file LICENSE or go to <https://www.gnu.org/licenses/#GPL>.
+"""Tests using several linters.
+
+Instead of failing tests are skipped on machines where the linters are not
+available. As an exception only on TravisCI tests are forced to run and will
+fail if linters not available.
+"""
 import unittest
 import os
 import pathlib
 import subprocess
 import shutil
 from typing import Iterable
+try:
+    import pycodestyle
+    PYCODESTYLE_AVAILABLE = True
+except ImportError:
+    PYCODESTYLE_AVAILABLE = False
+
+BASE_REASON = ('Using package {0} is mandatory on TravisCI, on '
+               'other systems it runs only if `{0}` is available.')
 
 ON_TRAVIS = os.environ.get('TRAVIS', '') == 'true'
-PYLINT_AVIALBE = not shutil.which('pylint') is None
-PYLINT_REASON = ('Using PyLint is mandatory on TravisCI, on other systems'
-                 'it runs only if `pylint` is available.')
+TRAVIS_REASON = ('Running linter tests is mandatory on TravisCI only. On '
+                 'other machines they will run only if linters available.')
+
+PYLINT_AVAILABLE = shutil.which('pylint') is not None
+RUFF_AVAILABLE = shutil.which('ruff') is not None
+
+ANY_LINTER_AVAILABLE = any((
+    PYLINT_AVAILABLE,
+    RUFF_AVAILABLE,
+    PYCODESTYLE_AVAILABLE,
+))
+
+# Files in this lists will get the full battery of linters and rule sets.
+full_test_files = [pathlib.Path(fp) for fp in (
+    'bitbase.py',
+    'schedule.py',
+    'version.py',
+    'test/test_lint.py',
+)]
+
+# Not all linters do respect PEP8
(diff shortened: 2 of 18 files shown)


.travis.yml
```diff
@@ -42,7 +50,7 @@ jobs:
 
 install:
   - pip install -U pip
-  - pip install pylint pyfakefs keyring
+  - pip install pylint ruff pycodestyle pyfakefs keyring
```
daviewales, line 53: `ruff` claims to be a drop-in replacement for `pycodestyle`:

https://docs.astral.sh/ruff/linter/#rule-selection

Also, the [pycodestyle repository](https://github.com/PyCQA/pydocstyle) states that pycodestyle is deprecated, and that users should switch to `ruff`.

What's the reason for including pycodestyle?

common/test/test_lint.py
```diff
@@ -1,22 +1,104 @@
...
+# Files in this lists will get the full battery of linters and rule sets.
+full_test_files = [pathlib.Path(fp) for fp in (
+    'bitbase.py',
+    'schedule.py',
+    'version.py',
+    'test/test_lint.py',
+)]
 
+# Not all linters do respect PEP8
+PEP8_MAX_LINE_LENGTH = 79
```
daviewales, line 51: Consider setting [max line length](https://pylint.readthedocs.io/en/latest/user_guide/messages/convention/line-too-long.html) in a top-level config file for the respective linter.
For example, for pylint, you could have a config file such as: `pylintrc` at the top level containing the following:

``` ini
[MAIN]
max-line-length=79
```

Using the standard config allows the correct configuration to be used when running the linters directly.

Also, the other linters such as `ruff` [also check line length](https://docs.astral.sh/ruff/rules/line-too-long/), so you might consider disabling the line length check in all linters but one.

common/test/test_lint.py
```diff
@@ -1,22 +1,104 @@
...
+        '--score=n',
+        # prevent false-positive no-module-member errors
+        '--extension-pkg-allow-list=PyQt6,PyQt6.QtCore',
+        # Because of globally installed GNU gettext functions
+        '--additional-builtins=_,ngettext',
+        # PEP8 conform line length (see PyLint Issue #3078)
+        '--max-line-length=79',
+        # Whitelist variable names
+        '--good-names=idx,fp',
+        # '--reports=yes',
```
daviewales: I think these options can all be set in a config file as described above.

common/test/test_lint.py
```diff
@@ -37,43 +119,106 @@ def _collect_py_files(self) -> Iterable[pathlib.Path]:
         path = path.parent
 
         # Find recursive all py-files.
-        return path.rglob('**/*.py')
+        py_files = path.rglob('**/*.py')
```
daviewales: I think this should either be:

```suggestion
        py_files = path.glob('**/*.py')
```

OR

```suggestion
        py_files = path.rglob('*.py')
```

See the docs for [rglob](https://docs.python.org/3/library/pathlib.html#pathlib.Path.rglob), which state:

> This is like calling [Path.glob()](https://docs.python.org/3/library/pathlib.html#pathlib.Path.glob) with "`**/`" added in front of the pattern

common/test/test_lint.py
```diff
@@ -1,22 +1,104 @@
...
+        # autodetec number of parallel jobs
+        '--jobs=0',
+        # Disable scoring  ("Your code has been rated at xx/10")
+        '--score=n',
+        # prevent false-positive no-module-member errors
+        '--extension-pkg-allow-list=PyQt6,PyQt6.QtCore',
+        # Because of globally installed GNU gettext functions
+        '--additional-builtins=_,ngettext',
+        # PEP8 conform line length (see PyLint Issue #3078)
+        '--max-line-length=79',
```
daviewales: If you keep the options here in the Python script, you'll want to use the variable defined above for `max-line-length`:
```suggestion
        f'--max-line-length={PEP8_MAX_LINE_LENGTH}',
```

common/test/test_lint.py
```diff
@@ -37,43 +119,106 @@ def _collect_py_files(self) -> Iterable[pathlib.Path]:
...
-            '--max-line-length=79',
-            # Whitelist variable names
-            '--good-names=idx,fp',
-            # '--reports=yes',
+            '--config', 'builtins=["_", "ngettext"]',
+            # Ruff counting branches different from PyLint.
+            # See: <https://www.reddit.com/r/learnpython/comments/
+            #      1buojae/comment/kxu0mp3>
+            '--config', 'pylint.max-branches=13',
+            '--quiet',
```
daviewales, line 151: Should all these options be moved to the config file?

qt/schedulewidget.py
```diff
@@ -205,7 +206,7 @@ def _repeated_unit_combobox(self):
 
         return combobox.BitComboBox(self, repeatedly_units)
 
-    def _slot_schedule_mode_changed(self, idx):
+    def _slot_schedule_mode_changed(self, _idx):
```
daviewales, line 209: Is there a reason we can't remove the `idx` argument altogether? I guess the calling code expects to pass an argument?

qt/test/test_lint.py
```diff
@@ -30,50 +110,114 @@ def _collect_py_files(self) -> Iterable[pathlib.Path]:
...
-                               'inside the test folder but current folder '
+                               'inside the test folder but the current folder '
                                f'is {path}.')
 
         # Workaround
         path = path.parent
 
         # Find recursive all py-files.
-        return path.rglob('**/*.py')
+        py_files = path.rglob('**/*.py')
```
daviewales: Same as above, this should be either:

```suggestion
        py_files = path.glob('**/*.py')
```

OR

```suggestion
        py_files = path.rglob('*.py')
```

qt/test/test_lint.py
```diff
@@ -1,22 +1,102 @@
...
+        # autodetec number of parallel jobs
+        '--jobs=0',
+        # Disable scoring  ("Your code has been rated at xx/10")
+        '--score=n',
+        # prevent false-positive no-module-member errors
+        '--extension-pkg-allow-list=PyQt6,PyQt6.QtCore',
+        # Because of globally installed GNU gettext functions
+        '--additional-builtins=_,ngettext',
+        # PEP8 conform line length (see PyLint Issue #3078)
+        '--max-line-length=79',
```
daviewales: Same as above, use `PEP8_MAX_LINE_LENGTH`.

.travis.yml
buhtz: It claims, but it doesn't fulfill this goal yet. Somewhere in there docu is kind of a status page reflecting this situation. I often have the situation that the ruff stage is green but pycodestyle is not.

common/test/test_lint.py
buhtz: That constant is for ruff and pylint because ruff do not follow PEP8 in all aspects. It checks the line length with a default value different from 79. ruff is round about 88, because they will follow "black" (:imp:) and Pylint is way above that value.

A global config file, a pyproject.toml, is the goal. But before having this we need to fix #1575 . Other config files different from pyproject.toml are problematic because we would need several of such files, each for every linter/tool. And other tools will be added to the current chain. Currently we only of .codespellrc file in the repos root folder for the reason that codespell need to scan all files and not only the packages files (in common and qt). But I would like to have the config centralized. Having the config in the test_lint.py file is kind of centralized and of course a workaround. It is IMHO easier to manage this way until #1575 is fixed.

Even having two test_lint.py is an interim solution that will be fixed when #1575 is done.

qt/schedulewidget.py
buhtz: Qt calls this slot (a signal handler) and pass that argument.

common/test/test_lint.py
daviewales: Understood.

PR #1882, opened 2024-09-23
Add AI-prohibition to PR template

buhtz:


daviewales:
Might be good to add an example showing the difference between using AI as a tool, and 'AI generated content'.

buhtz:
> Might be good to add an example showing the difference between using AI as a tool, and 'AI generated content'.

OK, I removed the tool part from the text. I don't want to blow up the text. The message should be that accounts contributing AI-only content will be abused and most of the time deleted by Microsoft GitHub. I don't want to lay out details rules.

daviewales:
Sounds good. Just an English language note: When you say that accounts will be 'abused', it sounds like _you_ will do a bad thing, but I think you mean 'reported for abuse', which means that someone else has done a bad thing.

PR #1896, opened 2024-10-05
chore: Refactor and full linting ssh-max-arg

buhtz:
Just minor refactoring of the ssh-max-arg script.

EDIT: It seems that the "Files changed" view of Microsoft GitHub is buggy and does not show the modifications in the ssh_max_arg.py script just because I renamed the file. On local git the diff is much more verbose and detailed.


daviewales:
I had to use the following to make the diff show a file move rather than delete and create:

    git diff dev refactor/sshmaxargs -M20 -- common/sshMaxArg.py common/ssh_max_arg.py

It's often good to do file moves in a stand-alone commit with no other changes, as git detects file moves based on similarity of contents.

common/ssh_max_arg.py
```diff
@@ -0,0 +1,169 @@
...
+        size_offset: int = _INITIAL_SSH_CMD_SIZE) -> int:
+    """Determine the maximum length of SSH commands for the current config
+
+    Try a SSH command with length ``ssh_command_size``. The command is
+    decreased by ``size_offset`` if it was too long or increased if it worked.
+    The function calls itself recursively until it finds the maximum
+    possible length. The offset ``size_offset`` is bisect in each try.
+
+    Args:
+        config Back In Time config instance including the details about the
```
daviewales: Missing a colon (`:`) after `config`.
```suggestion
        config: Back In Time config instance including the details about the
```

common/ssh_max_arg.py
```diff
@@ -0,0 +1,169 @@
...
+        nice=False,
+        ionice=False,
+        prefix=False)
+
+    try:
+        with subprocess.Popen(ssh,
+                              stdout=subprocess.PIPE,
+                              stderr=subprocess.PIPE,
+                              universal_newlines=True) as proc:
+            out, err = proc.communicate()
```
daviewales: Consider using `subprocess.run`:

```suggestion
        proc = subprocess.run(capture_output=True, text=True)
        out, err = proc.stdout, proc.stderr
```

Details from [subprocess docs](https://docs.python.org/3/library/subprocess.html):

- The recommended approach to invoking subprocesses is to use the [run()](https://docs.python.org/3/library/subprocess.html#subprocess.run) function for all use cases it can handle.
- If _capture_output_ is true, stdout and stderr will be captured. When used, the internal [Popen](https://docs.python.org/3/library/subprocess.html#subprocess.Popen) object is automatically created with stdout and stderr both set to [PIPE](https://docs.python.org/3/library/subprocess.html#subprocess.PIPE).
- If ... _text_ is true, file objects for stdin, stdout and stderr are opened in text mode using the specified encoding and errors or the [io.TextIOWrapper](https://docs.python.org/3/library/io.html#io.TextIOWrapper) default. The _universal_newlines_ argument is equivalent to _text_ and is provided for backwards compatibility.

buhtz:
> I had to use the following to make the diff show a file move rather than delete and create:

On my local machine diff shows it correct without problem. I am assuming it is just a bug in the Microsoft GitHub web-frontend.

common/ssh_max_arg.py
```diff
@@ -63,20 +63,19 @@ def probe_max_ssh_command_size(
...
     try:
-        with subprocess.Popen(ssh,
-                              stdout=subprocess.PIPE,
-                              stderr=subprocess.PIPE,
-                              universal_newlines=True) as proc:
-            out, err = proc.communicate()
+        proc = subprocess.run(ssh, capture_output=True, text=True, check=True)
+        out, err = proc.stdout, proc.stderr
+        print(out)
+        print(err)
```
daviewales: Should this be `logging` rather than `print`?

buhtz: Just stupid debug code. My fault.

common/ssh_max_arg.py
```diff
@@ -0,0 +1,168 @@
...
+        cmd=['printf', command_string],
+        nice=False,
+        ionice=False,
+        prefix=False)
+
+    try:
+        proc = subprocess.run(ssh, capture_output=True, text=True, check=True)
+        out, err = proc.stdout, proc.stderr
+        print(out)
+        print(err)
```
buhtz:
```suggestion
```

buhtz:
I added the `LICENSES` folder conforming to SPDX/REUSE standard.
The file `LICENSE` will stay until all files/folders do contain SPDX meta data and `$ reuse lint` is satisfied.

PR #1921, opened 2024-11-07
fix: remove unused variables and clean up code

ivanrdgz03:
Removed several unused variables to improve code readability and maintain a cleaner codebase. This change resolves the linting issues related to unused variables mentioned in issue #1755.


buhtz:
Hello Iván,

Your PR will be reviewed as soon as possible.

Best,
Christian

common/encfstools.py
```diff
@@ -511,7 +511,7 @@ def __init__(self, cfg, string = True):
             .replace('Schnappschuss', '(?:Schnappschuss|Snapshot)')
 
         #precompile some regular expressions
-        host, port, user, path, cipher = cfg.sshHostUserPortPathCipher()
+        host, user, path = cfg.sshHostUserPortPathCipher()
```
buhtz: This won't work. Have look into `sshHostUserPortPathCipher()` and what it returns. It still returns a tuple with 5 values, not 3.

In cases like this it is usual to use `_` as a variable name to indicate that this value will be ignored and not used. **But** in our case the name `_` is reserved by the function `_()` and used for translating strings with the GNU Gettext toolkit. Because of that it is an alternative to use the `_` plus a name like this.

```suggestion
        host, _post, user, path, _cipher = cfg.sshHostUserPortPathCipher()
```

common/snapshots.py
```diff
@@ -844,15 +844,14 @@ def backup(self, force=False):
                                 'Cannot start snapshot yet: target directory '
                                 'not accessible. Will retry each second in '
                                 'the next 30 seconds. Please wait.')
-
                             counter = 0
```
buhtz: Your solution works but I would suggest something else. Please also see the next two comments/suggestion.

```suggestion
```

common/snapshots.py
```diff
@@ -844,15 +844,14 @@ def backup(self, force=False):
...
-                            for counter in range(0, 30):
+                            while counter < 30:
```
buhtz:
```suggestion
                            for _ in range(30):
```

common/snapshots.py
```diff
@@ -844,15 +844,14 @@ def backup(self, force=False):
...
 
                                 time.sleep(1)
 
                                 if self.config.canBackup():
                                     break
-
+                                counter += 1
```
buhtz:
```suggestion
```

buhtz (changes requested):
Hello Iván,
thank you again for your contribution.
I added some comments and suggestions for improvements.

Do you see the "commit suggestions" button in that comments?

I scheduled your PR for the 2nd release from now. This is we are in the middle of testing the release candidate for version 1.5.3. I might release it at the weekend. After it we do prepare for version 1.6.0 where your PR will get merged into.

Thanks in advance,
Christian

ivanrdgz03:
Thanks, I knew I could use _ but it gave me errors because it was used previously, I didn't know I could also use _port for example. I hope I can be of help in future versions of backintime.

common/snapshots.py
```diff
@@ -844,15 +844,12 @@ def backup(self, force=False):
                                 'Cannot start snapshot yet: target directory '
                                 'not accessible. Will retry each second in '
                                 'the next 30 seconds. Please wait.')
-
-                            counter = 0
-                            for counter in range(0, 30):
+                            for _ in range(30):
```
buhtz: Stupid me! Even if I explained it to you I made the same mistake. That is why TravisCI failes.

```suggestion
                            for _idx in range(30):
```

buhtz (changes requested):

common/snapshots.py
ivanrdgz03: Don't worry!! I noticed and just changed it

buhtz (approved):

buhtz:
Hey Iván, congrats to your contribution and welcome.

If you are looking for something else to do. You might could take a look at the current [Release Candidate](https://github.com/bit-team/backintime/releases/tag/v1.5.3-rc1) and give it a [manual test](https://github.com/bit-team/backintime/blob/dev/doc/maintain/BiT_release_process.md#manual-testing---recommendations). I know its boring but could really help.

PR #1937, opened 2024-11-21
fix: Copy symlink as symlink in snapshot compare

sevens:
When comparing snapshots, copy symlinks in the snapshots as symlinks, not as
their target.  The old behavior made the comparison not fully compare the actual
snapshot data and, even worse, could fill up your entire disk (when having
circular symlinks, e.g. a symlink to `/`, which copies `/` and eventually
copies the original symlink which points to `/` which ...).

Additionally fixes a crash when the compared snapshot(s) contain a symlink
pointing to a nonexistent target.

Fixes #1902.

Issue #1902, opened 2024-10-12
Comparing 2 snapshots can fill your disk if it includes a symlink to e.g. `/` (doesn't copy symlink as symlink)


sevens:
First of all thanks for the software, have been using it quite a bit through the years!

I was trying to compare 2 snapshots (one of <30GB with one of a few 100MB) and suddenly noticed my system acting weird due to what turned out to be my disk being full, even though nearly 1TB of 1.8TB was free beforehand...

Long story short, it turned out that symlinks in the snapshots were having their targets copied as a regular dir (instead of being copied as symlinks themselves) to the temp location BiT uses for comparing.  The symlink's target included another symlink somewhere within, this one pointing to `/` (Proton install from Steam which has a `z:` drive/symlink pointing to `/`).  That resulted in an infinite dir structure being copied: `/` includes `~/.steam/root/steamapps/compatdata/ID/pfx/dosdevices/z:` which points to `/` which includes `~/.steam/root/steamapps/compatdata/ID/pfx/dosdevices/z:` which points to ...

After digging into the source a bit it seems that BiT creates a copy of the snapshots its comparing ( https://github.com/bit-team/backintime/blob/11006532ee983e96632b0573c8f707329e9802da/qt/snapshotsdialog.py#L370 and line 372).  This uses Python's `shutil.copytree` ( https://github.com/bit-team/backintime/blob/11006532ee983e96632b0573c8f707329e9802da/qt/app.py#L1700 ) which by default (and in the form BiT calls it) doesn't copy symlinks as symlinks but copies their targets:

> If `symlinks` is true, symbolic links in the source tree are represented as symbolic links in the new tree and the metadata of the original links will be copied as far as the platform allows; if false or omitted, the contents and metadata of the linked files are copied to the new tree.  [Default is `False`]
> -- https://docs.python.org/3/library/shutil.html#shutil.copytree

presumably causing the issue.

Easy reproduction of the base issue:
  * Create 2 snapshots, at least one of them containing a symlink
  * Compare the 2 snapshots
  * Check the temp dir of the snapshots and notice there is no symlink but the target is included instead
    * E.g.:
      
       ```sh
       $ ls -l ~/backup_source/
       total 0
       lrwxrwxrwx 1 USER users 3 10-12 15:16 bar -> foo
       -rw-r--r-- 1 USER users 0 10-12 15:48 baz
       -rw-r--r-- 1 USER users 8 10-12 15:48 foo

[...]

PR #1940, opened 2024-11-21
feat: Prefix temp snapshot copies with `backintime_`

sevens:
I checked if it was easily possible to add a more specific prefix (e.g. `backintime-snapshot_compare_` for when comparing snapshots) but the function gets used in 3 places (2 through calls to `app.MainWindow.openPath()`), all quite different use cases.  Couldn't find a nice way to pass more specific info to `tmpCopy()` from them (don't know the code well enough to decide on a clean and good way to do this from all places, if it's even desirable at all).

Don't think this warrants a `CHANGES` entry?!  Only case I can think of is if there are people who have e.g. some `/tmp` (and `/var/tmp`) cleanup scripts that assume the current format, but AFAICT it's nigh impossible to reliably determine if an old-format created dir is from BiT or not so that seems unlikely to exist to me.

Tested:
  * Compare 2 snapshots, it created temp dirs like `/tmp/backintime_yvvem486_20241119-165907-278` (last 3 parts being snapshot id)
  * `btnOpenCurrentItemClicked()`:
    * Navigate to file in snapshot that doesn't exist anymore on the local storage/source
    * Click `Alt+Down` (seems to be the only way to trigger this function)
    * Created temp dir: `/tmp/backintime_esh5heni_20241121-112648-601/`
  * `filesViewItemActivated()`:
    * Navigate to file in snapshot that doesn't exist anymore on the local storage/source
    * Double-click on said file
    * Created temp dir: `/tmp/backintime_ge543330_20241121-112648-601/`

There don't seem to be any relevant unittests (same as for #1937).

See also: https://github.com/bit-team/backintime/issues/1902#issuecomment-2491088686 and https://github.com/bit-team/backintime/issues/1902#issuecomment-2491177924

### DIFF
--- a/common/doc-dev/modules.rst
+++ b/common/doc-dev/modules.rst
@@ -25,6 +25,6 @@ common
    schedule
    snapshotlog
    snapshots
-   sshMaxArg
+   ssh_max_arg
    sshtools
    tools
--- a/common/doc-dev/sshMaxArg.rst
+++ b/common/doc-dev/sshMaxArg.rst
@@ -1,7 +1,0 @@
-sshMaxArg module
-================
-
-.. automodule:: sshMaxArg
-    :members:
-    :undoc-members:
-    :show-inheritance:
--- a/common/doc-dev/ssh_max_arg.rst
+++ b/common/doc-dev/ssh_max_arg.rst
@@ -1,0 +1,7 @@
+ssh_max_arg module
+==================
+
+.. automodule:: ssh_max_arg
+    :members:
+    :undoc-members:
+    :show-inheritance:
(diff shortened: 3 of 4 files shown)


sevens:
I also tested changing the suffix code to:

  ```py
  sid = f'_profile_{sid.profileID}_snapshot_{sid.sid}'
  ```

but not sure if that's too verbose.  Left it out of this PR for now, let me know if it should be added back...  Example dir with this change: `/tmp/backintime_ed3caw6l_profile_1_snapshot_20241119-170042-790`

buhtz:
`Alt+Down` is wired. 🤣 But won't change it today.
Thank you for testing and analyzing.

Ready to merge.

PR #1992, opened 2025-01-02
feat(gui): Toolbar context menu to display the buttons as or with text

### LINKED ISSUE #1105: Toolbar has no option to display items as text
The toolbar in Back In Time shows icons only. While tooltips appear after hovering on them, there appears to be no option to display the tools with text (or as text-only).

Compare to Kate text editor or Okular image viewer. Their toolbars provide options (available via right-click menu) to display entries as: Icons only, Text only, Text Alongside Icons, and Text Under icons.

Thanks for your consideration!

Version: 1.2.1
Source: Kubuntu 20.04 repositories

s4moore:
Added a context menu when right clicking on the tool tray which gives the option to customize the dsiplay of the icons by choosing between text, icon, text alongside icon, and text below icon

Close #1105

### DIFF
--- a/qt/app.py
+++ b/qt/app.py
@@ -742,10 +742,10 @@ class MainWindow(QMainWindow):
         """
         context_menu = QMenu(self)
         options = (
-            (_('Icon Only'), Qt.ToolButtonStyle.ToolButtonIconOnly),
-            (_('Text Only'), Qt.ToolButtonStyle.ToolButtonTextOnly),
-            (_('Text Under Icon'), Qt.ToolButtonStyle.ToolButtonTextUnderIcon),
-            (_('Text Beside Icon'), Qt.ToolButtonStyle.ToolButtonTextBesideIcon),
+            (_('Icons only'), Qt.ToolButtonStyle.ToolButtonIconOnly),
+            (_('Text only'), Qt.ToolButtonStyle.ToolButtonTextOnly),
+            (_('Text below icons'), Qt.ToolButtonStyle.ToolButtonTextUnderIcon),
+            (_('Text beside icon'), Qt.ToolButtonStyle.ToolButtonTextBesideIcon),
         )
         for text, style in options:
             text = '✔ ' + text if toolbar.toolButtonStyle() == style else '  ' + text
--- a/qt/app.py
+++ b/qt/app.py
@@ -3,6 +3,7 @@
 # SPDX-FileCopyrightText: © 2008-2022 Richard Bailey
 # SPDX-FileCopyrightText: © 2008-2022 Germar Reitze
 # SPDX-FileCopyrightText: © 2024 Christian Buhtz <c.buhtz@posteo.jp>
+# SPDX-FileCopyrightText: © 2025 moresamuel
 #
 # SPDX-License-Identifier: GPL-2.0-or-later
 #
@@ -737,27 +738,38 @@ class MainWindow(QMainWindow):
         restore.insertSeparator(self.act_restore_parent)
         restore.setToolTipsVisible(True)
 
-    def style_setter(self, point, toolbar):
+    def _context_menu_toolbar_style(self,
+                                    point: QPoint,
+                                    toolbar: QToolBar) -> None:
+        """Open a context menu to modify styling of tooblar buttons
+        buttons.
         """
-        Create a context menu for changing the style of the icons
-        """
-        context_menu = QMenu(self)
+        menu = QMenu(self)
         group = QActionGroup(self)
         options = (
-            (_('Icons only'), Qt.ToolButtonStyle.ToolButtonIconOnly),
-            (_('Text only'), Qt.ToolButtonStyle.ToolButtonTextOnly),
-            (_('Text below icons'), Qt.ToolButtonStyle.ToolButtonTextUnderIcon),
-            (_('Text beside icon'), Qt.ToolButtonStyle.ToolButtonTextBesideIcon),
+            (
+                _('Icons only'),
+                Qt.ToolButtonStyle.ToolButtonIconOnly),
+            (
+                _('Text only'),
+                Qt.ToolButtonStyle.ToolButtonTextOnly),
+            (
+                _('Text below icons'),
+                Qt.ToolButtonStyle.ToolButtonTextUnderIcon),
+            (
+                _('Text beside icon'),
+                Qt.ToolButtonStyle.ToolButtonTextBesideIcon),
         )
         for text, style in options:
             action = QAction(text, self)
             action.setCheckable(True)
             action.setChecked(toolbar.toolButtonStyle() == style)
             group.addAction(action)
-            action.triggered.connect(lambda _, s=style: toolbar.setToolButtonStyle(s))
+            action.triggered.connect(
+                lambda _, s=style: toolbar.setToolButtonStyle(s))
 
-        context_menu.addActions(group.actions())
-        context_menu.exec(toolbar.mapToGlobal(point))
+        menu.addActions(group.actions())
+        menu.exec(toolbar.mapToGlobal(point))
 
     def _create_main_toolbar(self):
         """Create the main toolbar and connect it to actions."""
@@ -765,9 +777,10 @@ class MainWindow(QMainWindow):
         toolbar = self.addToolBar('main')
         toolbar.setFloatable(False)
 
-        # create context menu for right click
+        # Context menu to modify button styling
         toolbar.setContextMenuPolicy(Qt.ContextMenuPolicy.CustomContextMenu)
-        toolbar.customContextMenuRequested.connect(lambda point: self.style_setter(point, toolbar))
+        toolbar.customContextMenuRequested.connect(
+            lambda point: self._context_menu_toolbar_style(point, toolbar))
 
         # Drop-Down: Profiles
         self.comboProfiles = qttools.ProfileCombo(self)
--- a/qt/app.py
+++ b/qt/app.py
@@ -32,7 +32,7 @@ import tools
 tools.initiate_translation(None)
 
 import qttools
-from textwrap import TextWrapper
+from textwrap import fill
 
 import backintime
 import tools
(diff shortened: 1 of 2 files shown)


buhtz:
Great. I will look later deeper into it. Do you know the application "peek"? With that you can create screencast, GIF-animated Screenshots of the Gui. 

Here is an example:
Peek 2025-01-03 11-09

qt/app.py
```diff
@@ -735,12 +735,57 @@ def _create_menubar(self):
         restore.insertSeparator(self.act_restore_parent)
         restore.setToolTipsVisible(True)
 
+    def wrap_text(self, text, max_len):
```
buhtz: There is no need for this function. Python offers the `textwrap` package for this. Have look into `qttools.set_wrapped_tooltip()` how we are using it.

qt/app.py
```diff
@@ -735,12 +735,57 @@ def _create_menubar(self):
...
+                line = [word]
+        if line:
+            lines.append(" ".join(line))
+        return '\n'.join(lines)
+
+    def style_setter(self, point, toolbar):
+        """
+        Create a context menu for changing the style of the icons
+        """
+        context_menu = QMenu(self)
```
buhtz: I would like to have an indicator (e.g. checkmark) in the context menu. This way it would be obvious to the user which option is currently active.

e.g. something like this

image

buhtz (changes requested):
Hello Samuel,
great work. Thank you for this. The text wrapping is a good idea.

I added two comments suggesting some modifications.

qt/app.py
s4moore: I couldn't get this working properly but have used the TextWrapper class from the provided textwrap.py and removed the custom function. I also couldn't get peek working as I'm using WSL (been trying to solve for a couple of hours now :/ )

qt/app.py
s4moore: done, let me know if you're happy with the ✔ or prefer something else

s4moore:
Just realised I said in my commit message that the textwrapper was from qttools but I think that's a mistake..

Also, I added the _() around the icon text as it seems that this is how you mark the text for translation?

qt/app.py
```diff
@@ -32,7 +32,7 @@
 tools.initiate_translation(None)
 
 import qttools
-from textwrap import TextWrapper
+from textwrap import fill
```
buhtz:
```suggestion
import textwrap
```

qt/app.py
```diff
@@ -799,7 +799,7 @@ def _create_main_toolbar(self):
 
                 toolbar.widgetForAction(act).setToolTip(button_tip)
 
-            act.setText(TextWrapper(width=8, break_long_words=False).fill(act.text()))
+            act.setText(fill(act.text(), width=8, break_long_words=False))
```
buhtz: The term `fill` is to broad. I would like to keep the package name to make the context clear to the readers.
```suggestion
            act.setText(textwrap.fill(act.text(), width=8, break_long_words=False))
```

buhtz (changes requested):

buhtz:
> Also, I added the _() around the icon text as it seems that this is how you mark the text for translation?

Yes, this is correct.

qt/app.py
buhtz: You can run Back In Time with Microsoft Windows WSL? Just the command line version "backintime" or "backintime-qt", too?

Anyway: What do you mean by "not properly". Please describe the goal you want to achieve with text wrapping. I might be able to help.

qt/app.py
```diff
@@ -735,12 +736,34 @@ def _create_menubar(self):
...
+        """
+        context_menu = QMenu(self)
+        options = (
+            (_('Icon Only'), Qt.ToolButtonStyle.ToolButtonIconOnly),
+            (_('Text Only'), Qt.ToolButtonStyle.ToolButtonTextOnly),
+            (_('Text Under Icon'), Qt.ToolButtonStyle.ToolButtonTextUnderIcon),
+            (_('Text Beside Icon'), Qt.ToolButtonStyle.ToolButtonTextBesideIcon),
+        )
+        for text, style in options:
+            text = '✔ ' + text if toolbar.toolButtonStyle() == style else '  ' + text
```
buhtz: I have to 😄 . I like this solution because it is so simple and easy. Very pythonic.
But I am assuming the visual appearance is unusual regarding the experience of our users.

I had something like the `QActionGroup` in mind.
image

See [PyQt6 docu](https://www.riverbankcomputing.com/static/Docs/PyQt6/api/qtgui/qactiongroup.html#QActionGroup) (without pictures) and original [C++ Qt docu](https://doc.qt.io/qt-5/qactiongroup.html) (with pictures).

buhtz (changes requested):

qt/app.py
```diff
@@ -735,12 +736,34 @@ def _create_menubar(self):
...
+    def style_setter(self, point, toolbar):
+        """
+        Create a context menu for changing the style of the icons
+        """
+        context_menu = QMenu(self)
+        options = (
+            (_('Icon Only'), Qt.ToolButtonStyle.ToolButtonIconOnly),
+            (_('Text Only'), Qt.ToolButtonStyle.ToolButtonTextOnly),
+            (_('Text Under Icon'), Qt.ToolButtonStyle.ToolButtonTextUnderIcon),
+            (_('Text Beside Icon'), Qt.ToolButtonStyle.ToolButtonTextBesideIcon),
```
buhtz: I am not a native speaker but this sounds easier to me. What do you think?

```suggestion
            (_('Icons only'), Qt.ToolButtonStyle.ToolButtonIconOnly),
            (_('Text only'), Qt.ToolButtonStyle.ToolButtonTextOnly),
            (_('Text below icons'), Qt.ToolButtonStyle.ToolButtonTextUnderIcon),
            (_('Text beside icon'), Qt.ToolButtonStyle.ToolButtonTextBesideIcon),
```

buhtz (changes requested):

qt/app.py
s4moore: I have been able to get backintime-qt working but do get the following errors:

WARNING: Failed to connect to Udev serviceHelper daemon via D-Bus: org.freedesktop.DBus.Error.Spawn.ChildExited
WARNING: D-Bus message: Launch helper exited with unknown return code 1
WARNING: Udev-based profiles cannot be changed or checked due to Udev serviceHelper connection failure

I've looked into it a bit and it seems to be because WSL doesn't interact with hardware in the same way as native Linux but the program still does backups and seems okay otherwise

buhtz: Wow. Good to know. But as a maintainer I would not recommend doing it that way.

qt/app.py
s4moore: Yes, it reads much better

buhtz:
Just a tip. You can prevent TravisCI from running if you add "[skip ci]" at the end of your commit message. So if you know one commit won't be your last or final one you can skip the CI.

The background is everytime you push your commits to the PR the TravisCI starts running. But the "credits" on this machine are limited for us but I have no elegant solution to switch it of. And there is also the CO2 the machine produces.

qt/app.py
s4moore: Okay, so I should implement the action group right? Regarding the visual appearance, do you mean of the code or the execution? Sorry if the question sounds strange, I'm just not entirely certain what would be best here?

buhtz: Not strange. Don't worry. 😄 I mean the GUI. The code es beautiful. And it will be a bit more uglier when using QActionGroup. But the GUI will look "better" and more usual to our users. From a coding perspective I really like your solution.

s4moore: Okay, I've implemented the QActionGroup (and pushed with [skip ci]). The default icon is a bullet point, should I change to radio buttons or what is your preference?

s4moore: And thanks for the kind words 😄

buhtz (approved):
Congratulations! 🥳  It now looks fine to me and ready to merge. It is our project policy to cool down some days before merging a PR.

- I changed the name of  the method and also made it "private" (using the `_` as prefix).
- I added your nickname in the SPDX header (see line 6 in app.py).
- I added a changelog entry also with your nickname.
- Feel free to add your real name and email (not recommended because of spam) there.
- The styling of the radio buttons is OK that way. I am assuming this depends on the desktop theme used. I am not aware of another application as an example but I hope it is consistent.

Don't forget to update your local branch after my modifications.

A next step (in another PR) would be to make this toolbar style persistent between sessions. But please wait until #1989 is merged. Then you can use the new introduce state file to solve this.

Please let me know if you want to take this next step. But there is no hurry.

s4moore:
It's really exciting that my name will be published, thank you so much! I wanted to change my name in app.py to "Samuel Moore" but not sure if pushing that change from my side would require an unneccesary run of TravisCI?

More than happy to change the style of buttons if you do decide that you would prefer something different, just let me know.

And yes, I will be happy to work on next step 😄

buhtz:
> I wanted to change my name in app.py to "Samuel Moore" but not sure if pushing that change from my side would require an unneccesary run of TravisCI?

Just make sure to do `git fetch --prune --all && git pull`. This will update your local branch with the latest commit I have done. Then you are free to modify the name or anything else and commit (plus `[skip ci]`) and push again into the PR.

> And yes, I will be happy to work on next step 😄

Than just watch #1989 . You can start when it is merged.

buhtz:
Merged #1989

s4moore:
Hi Christian, 

I had some issues updating to the latest version of BIT. When I first tried to use the updated version, ``backintime.json`` was not found. I toucuhed the file to create it which got me further but then I was getting key errors because some of the fields are not present in the template used in ``statedata.py 30 -> 45`` . 

Am I missing a something in set up?

buhtz:
Are you sure that you have the latest "dev" version?
The missing backintime.json should be a WARNING on the first run, without interrupting BIT. BIT then will auto-migrate values from the config file into a new backintime.json.

But even if the backintime.json is empty there shouldn't be a keyerror.

Mhm... I need to think about that.

Please try to reproduce the problem and report the terminal output in a new issue. Delete your backintime.json and show me the error. Touch the backintime.json and show me the error.

s4moore:
Hi Christian,

I have completed a first attempt at saving the icon styling in state data. Should I simply open a pull request?

buhtz:
Yes, present your work in a pull request please.

Issue #1018, opened 2019-07-14
FeatureRequest: Activity log


noMICROSOFTbuhtz:
This was on my own todo-List but I am far away from implementing this myself. So just for documentation (**not** for hurring it up!) I describe this feature here.

**The problem**
I don't know if my BackInTime (BIT) does a backup job or not even I don't know when it did it.

From time to time I open BIT and open each profile and checking the left list-view with the snapshot dates. Also for this I have to boot my backup-NAS because BIT won't show my anything without a backup destination available.

**Solution**
I want to aks BIT with one click or one command about the last successfull backup of **all** exiting jobs. There could be a `.get_last_snapshot_dates()` resulting in this

```
{'music_backup': time_struct<>,
 'other_backup_profile': time_struct<>}

```

This could be used by another methode pretty printing it to `stdout` (for command-line access) or by the GUI or by the plugin-API. There could be plugins showing the result in systray or sending an email about it or generating an RSS feed or make a robot dance the result.

buhtzz:
@emtiu My first impulse was to ask for "still relevant?" and tag this. Then I saw my own face.  😆 

This feature is still in my mind. Just do something like `backintime --status` and BIT gives you (in json format!) the last dates for each snapshot profile and maybe some more information. You can write your own daily running checker script just to check if one of the profiles get outdated because of unknown and uncaught errors.

Tag: Feature

buhtz:
This feature give users the opportunity to ask BIT (on command line) about the status of all existing profiles in a machine readable way. This can be used in (e.g. via cron executed) scripts to monitor BIT and start something (e.g. send mail or flash a big red light) when something is wrong.

aryoda:
This feature request looks similar to "log aggregator" functionality.

I think a centralized summary of all result states is very helpful.

Still it requires a functionality in BiT (or outside) to inform the user based on this file and this currently the pain point IMHO.

We have user-callback scripts for that to trigger actions per snapshot directly to

1. send emails in case of problems (I don't like emails for privacy reasons if they contain paths and file names):
      https://github.com/bit-team/user-callback/blob/master/user-callback.sendmail

2.  send a desktop notification (works only on desktop systems I guess):
     https://github.com/bit-team/user-callback/blob/master/user-callback.notify



We could "lift" this functionality into a BiT config + GUI for normal users but configuring a `sendmail` or `smtp` profile could be overkill.

s4moore:
Hello Christian,

[...]

PR #2002, opened 2025-01-11
feat: Persistent icon style in toolbar

s4moore:
Added toolbar_button_style property to the state data so the selected main toolbar icon styling persists between sessions.

### DIFF
--- a/common/statedata.py
+++ b/common/statedata.py
@@ -319,3 +319,17 @@ class StateData(dict, metaclass=singleton.Singleton):
     @mainwindow_second_splitter_widths.setter
     def mainwindow_second_splitter_widths(self, vals: tuple[int, int]) -> None:
         self['gui']['mainwindow']['splitter_second_widths'] = vals
+
+    @property
+    def toolbar_button_style(self) -> int:
+        """Return style of icons for the main toolbar."""
+        try:
+            return self['gui']['mainwindow']['toolbar_button_style']
+        except KeyError:
+            self['gui']['mainwindow']['toolbar_button_style'] = 0
+            return self['gui']['mainwindow']['toolbar_button_style']        
+    
+    @toolbar_button_style.setter
+    def toolbar_button_style(self, value) -> None:
+        self['gui']['mainwindow']['toolbar_button_style'] = value
+        
--- a/common/statedata.py
+++ b/common/statedata.py
@@ -327,9 +327,8 @@ class StateData(dict, metaclass=singleton.Singleton):
             return self['gui']['mainwindow']['toolbar_button_style']
         except KeyError:
             self['gui']['mainwindow']['toolbar_button_style'] = 0
-            return self['gui']['mainwindow']['toolbar_button_style']        
-    
+            return self['gui']['mainwindow']['toolbar_button_style']
+
     @toolbar_button_style.setter
     def toolbar_button_style(self, value) -> None:
         self['gui']['mainwindow']['toolbar_button_style'] = value
-        
--- a/common/statedata.py
+++ b/common/statedata.py
@@ -322,12 +322,16 @@ class StateData(dict, metaclass=singleton.Singleton):
 
     @property
     def toolbar_button_style(self) -> int:
-        """Return style of icons for the main toolbar."""
+        """Style of icons for the main toolbar.
+
+        Returns:
+           Style value as integer (default: 0 as ``ToolButtonIconOnly``)
+        """
         try:
             return self['gui']['mainwindow']['toolbar_button_style']
         except KeyError:
             self['gui']['mainwindow']['toolbar_button_style'] = 0
-            return self['gui']['mainwindow']['toolbar_button_style']
+            return self.toolbar_button_style
 
     @toolbar_button_style.setter
     def toolbar_button_style(self, value) -> None:
--- a/qt/app.py
+++ b/qt/app.py
@@ -752,10 +752,10 @@ class MainWindow(QMainWindow):
                 Qt.ToolButtonStyle.ToolButtonTextBesideIcon),
         )
     
-    def _set_toolbar_button_style(self, toolbar, style, index):
+    def _set_toolbar_button_style(self, toolbar, style):
         """Set toolbar button style and store the selected index."""
         toolbar.setToolButtonStyle(style)
-        self.toolbar_button_style = int(style)
+        self.toolbar_button_style = style.value
 
     def _context_menu_button_style(self,
                                     point: QPoint,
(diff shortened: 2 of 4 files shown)


common/statedata.py
```diff
@@ -319,3 +319,17 @@ def mainwindow_second_splitter_widths(self) -> tuple[int, int]:
...
         self['gui']['mainwindow']['splitter_second_widths'] = vals
+
+    @property
+    def toolbar_button_style(self) -> int:
+        """Return style of icons for the main toolbar."""
+        try:
+            return self['gui']['mainwindow']['toolbar_button_style']
+        except KeyError:
+            self['gui']['mainwindow']['toolbar_button_style'] = 0
+            return self['gui']['mainwindow']['toolbar_button_style']
```
buhtz: I like it how you realized the default-value.

qt/app.py
```diff
@@ -755,15 +751,34 @@ def _context_menu_toolbar_style(self,
                 _('Text beside icon'),
                 Qt.ToolButtonStyle.ToolButtonTextBesideIcon),
         )
-        for text, style in options:
+    
+    def _set_toolbar_button_style(self, toolbar, style, index):
+        """Set toolbar button style and store the selected index."""
+        toolbar.setToolButtonStyle(style)
+        self.toolbar_button_style = index
```
buhtz: The ToolButtonStyle's are just integers. See https://www.riverbankcomputing.com/static/Docs/PyQt6/api/qtcore/qt.html#ToolButtonStyle

`enumerate()` makes it more complex than it need to be. Just cast them to integers before store them in the state file.

```suggestion
    def _set_toolbar_button_style(self, toolbar, style, index):
        """Set toolbar button style and store the selected index."""
        toolbar.setToolButtonStyle(style)
        self.toolbar_button_style = int(style)
```

qt/app.py
```diff
@@ -772,10 +787,14 @@ def _create_main_toolbar(self):
...
-        # Context menu to modify button styling
+        # Context menu to modify button styling for main toolbar
         toolbar.setContextMenuPolicy(Qt.ContextMenuPolicy.CustomContextMenu)
         toolbar.customContextMenuRequested.connect(
-            lambda point: self._context_menu_toolbar_style(point, toolbar))
+            lambda point: self._context_menu_button_style(point, toolbar))
+
+        # Resore button styling for main toolbar
+        toolbar.setToolButtonStyle(
+            self._button_styles()[self.toolbar_button_style][1])
```
buhtz: This might work.
```suggestion
        toolbar.setToolButtonStyle(Qt.ToolButtonStyle(self.toolbar_button_style))
```

buhtz (changes requested):

qt/app.py
```diff
@@ -123,15 +123,18 @@ def __init__(self, config, appInstance, qapp):
         # window icon
         self.qapp.setWindowIcon(icon.BIT_LOGO)
 
+        state_data = StateData()
+
+        # main toolbar icon styles
+        self.toolbar_button_style = state_data.toolbar_button_style
```
buhtz: I don't understand why you create the instance variable `self.toolbar_button_style`.

s4moore: Ah yes, I see now it's not neccessary, I've updated the file.

buhtz (approved):
Hello Samuel,
thank you very much for this contribution. I appreciate that.

Have you installed one of that linters on your system? `ruff`, `pylint` or `flake8`?
I recommend to install all of them and then run the unit tests. The tests also run that linters and you will find some minor warnings related to code styling.

You can also try to add a changelog entry.

buhtz:
I have a question related to the state file in general and not to this PR.

Might it be appropriate to move the state file code into the "qt" part of the code base? As I can see all variables in this file are somehow related to the GUI and not relevant for the command line version of BIT.

Moving it to the GUI would solve two other problems: 1) Saving it at application exit without using "atexit" module but exit handler of Qt and 2) risk to lose state data by overwriting the file from different BIT instances running at the same time.

What do you think?

s4moore:
Regarding the linters, I have started using ruff but need to get into the habit of checking every time. I will start to use flake8 and pylint as well as per your suggestion. Do you have a preferred configuration to share or should I just set them up myself. I assume that with ruff, you're ignoring E: 402, 712, 713, 714, 721, 722, 731, 741, 751; F: 401, 403, 405, 811, 821, 841, is that right? (I'm quite new to linters (and contributing in general) so any suggestions/advice would be appreciated)

Your suggestion to move the state data into the qt code makes sense to me. The only issue I could think of would be if two gui instances were open at once but that seems quite unlikely and is not handled in the current setup anyway, I guess it's much more likely that a user might exit a gui instance while a cron job is running so it seems like a logical step to move it. Do yout want me to work on that? Otherwise I was thinking of working on #1018 (unless there's something else you suggest)?

s4moore:
I've added an entry to the changleog, please let me know if any issues with format / wording.

buhtz:
> Regarding the linters, I have started using ruff 
> [..]
> a preferred configuration 
> [..]

Please see `common/test/test_lint.py` and `qt/test/test_lint.py`. It might be unusual but we do have "unit tests" that do execute the linters (with our preferred configuration) and fail if the linters are not satisfied.

I recommend to stick to the unit tests with running `$ pytest test/test_lint.py`. Using linters the regular way, with their default config and/or integrated in your IDE, won't work well with BIT. The code base of BIT is extrem smelly and will produce thousands of linter errors.

As you can see in the unit tests. The current approach is to use the linters with all warnings disabled and only some of them enabled. See Issue #1755 for details about how we reduce the amount of linter errors step by step. We also have tests using a linter in its default configuration (all warnings enabled) but only for an exclusive set of files.

Let me give you a short overview via looking into `test_lint.py`.

1. `test010_ruff_default_ruleset()` - Default ruff plus some extended rules on an exclusive list of files (see module variable `full_test_files`)
2. `test020_flake8_default_ruleset()` - Same as the previous test but using `flake8`.
3. `test030_pylint_default_ruleset()` - Same as the previous test but using `pylint`.
4. `test050_pylint_exclusive_ruleset()` - PyLint with all warnings disabled (`--disable=all`), then some warnings enabled (`--enable=`, see variable `err_codes`), running on all py files.

The numbers in the method names are there to set the execution order of them. Ruff is very fast and will cover a big amount of possible errors, so use it as first. Next use flake8 to cover errors not caught by Ruff. At the end use PyLint which is extrem slow but also extrem pedantic and covering much more then code styling but also common bad coding habits and other code smells.

> Your suggestion to move the state data into the qt code makes sense to me. [..] Do yout want me to work on that?

Thanks for your thought. I'll work on that.

> I was thinking of working on #1018 (unless there's something else you suggest)?

Looks OK to me.

buhtz (approved):

PR #2003, opened 2025-01-11
fix(test): Remove state file migration tests

### LINKED ISSUE #2001: Unit tests do reset the state file
~~Somehow running BIT via cron affects the state file and reset it.
I need to investigate that.~~

State files not reset but overwritten by a unit test. The PyFakeFS does not work in that test because the `save()` method is triggered via `atexit` module. At that time PyFakeFS is not active anymore.

buhtz:
Maybe don't merge because of
- #2010 

Some unit tests has overwritten the real state file, no matter that PyFakeFS was used.
The fake file system was not active anymore because StateData.save() is triggered at application exit (atexit module) when PyFakeFS is not active anymore.

After calculating maintenance cost and benefit the test was removed. Mocking around would have made to complex.

Fix #2001

### DIFF
--- a/common/statedata.py
+++ b/common/statedata.py
@@ -158,7 +158,8 @@ class StateData(dict, metaclass=singleton.Singleton):
         self['_meta'] = meta
 
     def save(self):
-        """Store application state data to a file."""
+        """Store application state data to a file.
+        """
         logger.debug('Save state data.')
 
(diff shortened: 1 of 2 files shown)


buhtz:
Closing unmerged in favor of PR #2010

PR #2026, opened 2025-02-03
Update documentation to include offset minutes

fallingrock:
Added note to docs about offset minutes.

@buhtz, do you want to update the screen shots so they are consistent?


buhtz:
> Added note to docs about offset minutes.

Cool.

> @buhtz, do you want to update the screen shots so they are consistent?

Good that you consider such details. ❤️ I am not aware of a screenshot that shows the schedule options currently. I also think there is now high need for it now. The docu and the GUI will have a big redesign in the next months. I would say no.

buhtz (approved):

Issue #2073, opened 2025-04-01
doc: Finalize snapshot-to-backup-rephrasing


buhtz:
# ✨Welcome to the project.✨
🎯 If you pick up this Issue because of the `GOOD FIRST USE` label. You will be mentored through the process if you want. 🤝
First the Issue is explained. At the end you will find some guidance for first contributors.

📡 Please do not hesitate to ask questions.
📡 Your solution don't need to be perfect.
🚫 Participation or contribution is only meaningful if you are familiar with the Back In Time application and are an active user. This project is not intended for learning basic Python skills or enhancing your profile at Microsoft GitHub.

# 📝Problem & Situation
Back In Time used the term "snapshots" in the past. Today the term is dropped and replaced with "backup". The process is nearly finished in the GUI but not in additional documents.

# 🏁Solution
Check the following documents and modify "snapshots" to "backups". Don't use automatic search-and-replace-features. Check the context.

* [x] man pages (blocked by #2085, #2124) Exception: config-manpage (#1923)
* [x] user manual
* [x] FAQ entry about the rephrasing
* [x] Check again the source code for print and log messages.
* [x] User callback scripts

# 👣Your next steps
1. If this is your first contribution in this project please introduce your self and tell us about your skills, wishes and plans. Also let us know how you found the issue and the project.
2. Don't ignore the previous step.
3. Read the existing [contributors documentation](https://github.com/bit-team/backintime/blob/dev/CONTRIBUTING.md).
4. We can develop the next steps in the further discussion. Don't hesitate to ask.

# Additional details
Follow up of #1929

shivarag200701:
@buhtz I would like to work on this, I have skills in Python and build scaleable application using FASTAPI and I came across this issue using the GOOD-FIRST-ISSUE label. I am very eager to make my first contribution through this.

What are the steps I need to take?

buhtz:
Hello Shiva,
thank you for your efforts to contribute. Please see the section "Your next steps" for the next steps.

Are you a regular user of Back In Time? Are you familiar with the application?

Regards,
Christian

shivarag200701:
hey, sorry for the late response, Actually i ahvent used it per say but but i have heard about it

buhtz:
Hello Shiva

I've noticed you use all lowercase. Proper capitalization improves readability and is gentle to the eyes of your readers. In English, capital letters are used at the beginning of sentences, for the word "I", and for proper nouns. It would be nice and helpful if you could consider that.

[...]

PR #2074, opened 2025-04-01
fix(translation): Rephrase "snapshots" to "backups" and also rephrase and neutralization of "you" strings

### LINKED ISSUE #1929: Rephrase "snapshot" to "backup"
Back In Time (BIT) does use the term "snapshot" in the GUI, CLI and documentation. The term will be rephrased into "backup".

Reason:
- Term "Backup" is more appropriate and pointing technical to what BIT is doing.
- Snapshot might be understood by some people as an image of the file system.

This happens after discussing this on the mailing list. https://mail.python.org/archives/list/bit-dev@python.org/thread/R7RJDJIW5IPXX4MBDBS44U2UROROZO57/

### LINKED ISSUE #2050: Depersonalize "you" source strings
### You-strings

There are still source strings using the word "you" speaking directly to the user.

This is unprofessional and the goal was to make BIT "speak" more seriously. Find (grep) all "you" strings and modify them in a way that they are de-personalized.

buhtz:
Fix #2050
Fix #1929

Follow-up about snapshots-to-backup: #2073


common/snapshots.py
```diff
@@ -1279,15 +1279,15 @@ def takeSnapshot(self, sid, now, include_folders):
         # instead, e.g. a DataClass
 
         if new_snapshot.exists() and new_snapshot.saveToContinue:
-            logger.info(f"Found leftover snapshot '{new_snapshot.displayID}' "
+            logger.info(f"Found leftover backup '{new_snapshot.displayID}' "
```
s4moore: I would use `partial` or `incomplete` rather than `leftover`.

qt/app.py
```diff
@@ -1396,8 +1397,8 @@ def hideItem(item):
 
         question_msg = '{}\n{}'.format(
             ngettext(
-                'Are you sure you want to remove this snapshot?',
-                'Are you sure you want to remove these snapshots?',
+                'Remove this backup?',
+                'Remove this backups?',
```
s4moore: Remove `these` backups?

qt/manageprofiles/tab_options.py
```diff
@@ -38,21 +38,21 @@ def __init__(self, parent):
...
-        self.cbGlobalFlock = QCheckBox(_('Run only one snapshot at a time'))
+        self.cbGlobalFlock = QCheckBox(_('Run only one backup at a time'))
         tab_layout.addWidget(self.cbGlobalFlock)
         qttools.set_wrapped_tooltip(
             self.cbGlobalFlock,
-            _('Other snapshots will be blocked until the current snapshot '
+            _('Other backups will be blocked until the current backup '
               'is done. This is a global option. So it will affect all '
-              'profiles for this user. But you need to activate this for all '
+              'profiles for this user. But it need to be activated for all '
```
s4moore: But it need`s` to be

s4moore (commented):
I've left a few minor suggestions in the notes.

PR #2077, opened 2025-04-03
chore: Fix git history

buhtz:
Restore the git history of bitwidgets.py and timeline.py, which where split from qttools.py.


buhtz:
This PR was an experiment. It did not work but also did not destroy anything.

buhtz:
For the record: It works nearly as expected. Only Microsoft GitHub can not handle it.

The history for `qt/timeline.py` and `qt/bitwidgets.py` does contain all history entries regarding `qt/qttools.py` because they are split from that file. Use `git log --follow qt/timeline.py` and you see a history going back to the year 2008.

But viewing the history of that file on MS GitHub does not show this.

Issue #1936, opened 2024-11-20
translation: Improve meta-data handling (authors, licence, copyright, ...)


buhtz:
The meta data in the po-files is messy (e.g. authors, license, version). Our Weblate platform seems not to be able to handle things like this. So we need to manage that by our own. This is an example of one po-file header

```
# Swedish translation for backintime
# Copyright (c) 2008 Rosetta Contributors and Canonical Ltd 2008
# This file is distributed under the same license as the backintime package.
# FIRST AUTHOR <EMAIL@ADDRESS>, 2008.
#
msgid ""
msgstr ""
"Project-Id-Version: backintime\n"
"Report-Msgid-Bugs-To: https://github.com/bit-team/backintime\n"
"POT-Creation-Date: 2024-11-13 09:21+0100\n"
"PO-Revision-Date: 2024-09-25 08:52+0000\n"
"Last-Translator: Umeaman <Umeaman@users.noreply.translate.codeberg.org>\n"
"Language-Team: Swedish <https://translate.codeberg.org/projects/backintime/common/sv/>\n"
"Language: sv\n"
"MIME-Version: 1.0\n"
"Content-Type: text/plain; charset=UTF-8\n"
"Content-Transfer-Encoding: 8bit\n"
"Plural-Forms: nplurals=2; plural=n != 1;\n"
"X-Generator: Weblate 5.7.2\n"
"X-Launchpad-Export-Date: 2022-11-06 19:32+0000\n"
```

This is just a bunch of ideas I need to thinking through.

- [x] ~~Convert `TRANSLATORS` file into json format.~~
- [x] ~~Use `msgcat` to manipulate the headers.~~
- [x] ~~Add all known translators (from translator-placeholder-string file) to `Last-Translator` field.~~
- [x] ~~Use `updateversion.sh` ~~with `msgcat`~~ to set the `Project-Id-Version` field.~~ No need for this. Handled by `update_language_files.py` reading from `version.py`.
- [X] ~~Warn in `update_language_files.py` if `Project-Id-Version` does not fit to the `VERSION` file.~~ Obsolete.
- [x] SPDX meta data
  - [x] Check if Weblate does overwrite the SPDX comments in po files. If not ...
  - [x] Manually create SPDX header for each po file based on the TRANSLATORS file.
  - [X] Remove TRANSLATORS file.
  - [x] ~~Use `update_language_files.py` script to recreate a Python readable (like language.py) list of translators usable in the About dialog.~~
[...]

PR #2082, opened 2025-04-04
feat: Redesign the About dialog

buhtz:
In preparation for #1936 to integrate translator credits via "translator-string" in the po-files.

image

The Authors field is fed by the `AUTHORS` file.

The Translators field is fed, not from the `TRANSLATORS` file (that will be removed in the near future), but by the translation system itself using a placeholder string. See [this example](https://hosted.weblate.org/translate/peek/translations/de/?checksum=2736a69f5de012b1&q=translator-credits) please. Translators do add their names by themself as a translation.


buhtz:
Here is a variant with thin borders. What do you think?
image

buhtz:
And this, without borders but different background.

image

image

buhtz:
And another one.
image
image

buhtz:
Next
image
image

buhtz:
OK, the last variant for today

image
image

buhtz:
OK, this might be the final version.

Light mode on top, dark mode buttom.
Left in "German" mode (see Translators field).
Right in "English" mode (the source language without translation).
The git-line appear only if running from a git repo.

...following...

PR #2102, opened 2025-04-24
feat(gui): Add shutdown confirmation dialog

LatiosInAltoMare:
Add a window so that user can interrupt the shutdown process.
Modifications are on the "tools.py", adding a window and show a counter to user.


buhtz:
Hello Huaide Jiang,
that PR is good looking and a nice start to that new feature. Thank you for your efforts and contribution. Please let me made some suggestions.

Back In Time (BIT) consists of two components (or applications). There is the command-line interface (CLI) represented by the `common` folder. And there is the graphical user interface (GUI) associated with the folder named `qt`. Because of that I would suggest to move your code from `common/tools.py` somewhere into `qt/*`.

Feel free to create a new fresh file instead of blowing up existing files. You do have this nice `class ShutdownWarningDialog`. So why not creating a file `qt/shutdowndlg.py` or `qt/shutdownwarningdlg.py`?

You can also prepare the header of that file fitting our SPDX/REUSE standards liket this:

```
# SPDX-FileCopyrightText: © 2025 YOUR_IDENTIFIER_AND_IF_YOU_WANT_EMAIL
#
# SPDX-License-Identifier: GPL-2.0-or-later
#
# This file is part of the program "Back In Time" which is released under GNU
# General Public License v2 (GPLv2). See LICENSES directory or go to
# <https://spdx.org/licenses/GPL-2.0-or-later.html>.
```

But I also realize that the shutdown mechanic is realized via `class ShutDown` in file `common/tools.py` this will make things about more complex. And I am not yet sure about a good solution.

Please hold. I'll do some refactoring before you can go further.

buhtz:
I updated the dev branch. So please update your forked repo and this branch. You know how to do it?

Be aware of merge conflicts coming up.

I move the `class ShutDown` into its own file.

LatiosInAltoMare:
Yes I know how to do it. Thanks for your kind and patient guidance!

So I think the next step is to move the logic of warning window into `/qt` ?

buhtz:
> So I think the next step is to move the logic of warning window into `/qt` ?

Yes, correct.

LatiosInAltoMare:
Please check now, Thank you!

buhtz (changes requested):
Looks nice. I recommend some minor adjustments to improve code quality (which is still quit high :heart: ).

- Please rename `qt/shutdowndialog.py` into `qt/shutdowndlg.py`.
- Is there a good reason why `class ShutdownWarningDialog` is private in that function? If not I would suggest to move it outside of this function into the module.
- Please add the file `qt/shutdowndlg.py` to the (alphabetical ordered) list `full_test_files`.

https://github.com/bit-team/backintime/blob/391b4b90c5a15998e75a084df5fc3e987fed66cc/qt/test/test_lint.py#L41-L56

- Then install testing dependencies:
  - pylint (>= 3.3.0)
  - flake8
  - ruff (>= 0.6.0)
  - codespell
  - reuse (>= 4.0.0)
- Then [run the test suite](https://github.com/bit-team/backintime/blob/dev/CONTRIBUTING.md#testing). The test suite does call the linters. The linters will give you several errors. Please try to fix them.

LatiosInAltoMare:
Hello @buhtz , except for testing suite, other modifications are done. Thanks for your kind support!
But I'm having trouble with testing now. When I follows the test instruction, it tells me that `make: Nothing to be done for 'test'.` Can you please help me with that?

```
(backintime) huaide@huaide-VMware-Virtual-Platform:~/backintime/backintime_testing/backintime/qt$ make test
make: Nothing to be done for 'test'.
```

buhtz:
> make: Nothing to be done for 'test'.

That is my fault. I need to adjust the makefile and the docu about that.
Ignore the make-system and use pytest directly.

    user@host:~/backintime/backintime_testing/backintime/qt$ pytest

Or more verbose

    user@host:~/backintime/backintime_testing/backintime/qt$ pytest -v

Or stop at the first fail

    user@host:~/backintime/backintime_testing/backintime/qt$ pytest -vx

**EDIT**: Or if you want to spare some time. Tun the linter tests only.

    user@host:~/backintime/backintime_testing/backintime/common$ pytest test/test_lint.py
    user@host:~/backintime/backintime_testing/backintime/qt$ pytest test/test_lint.py

And you can check the build log of TravisCI: https://app.travis-ci.com/github/bit-team/backintime/builds/275020248

LatiosInAltoMare:
Thanks @buhtz , I have modified according to linters and passed all the tests, please check.

qt/shutdowndlg.py
```diff
@@ -0,0 +1,67 @@
...
+import os
+from PyQt6.QtWidgets import (QApplication,
+                             QDialog,
+                             QLabel,
+                             QPushButton,
+                             QVBoxLayout)
+from PyQt6.QtCore import QTimer
+
+
+class ShutdownWarningDialog(QDialog):
```
buhtz:
```suggestion
class ShutdownCountdownDlg(QDialog):
```

qt/shutdowndlg.py
```diff
@@ -0,0 +1,67 @@
...
+    """A UI class for a shutting down window."""
+    def __init__(self, countdown):
+        super().__init__()
+        self.countdown = countdown
+
+        # Initialize UI components
+        self.setWindowTitle("Back In Time - Shutdown Warning")
```
buhtz:
```suggestion
        self.setWindowTitle(_('Countdown to Shutdown'))
```

qt/shutdowndlg.py
```diff
@@ -0,0 +1,67 @@
...
+        self.setFixedSize(300, 150)
```
buhtz: I see no need for this `setFixedSize()`. Let Qt mange the size itself.

qt/shutdowndlg.py
```diff
@@ -0,0 +1,67 @@
...
+        self.label = QLabel(f"Shutdown in {self.countdown}s", self)
```
buhtz: Fill the lable in the update_countdown() method only.

```suggestion
        self.label = QLabel('', self)
```

qt/shutdowndlg.py
```diff
@@ -0,0 +1,67 @@
...
+        self.cancel_button = QPushButton("Cancel Shutdown", self)
```
buhtz:
```suggestion
        self.cancel_button = QPushButton(_('Cancel Countdown'), self)
```

qt/shutdowndlg.py
```diff
@@ -0,0 +1,67 @@
...
+def show_shutdown_warning(countdown=30):
+    """
+    Show a warning window with 30 seconds countdown
+    """
+    if not os.environ.get('DISPLAY'):
+        return True  # No GUI
+
+    app = QApplication.instance()
+    if not app:
+        app = QApplication([])
```
buhtz: I see no need for all this magic. You can assume that the application instance and GUI is already running.

I also suggest to rename the function to something like `shutdown_countdown()` or `countdown_to_shutdown()`. There is no real warning and the function does more than just "show" something.

qt/shutdowndlg.py
```diff
@@ -0,0 +1,67 @@
...
+
+        # Initialize timer
+        self.timer = QTimer(self)
+        self.timer.timeout.connect(self.update_countdown)
+        self.timer.start(1000)
+
+    def update_countdown(self):
+        """Update the countdown in the UI."""
+        self.countdown -= 1
+        self.label.setText(f"Shutdown in {self.countdown}s")
```
buhtz: This is how you would translate singular and plural.
Do `import gettext` in the beginning of the file.

```suggestion
        self.label.setText(gettext.ngettext(
            'Shutdown in {n} second.',
            'Shutdown in {n} seconds.',
            self.countdown).format(n=self.countdown))
```

Based on the value of `self.countdown` the translation backend (GNU gettext) will choose the right string. There are some languages using more than just singular and plural. The function `ngettext()` is a special form of `_()`.

buhtz (changes requested):
OK, great. Now, we can come to the details. :smile: 

I add suggestions in the diff view. You can directly "commit" them there if you want.

I made the strings translatable with using the `_('translate me')` structure.

This is how it looks on my machine.

image

- Please center the QLabel in the dialog.
- Add a second button, e.g. `_('Shutdown now')`
- Please run the linter tests also in the "common" folder not only in "qt".
- Maybe give the user more information in the dialog. Something like "The backup has finished. The computer will shut down {n} seconds." Just a suggestions.

Everything from me are suggestions and not "must haves".

qt/shutdowndlg.py
LatiosInAltoMare: This one is needed, otherwise in the first second there'll be nothing in the window.

buhtz: I understand. But than just make sure the update-label message is called once from the __init__() method.

    self.label = QLabel('', self)
    self.update_countdown()

My goal here is to reduce redundancies in the code.

qt/shutdowndlg.py
```diff
@@ -0,0 +1,82 @@
...
+        # Immediately accept on shutdown now
+        self.shutdown_button.clicked.connect(self.accept)
+
+        # Layout setup
+        layout = QVBoxLayout()
+        # Add stretch before label to center it vertically
+        layout.addStretch(1)
+        layout.addWidget(self.label)
+        # Add stretch after label to center it vertically
+        layout.addStretch(1)
```
buhtz: Try without stretch elements. For example:
```suggestion
        layout.addWidget(self.label, alignment=Qt.AlignmentFlag.AlignHCenter)
```

buhtz (changes requested):

LatiosInAltoMare:
Everything all set. Please check now.

buhtz:
image

Use a dot instead of a comma please.

Please rearrange the two sentences vertical. It would IMHO be more gentle to the eyes.

"Finished backup" is to short. Keep in mind that BIT is targeting desktop users, not technical professional users. So use easy and day-to-day language.


    The backup has finished.

    The computer will shut down in {n} seconds.

Or shorter

    The backup has finished.

    Shutdown in {n} seconds.

LatiosInAltoMare:
Much more friendly now I think 😄

qt/shutdowndlg.py
```diff
@@ -54,8 +54,10 @@ def update_countdown(self):
         """Update the countdown in the UI."""
         self.countdown -= 1
         self.label.setText(gettext.ngettext(
-            'Finished backup, shutdown in {n} second.',
-            'Finished backup, Shutdown in {n} seconds.',
+            'Backup completed successfully.\n' +
```
buhtz: But we don't know if the backup was "successful". We just know it has come to an end.

When it comes to translations we have the rule to now using escape characters (`\n`) in that strings translators will see. This is because some translators don't know about programming and won't understand the meaning of `\n`.

I suggest to use two QLabel's, each for one of the sentences.

buhtz (changes requested):

LatiosInAltoMare:
Hi @buhtz , I have modified it to 2 QLabels, please check now.

buhtz (approved):
Hello LatiosInAltoMare,

Your PR looks fine. It will get merged after a cool off period. To find out how we usually tread PRs please read ["What happens after you opened a PullRequest (PR)?"](https://github.com/bit-team/backintime/blob/dev/CONTRIBUTING.md#what-happens-after-you-opened-a-pull-request-PR).

Best,
Christian

Peek 2025-04-29 08-34

PR #2136, opened 2025-05-12
refactor: Rename modul "exceptions" to "bitexc"

buhtz:
The module "exceptions" was renamed to "bitexc" because of collusion with Python own "exceptions" module.


buhtz:
Stupid me. Python's exceptions do exist in the main namespace. There is not module "exceptions" need to be imported.

PR #2156, opened 2025-05-19
Fix: Prioritize id_ed25519 in default SSH key selection

### LINKED ISSUE #2094: Create new SSH profile: Existing SSH key not pre-selected

1. v1.5.4
2. Password-less SSH setup using a "modern" key named (different from "id_rsa") id_ed25519.
3. Setup a SSH profile.
4. The existing key file is not pre-selected in the key-file widget.

The reason might be that the key is not "id_rsa" but "id_ed*".

### Related
- #2095
- #1146
- #2194

Fan-hr:
Ensures that `Config.sshPrivateKeyFile()` correctly defaults to `~/.ssh/id_ed25519` if present and no key is explicitly configured, addressing the issue where it might not be pre-selected in the GUI. Includes unit tests for various key file scenarios.

Close #2094


buhtz:
Hello haoranfan,

```py
    def sshPrivateKeyFile(self, profile_id = None):
        ssh = self.sshPrivateKeyFolder()
        default = ''
        for f in ['id_dsa', 'id_rsa', 'identity']:
            private_key = os.path.join(ssh, f)
            if os.path.isfile(private_key):
                default = private_key
                break
        #?Private key file used for password-less authentication on remote host.
        #?;absolute path to private key file;~/.ssh/id_dsa
        f = self.profileStrValue('snapshots.ssh.private_key_file', default, profile_id)
        if f:
            return f
        return default
```

I was not aware that this logic is contained in `class Config`.
This might be an opportunity to optimize the situation and go beyond the original issue.

If you agree, I would suggest:
1. Make `Config.sshPrivateKeyFile()` as dumb as possible, without any pre-selecting or default values. `Config` should manage config values, nothing more.
2. The pre-selection behavior should be moved direct into the related dialog (`qt/manageprofiles/tab_general.py`) or in a helper module (e.g. `common/tools.py` or `common/sshtools.py`) used by that dialog.
3. Decouple the helper function from the file system, to make it testable without any mocking. The function (e.g. `select_private_key_file()`) should be given a list of file paths (type `pathlib.Path` not `str`).
4. I am not sure about this point. Let me know if it might be a good idea: Isn't it the case that some folks do use individual file names for their keys? If none of the pre-defined names are available but something else (e.g. `super-secret-key`, `trust-no-one`, `42`, ...) this file should be used.
   - If checking for other files, ignore (maybe case-insensitive) the following: `*.pub`, `*config`, `*backup`, `authorized_keys`, `known_hosts`, `*.old`
6. Make sure that there is no `.pub` suffix in the file name.
7. I also suggest to remove `identiy` and `id_dsa` from the list because these are damn old and unsecure keys. If some users still using them they would be covered by point 4 (use some other existing file if none of standard filenames are available).
8. Maybe add `id_ecdsa` to the list?

This are just quick n dirty suggestions and ideas. You don't have to make it that way. 😃 

About tests: Feel free to remove tests if you think they are not necessary or not valuable. When it comes to (unit) tests, please always calculate and balance benefit and maintenance costs.

Please see the projects background information to get an idea about our workflow and priorities:

- [Maintenance status](https://github.com/bit-team/backintime#maintenance-status)
- [The team](https://github.com/bit-team/backintime#the-team)
- [Strategy outline / Roadmap](https://github.com/bit-team/backintime/blob/dev/CONTRIBUTING.md#strategy-outline)

Regards,
Christian

**EDIT**: Feel free to split this into several PRs.

Fan-hr:
Thanks for the helpful suggestions!

 `Config.sshPrivateKeyFile()` now simply returns the config value, with no built-in default logic.
 Moved default key selection logic into a helper: `select_private_key_file()`.
 The helper takes a list of `pathlib.Path` objects, filters out irrelevant files (e.g., `.pub`, `known_hosts`, `config`, `*.old`, etc.), and prioritizes:
   `id_ed25519 > id_ecdsa > id_rsa`, with fallback to the first valid remaining key.

As you suggested, the function also supports non-standard private key names such as `super-secret-key`.  
If no preferred keys are found, the first matching file (not excluded by the ignore rules) will be selected.

qt/manageprofiles/tab_general.py
```diff
@@ -310,7 +310,18 @@ def load_values(self) -> Any:
         self.txtSshUser.setText(self.config.sshUser())
         self.txtSshPath.setText(self.config.sshSnapshotsPath())
         self.comboSshCipher.select_by_data(self.config.sshCipher())
-        self.txtSshPrivateKeyFile.setText(self.config.sshPrivateKeyFile())
+
+        key_path = self.config.sshPrivateKeyFile()
+        if not key_path:
+            try:
+                ssh_dir = Path.home() / '.ssh'
```
buhtz, line 314: There is `Config.sshPrivateKeyFolder()` returning the same value.

But I would say moving this "constant" into `common/bitbase.py` like this

    DIR_SSH = Path.home() / '.ssh'

Then you can remove `Config.sshPrivateKeyFolder()` and check via `git grep` if it is used somewhere else.

qt/manageprofiles/tab_general.py
```diff
@@ -310,7 +310,18 @@ def load_values(self) -> Any:
...
+                if ssh_dir.exists():
```
buhtz, line 315: Do we need it? We still have the try-except block catching missing ssh-folder, aren't we?

common/sshtools.py
```diff
@@ -1309,3 +1309,27 @@ def writeKnownHostsFile(key):
     with open(knownHostFile, 'at') as f:
         logger.info('Write host key to {}'.format(knownHostFile))
         f.write(key + '\n')
+
+
+def select_private_key_file(files: list[Path]):
```
buhtz, line 1321: Because you hand over a generator (Path.iterdir()) instead of a list.
```suggestion
def select_private_key_file(files: Iterable[Path]):
```

On top of the file: `from typing import Iterable`.

buhtz (changes requested):
Sometimes contributors can't see my in-code review comments. I am not sure why. So I am just asking if you can see them?

buhtz:
Hello haoranfan,
Can you please give me an update about status. I would like to bring this PR to an end because it affects other modifications on my road map.
Regards,
Christian

**EDIT**: FYI. I solved merge conflicts and updated your branch to the latest dev.

buhtz:
Without response I have no other choice, than closing the issue without merging.

buhtz:
I am in the middle of fixing #2094 and related issues.
This makes this PR obsolete.

PR #2157, opened 2025-05-19
feat: Add threshold "Less storage space" to warn the user

### LINKED ISSUE #2110: threshold "Less storage space" to warn the user
⚠️Be aware that his is not a task for beginners or learners.⚠️
⚠️You'll need some skills and experience with Python coding.⚠️

# ✨Welcome to the project.✨
🎯 If you pick up this Issue because of the `GOOD FIRST USE` label. You will be mentored through the process if you want. 🤝
First the Issue is explained. At the end you will find some guidance for first contributors.

📡 Please do not hesitate to ask questions.
📡 Your solution don't need to be perfect.
🚫 Participation or contribution is only meaningful if you are familiar with the Back In Time application and are an active user. This project is not intended for learning basic Python skills or enhancing your profile at Microsoft GitHub.

# 🏁Solution
* 📌 A new config option giving a threshold at which BIT does warn the user about less free space on the * backup destination.
* 📌 That value should always be equal or higher than "less free space" value in remove-&-retention-policy options.
* 📌 The check and warning should be done before the backup starts.
  * 📌 On GUI it should interrupt and ask in a dialog.
  * 📌 Headless (via CLI) a error should be shout out to the log and (if available) systray. But proceed with the backup anyway, don't stop it.

# 👣Your next steps
1. If this is your first contribution in this project please introduce your self and tell us about your skills, wishes and plans. Also let us know how you found the issue and the project.
2. Don't ignore the previous step.
3. Read the existing [contributors documentation](https://github.com/bit-team/backintime/blob/dev/CONTRIBUTING.md).
4. We can develop the next steps in the further discussion. Don't hesitate to ask.

This issue is a Follow-up from #1917

fest6:
Blocked by implementing details described in #2110.

Added a new config giving a threshold to warn the user about less free space.
Implemented GUI settings panel for adjustments.

~~Close #2110~~


buhtz (changes requested):
Hello fest6,

Looks good on a first sight. Would you mind to place the setting in the "Options" tab instead of "Remove & Retention"?

Best,
Christian

fest6:
Thank you for the feedback and kind words! I've moved the settings as requested.

common/config.py
```diff
@@ -959,6 +980,33 @@ def setRemoveOldSnapshots(self, enabled, value, unit, profile_id = None):
...
+            return value
+
+        value *= 1024  # Gb
+        if unit == self.DISK_UNIT_GB:
+            return value
+
+        return 0
+
+    def setWarnFreeSpace(self, enabled, value, unit, profile_id=None):
+        self.setProfileBoolValue('snapshots.warn_free_space.enabled', enabled, profile_id)
```
buhtz: I know that `minFreeSpace()` also has an `enabled` field. But I would prefer to vibe out that field and simply check `value` if it is `None` or `<= 0` to implicit indicate that the option is disabled.

qt/manageprofiles/tab_options.py
```diff
@@ -83,6 +85,23 @@ def __init__(self, parent):
...
+
+        WARN_FREE_SPACE_UNITS = {
+            config.Config.DISK_UNIT_MB: 'MiB',
+            config.Config.DISK_UNIT_GB: 'GiB'
+        }
+        self.suWarnFreeSpace = SpinBoxWithUnit(
+            self, (1, 99999), WARN_FREE_SPACE_UNITS)
+
+        self.cbWarnFreeSpace = StateBindCheckBox(
+            _('Show warning if the free space is less than'), self)
```
buhtz:
```suggestion
            _('Warn if the free disk space falls below'), self)
```

I am not very good in English. "...is less than" would also be OK for me.

buhtz (changes requested):
This is nice and smooth. I like it. Just some minor suggestions.

Please also add a tool tip to the setting and its two widgets. Quick n dirty (with LLM help): _Shows a warning when free space on the backup destination disk is less than the specified value.
If the Remove & Retention policy is enabled and old backups are removed based on available free space, this value cannot be lower than the value set in the policy._

I do have a helper function to handle long tooltips, wrap them and organize them in paragraphs:

```py
# Two paragraphs
# Having two translatable strings is on purpose because it avoid escape
# characters (e.g. \n) for non-technical translators.
tooltip = [
    _('Shows a warning when free space on the backup destination disk is less '
    'than the specified value.'),
    _('If the Remove & Retention policy is enabled and old backups are removed '
    'based on available free space, this value cannot be lower than the value '
    'set in the policy.')
]
qttools.set_wrapped_tooltip(self.suWarnFreeSpace, tooltip)
qttools.set_wrapped_tooltip(self.cbWarnFreeSpace, tooltip)
```

You can try to add a changelog entry if you want. I'll rephrase it later if needed.

Do you plan any further steps within or after this PR?

We can merge it in this state and going further in a new PR.
The thing is that the warning currently occurres only when `check-config` command is used on the CLI.
I would like to have warning in the logs and notify system, when BIT runs via cron in background or without GUI in CLI. And additionally a warning dialog when starting (but before running) a backup job manually in the GUI. The GUI then should warn and ask the user to proceed with the backup or not.

But this can be realized in the next PR.

common/config.py
fest6: I understand your point. However, I prefer to maintain the original method as it allows us to retain the previous value even after the option is disabled. This approach can be beneficial for users who may want to re-enable it without having to reset the value.

That said, I'm open to making the change if you believe it is necessary.

fest6:
Thank you for your feedback! I have made the requested changes, except for the one regarding the `enabled` field.

If everything looks good, please feel free to merge this PR, as I do not have any further plans for this one.

qt/manageprofiles/tab_options.py
fest6: I've made the suggested change.

common/config.py
buhtz: I get your point. The config file format will be modified into TOML (like an INI file) in the near future to make it easier for users without GUI to create config file in a text editor. For them I need to keep that file simple.

buhtz:
Hello fest6,
thank you for the modifications.

> If everything looks good, please feel free to merge this PR

The warning dialog itself is still missing.

> as I do not have any further plans for this one.

That is OK. Then I'll take over from here and finalize the PR.

Thank you very much.
Regards,
Christian

buhtz:
According to your plan described in #2110 the first of three items are implemented with this PR. I put this PR on hold because the other need to get implemented also.

buhtz:
This is it. I implemented the rest.

Issue #2258, opened 2025-09-15
Consider out-of-line bugfix-only release 1.5.6 for Debian GNU/Linux stable (13, Trixie)


buhtz:
Here is a list of potential fixes we have done in the past but not yet released. These fixed might be candidates for an updated BIT version in the current stable release of Debian GNU/Linux. These fixed need to be isolated and cherry-picked into an extra branch, because the latest development version (`dev` branch) is far beyond and not ready to be released.

The latest Debian release is quit young. Let's see if some other issues are coming up. None of them are grave bugs but it is known that the Debian maintainer would be willing to update the BIT version if the fixes are small and reasonable.

- #2257
- 7d45396 (maybe)
- #2084 & #2268
- #2128
- #2266

Despite the work that is needed I am also concerened about new restrictions or bugs we might introduce in a "stable" Debian release. So I prefer not to release a 1.5.6 targeting current Debian stable. But if...

---

It seems there is a need for an 1.5.6 release offering minor fixes/workarounds for known issues.

- Unit-xwindowid-conversion-crash:
  - Despite it is not reproducible, since the Debian 13 release, it seems to affect a relevant amount of users.
  - Fixing it with [this simple workaround](https://github.com/bit-team/backintime/issues/2084#issuecomment-2787602155).
  - #2084
  - #2268
- Crash when using `qt6ct`:
  - Fix with #2131
- Crash when open (missing) changelog:
  - #2266
- Consistent behavior with scheduling "Repeatedly (anacron)" (revert unintentional behavior change introduced in 1.5.3 (060324e859, #1791)
  - `older_than()` (PR #2257)

buhtz:
Fixed in release [1.5.6](https://github.com/bit-team/backintime/releases/tag/v1.5.6). Hope it will [make it into current Debian stable](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1117254).

EDIT (2025-10-23): Update the request in ticket with direct email contact to Debian Maintainer [Jonathan Wiltshire](https://qa.debian.org/developer.php?login=jmw%40debian.org).

PR #2268, opened 2025-09-29
fix-UInt32-overflow in common/tools.py

ixnewton:
Proposed fix in backintime/common/tools.py to resolve the out-of-range error in release 1.5.5-1 when attempting restore of folder and contents (kernel: 6.16.8-1 kde plasma: 6.3.6 qt: 6.9.2 python: 3.13) 

```
29/09/2025 14:50	python	Traceback (most recent call last):
  File "/usr/share/backintime/qt/app.py", line 1631, in restoreThis
    rd.exec()
    ~~~~~~~^^
  File "/usr/share/backintime/qt/restoredialog.py", line 82, in exec
    self.config.inhibitCookie = tools.inhibitSuspend(toplevel_xid = self.config.xWindowId, reason = 'restoring')
                                ~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/share/backintime/common/tools.py", line 2045, in inhibitSuspend
    cookie = proxy(*[(app_id, dbus.UInt32(toplevel_xid), reason, dbus.UInt32(flags))[i] for i in dbus_props['arguments']])
                              ~~~~~~~~~~~^^^^^^^^^^^^^^
OverflowError: Value 527453920 out of range for UInt32



```

### DIFF
--- a/common/tools.py
+++ b/common/tools.py
@@ -2028,7 +2028,12 @@ def inhibitSuspend(app_id = sys.argv[0],
     try:
         if not toplevel_xid:
             toplevel_xid = 0
-    except IndexError:
+        # Ensure toplevel_xid fits in UInt32 range (0 to 4294967295)
+        # If it's too large, use 0 as fallback since window ID is not critical for inhibit
+        elif toplevel_xid > 0xFFFFFFFF:  # 4294967295
+            logger.debug(f'Window ID {toplevel_xid} exceeds UInt32 range, using 0 as fallback')
+            toplevel_xid = 0
+    except (IndexError, TypeError, ValueError):
         toplevel_xid = 0
 
     for dbus_props in INHIBIT_DBUS:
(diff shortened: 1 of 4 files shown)


buhtz:
Hello Ian,
thank you for the PR and your efforts. I appreciate that.

But I am sorry to tell you that the problem is still fixed.
It is often a good idea to made an announcement before investing resources in a fix.

There is a fix in the latest dev version targeting upcoming release 1.6.0 (see changelog). See commit 4ec1f82.

But for Debian stable that fix is to complex and risky to be allowed in current "stable" Debian. I do have something else and more simple in mind.

Please just tell me if you experience that UInt32 overflow problem in your Debian 13 installation? Are you aware of other users?

In general I do have some minor issues/fixes that I consider for 1.5.6 in hope that Debian will allow them in current "stable" (13, Trixie). See #2258 as a meta issue about it. The point is I do not like that way. Stable is stable and shouldn't be changed; that is why Debian is rock solid. Anyway, there are exceptions of course. But I need to keep the diff between 1.5.5 and 1.5.6 as simple as possible, to lower the risk of introducing new bugs.

What do you think? And what can you tell me about your experience with that overflow problem?

Regards,
Christian

ixnewton:
Thanks for your reply, I should have done a deeper dive into the repo issues to see the status. However it was relatively straightforward to fix this problem locally and test. I'm not a Debian user so I can't comment on releases. In the Arch/Manjaro world I keep to the stable branch of the rolling release, so I'm sure I will get the 1.6.0 when available. For the time being my fix works without issues and will be overwritten on update.

I'm trying to get more experience in contributing to open-source projects, so thanks for your feedback.

Cheers,
Ian

buhtz:
Hello Ian,

> I should have done a deeper dive into the repo issues to see the status.

But it is also not that easy, especially in this project, to keep track of all issues. I am also not very good in writing well formulated commit messages and changelog entries.

> so I'm sure I will get the 1.6.0 when available.

I think I'll fix #2258 and release a 1.5.6 out-of-the-line just containing some minor fixes/workarounds.

Regards,
Christian

buhtz:
Fixed in release [1.5.6](https://github.com/bit-team/backintime/releases/tag/v1.5.6). Hope it will [make it into current Debian stable](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1117254).

buhtz:
Still no response from Debian maintainer. There is an open ticket about it without any response from him. Maybe it would help the process if you could also express your interest in an update from 1.5.5 to 1.5.6 ?
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1117254

Issue #2245, opened 2025-08-21
Allow execution without cron/crontab present


m4rcu5:
### Describe the problem, feature or ask a question:

Hiya,

Let me start of by acknowledging that I have done my research trough other issues, and I totally respect your position that cron is the scheduler and there is no support for systemd-timers, as you also stated in #1741.

I did notice that a lot of internal functions with regards to cron and crontab do fail graceful informing the user that you should install/start cron if they want scheduling to work (for example in #1761).
Except that you cannot run BIT when there is no `contab` or `fcrontab` command installed due to the following section:
https://github.com/bit-team/backintime/blob/5083d24cdfea3908ece2bdc6a9514fdd17eb0ebe/common/schedule.py#L27-L51

Patching it out, and setting `CRONTAB_COMMAND` to `/bin/false` seems to make BIT run as expected (sans the scheduling features) and I have backed up my machine without issues.

My request would be if it would be possible to let BIT execute without a dependency on crontab. Or ideally grey-out/lock-out the scheduling gui parts when there is not crontab available.

Just to clarify, I don't "run arch, btw", but OpenSUSE Tumbleweed, which also by default does not come with cron or crontab, nor does it come with any compatibility layer. They now seem to rely completely on timers.

buhtz:
Hello Marcus,

Thank you for taking the time to report the issue and providing the
details. I appreciate your feedback and work on a solution to the best of my ability.

It is a reasonable request I am open to implement that in the upcoming release.

Some questions:
- How do you install BIT on OpenSUSE Tumbleweed? Is there an official distro package?
- So you do the scheduling by yourself writing systemd-timers stuff?

If you have any more details to share, feel free to reach out.

Not sure when we'll find the time to work on it. Please see the projects background information to get an idea about our workflow and priorities:

- [Maintenance status](https://github.com/bit-team/backintime#maintenance-status)
- [The team](https://github.com/bit-team/backintime#the-team)
- [Strategy outline / Roadmap](https://github.com/bit-team/backintime/blob/dev/CONTRIBUTING.md#strategy-outline)

Best regards,
Christian

m4rcu5:
Hi Christian,

Your software is packaged for openSUSE TW (and Leap), right here: https://software.opensuse.org/package/backintime
I use that same package in my machines.

I don't schedule anything. I really use BiT as a manual backup/snapshot tool, to backup to my NAS. But I suppose I could schedule runs via systemd-timers if needed.

[...]

PR #2277, opened 2025-10-13
Gracefully handle missing cron/crontab, disable scheduling UI if unavailable

m4rcu5:
Hi,

As discussed in #2245, it would be nice to have BiT run, even if there is no cron or contab on the system. Just a a manual snapshot based backup system.

I took a stab at making this work.

The application will trow a warning informing the user whenever the crontab functions are called.
```
% qt/backintime-qt
which: no crontab in (/home/marcus/bin:/home/marcus/.local/bin:/home/marcus/bin:/usr/local/bin:/usr/bin:/bin:/sbin:/usr/sbin)
which: no fcrontab in (/home/marcus/bin:/home/marcus/.local/bin:/home/marcus/bin:/usr/local/bin:/usr/bin:/bin:/sbin:/usr/sbin)
WARNING: Command crontab and fcrontab not found.
```

The frontend has its scheduling functions greyed out and posts a warning message to inform the user.

**TODO:** Translations, I haven't touched those and I suppose they should be added.


buhtz:
Ah, now I know where I remember your nickname from. ;)

Thank you for this PR. Looks good in a first place.

Without this fix, does every user installing BIT 1.5.5 from the official OpenSUSE Tumbleweed repos, can not use BIT because it raise a RuntimeException because of a missing crontab? So BIT currently is useless for them?

m4rcu5:
> So BIT currently is useless for them?

No. The spec file has a hard requirement on the `cron` package. Which (at least on Tumbleweed) is not installed by default, as they are using systemd-timers.

A user can therefore just accept the dependencies and _also_ get cron installed. But this does reel in some of its own (recommended) dependencies you might not want on a desktop system; such as a `postfix`.
These could in turn be locked out in the package manager. But I that still leaves the user with either 2 scheduling daemons running, or if the cron service is not enabled, the false sense that scheduling is working.

My goal would be to have no hard dependency on cron in code, and get the package maintainer to change cron as a recommendation.

Ideally BiT would support systemd-timers, but I think there has already been a clear statement on that.

buhtz:
> My goal would be to have no hard dependency on cron in code, and get the package maintainer to change cron as a recommendation.

OK, I get it.

> Ideally BiT would support systemd-timers, but I think there has already been a clear statement on that.

But current "no" to that is not written in stone. It is more a matter of resources. Before adding something like this the code base need much more restructuring and cleaning.

buhtz:
The direction is OK for me. Let me add some suggestions.

**1. Visual**

The orange bold string seems to be to loud. I suggest to use an info-label or warning-label like this.





See `qttoosl.py::create_icon_label_info()` or `qttools.py:create_icon_label_warning()`.

**2. Import**

A simple `from common.schedule import XYZ` won't work because BIT does not follow the "modern Python way" (see #1575). In fact BIT does old-school `sys.path` hacking. Simply do `from schedule import CRONTAB_COMMAND`. And please put it on top of the file without try-except block.

**3. String**

The warning string itself sounds good to me. But I am not a native speaker. The translation is done via Weblate by our translators community. Just wrap a string between `_(...)` and the rest will follow. To support translation you can provide a final screenshot of the situation. I will add this to the Weblate translation platform to give translators more context.

buhtz (changes requested):
See my last comment.

buhtz:
It seems I finally broke my good old main machine. It will take 1 to 3 weeks until I am full operational again, because I have to replace it with a new one.

buhtz:
I am back on track. How can we proceed further?

buhtz:
image

PR #2278, opened 2025-10-20
feat: Add application logo

buhtz:
Back In Time's own application logo designed by @gregorydk Gregory Deseck.

- Logo source file is `backintime.Source.svg` containing full-colored version and symbolic version.
- PNG versions in many resolutions
  - Resolutions: 16 22 24 48 64 128 256 512
  - Reason: Not all desktop environments are able to handle SVG files.
  - Helper script `qt/icons/png_from_svg.sh`
    - `rsvg_convert` used for SVG to PNG conversion
    - `optipng` used to optimize PNG file size without losing quality
- Modified and adapted handling of `show-hidden.svg` icon.
- Makefile based install routine updated.
- Desktop shortcut files updated.
- New dependency `python3-pyqt6.qtsvg`

Close #1961

The PR is finished and free to be tested by others.

While preparing the screenshots I found another silly bug (#2279). And it seems that the xfce4-whiskermenu-plugin has problems to show our icons ([gitlab.xfce.org/panel-plugins/xfce4-whiskermenu-plugin#161](https://gitlab.xfce.org/panel-plugins/xfce4-whiskermenu-plugin/-/issues/161)), but I am assuming the problem is on their side not ours.

# Screenshots
This is Debian 13, with XFCE4 using Adwaita theme.

## Light
bit_logo_de_light

## Dark
bit_logo_de_dark

## Highlighted (glowing) starter icon on the desktop
bit_logo_starter_light

## Systray light
bit_systray_light

## Systray dark
bit_systray_dark

Issue #1961, opened 2024-12-06
🎨 Design & create a distinctive application logo


buhtz:
# 📌 What is this about?

Back In Time is a [FOSS](https://en.wikipedia.org/wiki/Free_and_open-source_software) (aka [open source](https://en.wikipedia.org/wiki/Open-source_software)) backup tool. It finally needs a proper logo!
This is a follow-up to [Issue #215](https://github.com/bit-team/backintime/issues/215), where some ideas and early drafts were shared.

We’re looking for a designer who can help us create a strong application icon and possibly extending to tray icons and a banner image.

We know this is a big ask, so we want to be transparent about everything from the start.

# 🎨 What we're looking for
- A main application icon. This is the first priority. Think of how it appears in the [About Dialog](https://translate.codeberg.org/media/screenshots/bit_aboutdlg_english.png).
- The icon should also work as a tray icon in light and dark themes.
  - Minimum supported size: 22×22 px
- Bonus: a banner image (1500×500 px) for our [Mastodon presence](https://fosstodon.org/@backintime) would be lovely.
- We would also like access to the source file (e.g. SVG), in addition to rendered formats (PNG, ICO, ...).
- Creative freedom: We have a few ideas and sketches (see below), but we’re open to new, fresh concepts!
- Inspiration and further resources:
  - [Gnome Icon Guidelines](https://developer.gnome.org/hig/guidelines/app-icons.html)
  - [AppIconPreview](https://flathub.org/apps/org.gnome.design.AppIconPreview) – useful tool for previewing icons
  - [Blogpost: Designing a GNOME App Icon](https://blogs.gnome.org/tbernard/2019/12/30/designing-an-icon-for-your-app/)
  - [SVG favicons + dark mode](https://css-tricks.com/svg-favicons-in-action/)
  - Also consider this comment https://github.com/bit-team/backintime/issues/215#issuecomment-1646945614 .

# ⚖️ Legal & Licensing
We know this is the boring part, but it’s important.

If you're familiar with licensing creative work, feel free to propose your own approach. If not, here’s what we suggest:
- You remain the copyright holder.
- The logo, including its source file, should be released under a license that’s compatible with GPL-2.0-or-later (the project’s license).
- Preferred license: GPL-2.0-or-later, but we’re open to other free licenses that work well for artwork.
- Important: We don’t own a trademark or plan to register one. That means it’s theoretically possible for others to re-use the logo under the license terms. But we’re not worried about that in practice.

[...]

PR #2293, opened 2025-11-10
fix: Handle dark and light mode with symbolic icon

buhtz:
- Fixed blurred symbolic icon. Now it scales well as a real SVG.
- Update the symbolic icon.
- Only SVG icons used. Removed all PNG versions from the repo.
- Related to issue #1961
- Follow-up of PR #2278

# Current state (update from 2025-11-11)
Icon scaling in XFCE is much more complex than I thought. 🤣 
In this screenshot the XFCE Panel is set to 128px self-height. The green numbers are the pixels settings for the icons (in xfce panel) them self. The systray addon in that panel is set to auto-scale the icons (and should adapt the panels icon scaling settings).

image

This is a more "normal" setup using 32px panel height with auto-scalled icons.

image

# Previous icon version just for comparision
image

image


buhtz:
OK, I think the PR is ready merge. I'll do a final test in a VM to be sure.

buhtz:
I installed Debian 13 (stable / Trixie) on a VirtualBox VM including all available desktop environments.

# Location of SPDX is important
It is essential to place the SPDX license info not on top of the SVG file but after the first `<svg>` tag. Otherwise the SVG file wont be read by most of the desktop environments.

# Application logo
The application logo in BIT itself worked.

- 🟢 LXDE
- 🟢 Cinnamon
- 🟢 Cinnamon
- 🟢 Gnome
- 🟢 Gnome Flashback
- 🟢 LXQt
- 🟢 MATE
- 🟢 KDE Plasma
- 🟢 XFCE

# Symbolic systray icon
Gnome and KDE are real problems and need further investigation by regular users of that desktop environments.

- 🟢 LXDE: Systray icon works in dark & light mode.
- 🔵 Cinnamon: Qt/Bit ignores dark/light theme and is always light. So the systray icon appears always in black.
- 🔵 Cinnamon on Wayland: The same as without Wayland.
- 🔴 Gnome: Does [not provide a systray](https://fosstodon.org/@backintime/115632665072528095).
- 🔴 Gnome Flashback: Does [not provide a systray](https://fosstodon.org/@backintime/115632665072528095).
- 🔴🔵 LXQt: Surrogate (gear) icon in systray.
- 🔵 MATE: Qt/Bit ignores dark/light theme and is always light. So the systray icon appears always in black.
- 🔵 KDE Plasma: Icon appears. But dark/light mode determination is not reliable.
- 🟢 XFCE: Appears

Issue #2296, opened 2025-11-10
Problem with unmounting remote filesystem after backup / command "status" involved


etyurin:
### Describe the problem, feature or ask a question:

This is with the version backintime 1.6.0-dev.9b8291cf, which is the most up-to-date version available in ArchLinux AUR archive.

My backup destination is mounted using sshfs. If I just open backintime and do not do a backup, there is no error. I get clean mount/unmount. However, if I start a backup, it never finishes, because of unmounting problems:

```

``
INFO: Released suspend mode inhibition
Traceback (most recent call last):
  File "/usr/share/backintime/common/backintime.py", line 173, in <module>
    startApp()
    ~~~~~~~~^^
  File "/usr/share/backintime/common/backintime.py", line 155, in startApp
    args.func(args)
    ~~~~~~~~~^^^^^^
  File "/usr/share/backintime/common/clicommands.py", line 110, in backup
    _do_backup(args, force)
    ~~~~~~~~~~^^^^^^^^^^^^^
  File "/usr/share/backintime/common/clicommands.py", line 132, in _do_backup
    ret = snapshots.Snapshots(cfg).backup(force)
  File "/usr/share/backintime/common/snapshots.py", line 1058, in backup
    .umount(self.config.current_hash_id)
     ~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/share/backintime/common/mount.py", line 269, in umount
    backend.umount()
    ~~~~~~~~~~~~~~^^
  File "/usr/share/backintime/common/mount.py", line 597, in umount
    self.removeSymlink()
    ~~~~~~~~~~~~~~~~~~^^
  File "/usr/share/backintime/common/mount.py", line 1070, in removeSymlink
    os.remove(self.config.snapshotsPath(
    ~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^
        profile_id=profile_id,
        ^^^^^^^^^^^^^^^^^^^^^^
        mode=self.mode,
        ^^^^^^^^^^^^^^^
        tmp_mount=tmp_mount))
        ^^^^^^^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: '/home/xxxxxxxxx/.local/share/backintime/mnt/1_5011'
``

```

buhtz:
Hello etyurin,

Thank you for taking the time to report the bug and providing the
details. I appreciate your feedback, will investigate the issue,
[...]

PR #2019, opened 2025-01-25
feat(cli): Status information about all backup profiles

s4moore:
Implemented the 'status' command to provide a snapshot of the status of all profiles. 

options:
  -h, --help            show this help message and exit
  --profile NAME        a more detailed summary of the profile with the name, NAME
  --profile-id ID       a more detailed summary of the profile with the id, ID
  --issues              show only profiles with errors on most recent run or no snapshot history at all
  --human-readable, -r  print in human readable format

Consider adding a CHANGELOG entry.
Run "codespell" to check for typos.

Close #1018

PR #2299, opened 2025-11-12
fix: Error but not crash while unmount if symlink is missing

buhtz:
While unmounting catch FileNotFoundException about missing symlink while unmounting and just log it as error.

It is a workaround to ease the burden with a still unknown bug introduced after 1.5.6 release. While unmounting (in SSH) symlink is deleted several times causing FileNotFoundException.

Diagnosis of the real problem was not successful. The code is complex, smelly and hard to debug. Further investigation and refactoring is planed.

Related to #2296
Follow-up issue #2321


buhtz:
I'll wait for [response from Samuel](https://github.com/bit-team/backintime/issues/2296#issuecomment-3522722124) ([second try](https://github.com/bit-team/backintime/pull/2019#issuecomment-3626148900)) before merging anything.

buhtz:
> I'll wait for [response from Samuel](https://github.com/bit-team/backintime/issues/2296#issuecomment-3522722124) ([second try](https://github.com/bit-team/backintime/pull/2019#issuecomment-3626148900)) before merging anything.

Still have no response from him. When it is time for the next release I think I'll remove/disable the `--status` feature until we find a solution.

- #2321

PR #2309, opened 2025-11-29
chore: Modify list of default excludes

### LINKED ISSUE #2307: Issue with Multiple periods in file name during backup

I recently installed BackInTime, from my Linux Mint 22.2 package repository, and tried creating a snapshot of a folder from one local drive to another local drive. The backup snapshot appeared to work, but I noticed that two of the seven files were not in the backup file directory and they did not appear in the log either. 

The two files that did not get copied had more than one period in the file name. These were the ones that did **not** get backed up. 

    LastPass.backup.kdbx
    NewDatabase.backup.kdbx

I was not able to find this as an existing issue.

The output from my diagnostics command follows

```json
{
    "backintime": {
        "name": "Back In Time",
        "version": "1.4.3",
        "latest-config-version": 6,
        "running-as-root": false,
        "keyring-supported": true
    },
    "host-setup": {
        "platform": "Linux-6.14.0-36-generic-x86_64-with-glibc2.39",
        "system": "Linux #36~24.04.1-Ubuntu SMP PREEMPT_DYNAMIC Wed Oct 15 15:45:17 UTC 2",
        "OS": {
            "/etc/os-release": "Linux Mint 22.2",
            "/etc/lsb-release": "DISTRIB_ID=LinuxMint\nDISTRIB_RELEASE=22.2\nDISTRIB_CODENAME=zara\nDISTRIB_DESCRIPTION=\"Linux Mint 22.2 Zara\"\n",
            "/etc/debian_version": "trixie/sid\n"
        },
        "display-system": "x11",
        "locale": "en_US, UTF-8",
        "PATH": "/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin",
    },
    "python-setup": {
        "python": "3.12.3 main Nov  6 2025 13:44:16 CPython GCC 13.3.0",
        "sys.path": [
            "/usr/lib/python312.zip",
            "/usr/lib/python3.12",
            "/usr/local/lib/python3.12/dist-packages",
            "/usr/lib/python3/dist-packages",
            "/usr/lib/python3.12/dist-packages"
        ],
        "qt": {
            "Version": "PyQt 5.15.10 / Qt 5.15.13",
            "Theme": "Mint-Y-Purple",
            "Theme Search Paths": [
                "/usr/share/cinnamon/icons",
                "/home/UsernameReplaced/.local/share/flatpak/exports/share/icons",
                "/usr/share/icons",
                ":/icons"
            ],
            "Fallback Theme": "",
            "Fallback Search Paths": []
        }
    },
    "external-programs": {
        "rsync": {
            "version": "3.2.7",
            "protocol": "31.0",
            "capabilities": {
                "file_bits": 64,
                "inum_bits": 64,
                "timestamp_bits": 64,
                "long_int_bits": 64,
                "socketpairs": true,
                "symlinks": true,
                "symtimes": true,
                "hardlinks": true,
                "hardlink_specials": true,
                "hardlink_symlinks": true,
                "IPv6": true,
                "atimes": true,
                "batchfiles": true,
                "inplace": true,
                "append": true,
                "ACLs": true,
                "xattrs": true,
                "secluded_args": "optional",
                "iconv": true,
                "prealloc": true,
                "stop_at": true,
                "crtimes": false
            },
            "optimizations": {
                "SIMD_roll": true,
                "asm_roll": false,
                "openssl_crypto": true,
                "asm_MD5": false
            },
            "checksum_list": [
                "xxh128",
                "xxh3",
                "xxh64",
                "md5",
                "md4",
                "sha1",
                "none"
            ],
            "compress_list": [
                "zstd",
                "lz4",
                "zlibx",
                "zlib",
                "none"
            ],
            "daemon_auth_list": [
                "sha512",
                "sha256",
                "sha1",
                "md5",
                "md4"
            ]
        },
        "ssh": "OpenSSH_9.6p1 Ubuntu-3ubuntu13.14, OpenSSL 3.0.13 30 Jan 2024",
        "sshfs": "(no sshfs)",
        "encfs": "(no encfs)",
        "shell": "/bin/bash",
        "shell-version": "GNU bash, version 5.2.21(1)-release (x86_64-pc-linux-gnu)"
    }
}
```

buhtz:
Remove and add some entries to the list of default excludes.
Improved the in-code documentation.
The PR is relevant only for new created backup profiles.

Close #2307


buhtz:
- The default selection is different between user and root mode.
- Typos?
- Wrong descriptions?
- Possible improvements in descriptions?
- Missing items?
- Opinion about default selection?
- Something else?

## Default selection of suggested excludes in user mode (left) and root mode (right)

image

## This is how to reach that extra dialog
Peek 2025-12-13 13-45

DerekVeit:
I have to say that I think this is a misstep.  This UI looks like it's being very helpful, categorizing and describing things, and giving options and suggestions, but overall it's burdening the user with a lot more to take in, to make sense of, and to make many little decisions about.  And on the development side, this seems like something that will forever need maintenance with more little considerations about what to add or remove and which things to give the default check marks and for which type of user, and even how to categorize each path.

I'm already confused myself by seeing things like /proc/* listed as unchecked for the non-root user.

The descriptions and any special guidance would be good to have in documentation, and for the GUI it might be helpful to provide a link or a Help button that takes the user right to it.

buhtz:
Hello Derek,

thank you for the feedback. Before I address your individual comments let me first describe the original problem I tried to solve.

The original default list did not fit into all use cases and for all users. That is not a surprise. Some entries are relevant for specific users other elements don't. And for some users the meaning of the entries are unknown (e.g. /proc). I can imagine that non-technical users are confused by the default entries and their meanings. They might be overwhelmed by the exclude list.

The concept behind a new solution is to have a minimal (nearly all fitting) list of pre-selected defaults and some additional not preselected suggestions. And to encourage and enpower the user to decide itself all entries need to be explained.

> This UI looks like it's being very helpful, categorizing and describing things, and giving options and suggestions, but overall it's burdening the user with a lot more to take in

I updated the previous comment with a screencast to make clear how to reach the suggestions dialog. In fact it is "hidden" like an "Export options" or "Advanced options" dialog. There is no need to use it at all. No need, no burden. It is optional.

> And on the development side, this seems like something that will forever need maintenance with more little considerations about what to add or remove and which things to give the default check marks and for which type of user, and even how to categorize each path.

That is pretty easy. It is a dict; see [`bitbase.EXCLUDE_SUGGESTIONS`](https://github.com/buhtz/backintime/blob/0e393bf0bbd8d91a63dae483227a7d193b81c97a/common/bitbase.py#L186-L256) in `common/bitbase.py`. Very simple. No need to touch the dialog or widget code. Very less maintenance costs.

But of course there is a need to make decisions about which entries to add and if they should be selected by default. That is why I am asking the community.

> I'm already confused myself by seeing things like /proc/* listed as unchecked for the non-root user.

Can you be more specific about that?
I would say that a regular user does not think about to backup something outside its own home folder. If there are users thinking outside of their home folder they should have enough expertise to decide themself what to include or exclude without the help of the "suggest excludes" dialog.

The default list is not to cover all cases, just the most common ones.

> The descriptions and any special guidance would be good to have in documentation

The GUI documents itself. That is the final goal. BIT always targeted low skilled desktop users. And an incremental backup is also a simple task. Simple users and simple tasks. In a perfect world (with a perfect GUI) there would be no need for a user manual.

The topic of exclude entries is IMHO way to simple to need to be covered by a user manual.

To take the discussion one level up. IMHO it should have be thought about if an application like BIT even need default entries in the exclude list. I would have never implemented this myself. But now it is there and I try to deal with it without removing everything.

Please let me know what you think about the specific entries and feel free to suggest modifications.

Regards,
Christian

DerekVeit:
I think three things seem not so good to me about using the checkboxes for these suggestions.

1. It treats each path as a distinct feature when it's just another value in an open-ended list.
2. Providing a checkbox communicates that it's worth considering as an option.  That's reasonable for some things like *~ but a strange suggestion, I think, for some others like /sys or /proc.
3. A checkbox feels very much like a positive thing even though it's in a dialog about exclusions, so it seems easier for someone to get confused and forget that "☑ /proc" means to _not_ backup that path.

But some good things about showing them as you have:

1. It allows providing the user some addition paths to easily add without making them defaults.
2. The choices are clearly presented.
3. Removals are more easily reversed than re-adding a removed path in the main dialog.

Maybe a dual list could be a good alternative way to do this:
image

This would treat the paths as values in a list.  The left list could be titled "Exclude:".  You would start with the defaults in the chosen list on the left and the suggestions on the right.  Maybe the defaults and recommended could be marked or styled distinctly from the opt-in suggestions.  Or changes from the default could be made bold (like in Firefox about:config).  This could easily communicate the bias of what is _normally_ in the list, unlike a checkbox, which is indifferent.

Maybe you could put the descriptions in hover text of each item.

The dual list might be more troublesome to implement, though.  What you have proposed isn't terrible.

buhtz:
Hello Derek,

thank you for the feedback.

>     1. It treats each path as a distinct feature when it's just another value in an open-ended list.

This point I don't understand.

>     2. Providing a checkbox communicates that it's worth considering as an option.  That's reasonable for some things like *~ but a strange suggestion, I think, for some others like /sys or /proc.

What is the difference between `*~` and `/sys` in this case?

>     3. A checkbox feels very much like a positive thing even though it's in a dialog about exclusions

Yes, that makes sense. I agree to this point.

> Maybe a dual list

That is a nice approach. Never had this kind of GUI element in mind. But I am not sure about all details about integrating it into this use case.

I would keep the current solution because it is an improvement. But I also treat it as sub-optimal. It is on my todo list to redesign the Exclude TAB and some other things of the Manage profiles dialog. Maybe then I will get a more clear idea how to improve the exlude suggestions and how to make use of a dual list in this case.

> Maybe you could put the descriptions in hover text of each item.

This is IMHO to much "hidden".

> The dual list might be more troublesome to implement

Just some work. Technically it is not complex.

Regards,
Christian

CBlakeB:
I am the person who originally opened this issue. I am new to BIT, and I really like it so far. The *.backup* default exclusion was keeping some of my KeepassXC backup files from getting backed up.  Not a huge issue, but they are pretty important files. 

I only use BIT to backup my /home directory.

I like the checkbox approach because it gives some good information about the type of files that will be excluded, and I think it would be a little less daunting for new or less experienced users.   

I have been wanting to experiment with the exclusions, because it is currently backing up some files which don't really need to be backed up. 

Best Wishes,
Blake

buhtz:
Hello Blake

thank you for your feedback.

> I have been wanting to experiment with the exclusions, because it is currently backing up some files which don't really need to be backed up.

Let me know if you have ideas for additional entries for the exclude suggestions dialog.

Regards,
Christian

Issue #1579, opened 2023-12-07
Snapshot log view: Add help for rsync itemize-changes 11-char output


aryoda:
## Current situation

The `Snapshot Log View` and `Last Log View` do contain the cryptic 11-char output of `rsync` to indicate the reason for an `rsync` action, eg.:

![image](https://github.com/bit-team/backintime/assets/11374410/b2b6fe71-0b03-46bd-af44-033bce3a97d0)

The meaning of `f>..t......` is unclear to normal users.

The meaning of the possible values at each char position is explained in `man rsync` in the `--itemize-changes` section
and in a more compact form in this SO post: https://stackoverflow.com/a/36851784

## Proposed feature

Add

- either a **"Rsync change tracking help" button** to briefly explain the possible values

- or provide some kind of **"hoover over"** functionality that shows kinda "intellisense" help for the 11 chars of the current line the mouse cursor is positioned.

  Alternatively a right-click context menu could work instead...

- or add both above proposals

evanjhoward11:
Hello, I would like to take a crack at this issue. Can you please assign it to me? A couple of questions:
1. Which of the 3 proposed features should I implement?
2. Is [qt/logviewdialog.py](https://github.com/bit-team/backintime/blob/dev/qt/logviewdialog.py) the correct file to make the changes in?

buhtz:
Hello evanjhoward11,

>  please assign it to me

It seems you have not read any of the information intended to be for new contributors, so 
please read this FAQ entries:
 * [Can you assign this to me?](#can-you-assign-this-to-me)
 * [Can I use @ mentions freely in issues or PRs?](#can-i-use--mentions-freely-in-issues-or-prs)
 * [Can I contribute without using the software?](#can-i-contribute-without-using-the-software)
 * [Can I boost my commit count?](#can-i-boost-my-commit-count)
 * [Can I submit AI-generated contributions?](#can-i-submit-ai-generated-contributions)

Provide a PR as minimal as possible. Show me something we can talk about.

evanjhoward11:
My apologies. I created PR #2325 for this issue. Please review it and let me know any areas I can improve.

buhtz:
Hello evanjhoward11,

why do you apologize?

I think about to re-write the CONTRIBUTING.md. Before providing a first PR to a new project a contributor should somehow introduce itself and describe its skills and motivation. This is very important in the current days where we have a lot of AI-crap and star-collecting folks spamming FOSS projects with problematic PRs.

[...]

PR #2325, opened 2025-12-29
Add help for rsync #1579

evanjhoward11:
This PR resolves issue [#1579](https://github.com/bit-team/backintime/issues/1579) by creating a help button in [qt/logviewdialog.py](https://github.com/bit-team/backintime/blob/dev/qt/logviewdialog.py) that explains the meaning of the different symbols used to describe rsync changes. When "Help" is clicked, a new window pops up with the info.

All ruff checks pass
1 pytest check fails for the following reason:
logviewdialog.py:305:0: R0903: Too few public methods (1/2) (too-few-public-methods)


buhtz:
Please see my comments in #1579.

Issue #2262, opened 2025-09-16
Investigate how and if BIT is using default icons with IceWM and similar (was: Icon dependencies not fully listed in CONTRIBUTING.md)


samo-sk:
### Describe the problem, feature or ask a question:

The file `qt/icon.py` lists multiple usable icon themes. `CONTRIBUTING.md` only lists Oxygen. All icon themes that fully work should be listed in `CONTRIBUTING.md`.

Also, in Debian, the older version that is currently in the package archives doesn't list icon themes as dependecies at all (not even in `Recommends`), which leads to non-fatal errors if one tries to launch BiT without installing a supported theme. I would be glad if (once the next release gets uploaded to Debian) the (upstream) maintainers checked if Debian lists the icon-theme dependencies correctly.

buhtz:
Hi Samo,
there is a misunderstanding. The oxygen dependency in CONTRIBUTING.md shouldn't be there. I'll fix that (b520a43).

There is no dependency to any icon theme. BIT will use the icons provided by the desktop environment. So the dependency managed by Debian are valid.

And the icon.py is a bit awkward. There is much work to do on it.

If you have other issues with Debians packaging you should open a bug report there.

Regards,
Christian

samo-sk:
OK, I will (maybe) report the missing-icons errors to Debian. (To clarify, these non-fatal errors happen when one uses IceWM instead of a full DE.)

buhtz:
I'll consider IceWM in my next tests. Thanks for bringing this up.

I need to investigate that. IMHO even with IceWM there should be some default icons to be installed. I mean if I install a window manager on Debian it should come with icons. But maybe I am wrong at this point. BIT only use default icons, except for "show-hidden" because no default exists for it.

EDIT: Note to myself. https://www.debian.org/doc/debian-policy/ch-opersys.html#menus

buhtz:
Debian 13 (with several DEs installed) and IceWM.

The application menu use the regular application logo

<img width="678" height="499" alt="Image" src="https://github.com/user-attachments/assets/83959218-ba91-4db9-ad2c-2dcb667329fc" />

Toolbar icons also available and fine looking.
I see no problem with the symbols/icons.

But I found another problem using BIT in root mode on IceWM if no polkit agent is active. PR #2328

samo-sk:
[...]

PR #2328, opened 2025-12-31
fix: Proper error message regarding pkexec problems

buhtz:
Using "Back In Time (root mode)" in desktop environments or window managers without an active polkit agent, now gives a proper message to the user via xmessage.

In the past this failed silently leaving the user without information.

Related #2262

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -51,6 +51,7 @@ Version 1.6.0-dev (development of upcoming release)
 * Removed: UI translation in Bosnian, Thai, and Occidental/Interlingue (#1914)
 * Removed: LICENSE file in favor of LICENSES directory and LICENSES.md file
 * Removed: SSH Cipher configuration in Manage profiles dialog (#2143)
+* Fixed: Show proper message to users when root mode fails due to inactive or missing polkit agent
 * Fixed: Enforce UTF-8 encoding in takesnapshot.log and some other files (#2298)
 * Fixed: Attribute error about missing 'cbCopyUnsafeLinks' (#2279)
 * Fixed: Use LC_ALL instead of LC_TIME to set the locale
--- a/CHANGES
+++ b/CHANGES
@@ -51,7 +51,7 @@ Version 1.6.0-dev (development of upcoming release)
 * Removed: UI translation in Bosnian, Thai, and Occidental/Interlingue (#1914)
 * Removed: LICENSE file in favor of LICENSES directory and LICENSES.md file
 * Removed: SSH Cipher configuration in Manage profiles dialog (#2143)
-* Fixed: Show proper message to users when root mode fails due to inactive or missing polkit agent
+* Fixed: Show proper message to users when root mode fails due to inactive or missing polkit agent (#2328, Derek Veit @DerekVeit)
 * Fixed: Enforce UTF-8 encoding in takesnapshot.log and some other files (#2298)
 * Fixed: Attribute error about missing 'cbCopyUnsafeLinks' (#2279)
 * Fixed: Use LC_ALL instead of LC_TIME to set the locale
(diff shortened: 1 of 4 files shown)


DerekVeit:
ERR will be getting the captured output of **backintime-qt** too, so that any output from the application would also cause the response.  So maybe it's better to check the exit code.  From the **pkexec** man page:

> Upon successful completion, the return value is the return value of PROGRAM. If the calling process is not authorized or an authorization could not be obtained through authentication or an error occured, pkexec exits with a return value of 127. If the authorization could not be obtained because the user dismissed the authentication dialog, pkexec exits with a return value of 126.

You can change the condition to `[ $STATUS -eq 127 ]`.

But then you also don't want to leave the application mute (all of its output captured and hidden) in the normal case.  So instead of capturing the output with $(), you could tee it to a temporary file:

```sh
pkexec --disable-internal-agent $PREFIX "/usr/bin/backintime-qt" "$@" 2>&1 | tee /tmp/backintime-output

STATUS=$?

if [ $STATUS -eq 127 ]]; then
    ERR=`cat /tmp/backintime-output`
    xmessage -nearmouse "Back In Time: Can't start in root mode.
    Error: $ERR
    Return code: $STATUS"
    exit $STATUS
fi
```

But this still does the wrong thing of merging the stdout and stderr outputs of the application.  So to capture just the stderr in the temporary file and keep the output and error streams from the application distinct:

```sh
{ { pkexec --disable-internal-agent $PREFIX "/usr/bin/backintime-qt" "$@" 2>&1 1>&3; printf $? >/tmp/backintime-exitcode; } | tee /tmp/backintime-error >&2; } 3>&1

STATUS=`cat /tmp/backintime-exitcode`

if [ $STATUS -eq 127 ]]; then
    ERR=`cat /tmp/backintime-error`
    xmessage -nearmouse "Back In Time: Can't start in root mode.
    Error: $ERR
    Return code: $STATUS"
    exit $STATUS
fi
```
This diverts the stdout (1) output through file descriptor 3 while stderr (2) is directed to 1 to go into tee.  The exit code is written to a file, because $? then gets set by tee.

This could be a little simpler if it were strictly for Bash, but the shebang line is for /bin/sh.

buhtz:
Hello Derik,
happy new year. Thanks for checking this out. I assumed that there is more behind it to take care of. :smile: I'll check it out.

Regards,
Christian

buhtz:
Thanks a lot for checking on this.
Your original code worked. I tested it.

After that I modified the shebang to Bash and **ask an LLM** (:biohazard:) to convert the code into Bash style. I also try to catch all errors. In my limited knowledge of pkexec it is not always 127. This will also cover errors by BIT itself.

image

Tests with icewm and xfce are fine. What do you think?

DerekVeit:
That's mostly the simpler version I had in mind for Bash.  Also, with Bash, you can replace `[[ ${STATUS} -ne 0 ]]` with `(( STATUS ))`.  The (()) is a numeric condition and doesn't require the $ prefix on the variable.  A non-zero STATUS will be truthy as in Python.

buhtz:
I realized that my current version would create a new dependency; to `bash`. Not sure yet but it is OK for me. I prefer an easy to understand Bash script instead of that sh-script that no one else would understand or be able to maintain.

What is your opinion on that?

DerekVeit:
Bash is common enough, and improves the code enough, that I prefer depending on it.  For a desktop GUI app, it doesn't seem to me like such a heavy dependency, and I would guess that most users of Back in Time will already have it.

Otherwise, I think the sh version is maintainable too.  It's not as exotic as it might appear if you're not used to shell redirections.  I would be glad to give a more detailed explanation, but I think nowadays one could also ask an LLM chatbot to explain it.  That being said, I prefer the Bash version, because it more directly expresses the intent.

buhtz:
Thank you for your feedback and help. So, then I'll keep the bash-dialect version and merge it soon.

Issue #1836, opened 2024-08-09
root mode: No toolbar symbols (Manjaro Arch)


buhtz:
BIT's log output told me to ask here. :rofl: 
![image](https://github.com/user-attachments/assets/e9cbcc93-b7a5-4c0c-8da2-ec1a391e7a9b)

I am using 1.5.3-dev on Manjaro Linux (Arch based). BIT as regular user looks fine. BIT as root (via poolkit) do not show me the icons in the toolbar.

Couldn't find something about this in FAQ or known problems section. Didn't we had this before somehow?

Side question: Why does "running-as-root" in the debug and in the diagnostics output show `False`?

### Full debug output
```
$ backintime-qt_polkit --debug
DEBUG: [common/backintime.py:613 argParse] Used argument(s): {'debug': True, 'quiet': False}
DEBUG: [common/backintime.py:614 argParse] Unknown argument(s): []
DEBUG: [common/backintime.py:513 startApp] backintime: {'name': 'Back In Time', 'version': '1.5.3-dev.59f0acab', 'running-as-root': False}
DEBUG: [common/backintime.py:513 startApp] host-setup: {'OS': {'/etc/os-release': 'Manjaro Linux', '/etc/arch-release': 'Manjaro Linux\n', '/etc/manjaro-release': 'Manjaro Linux\n', '/etc/lsb-release': 'DISTRIB_ID="ManjaroLinux"\nDISTRIB_RELEASE="24.0.1"\nDISTRIB_CODENAME="Wynsdey"\nDISTRIB_DESCRIPTION="Manjaro Linux"\n'}}

Back In Time
Version: 1.5.3-dev.59f0acab

Back In Time comes with ABSOLUTELY NO WARRANTY.
This is free software, and you are welcome to redistribute it
under certain conditions; type `backintime --license' for details.

DEBUG: [common/configfile.py:591 Config.setCurrentProfile] Change current profile: 1=Hauptprofil
DEBUG: [common/tools.py:223 initiate_translation] Language code "en".
DEBUG: [common/tools.py:274 set_lc_time_by_language_code] Try to set locale.LC_TIME to "en_US.UTF-8" based on language code "en".
DEBUG: [common/backintime.py:696 getConfig] config file: /root/.config/backintime/config
DEBUG: [common/backintime.py:697 getConfig] share path: /root/.local/share/backintime
DEBUG: [common/backintime.py:698 getConfig] profiles: 1=Main profile
[...]

PR #2354, opened 2026-01-10
fix(qt): Optimize list of icon themes to probe

buhtz:
This might solve some problems of toolbar icons missing in root mode. For example #1836.

Reported in https://github.com/bit-team/backintime/pull/2345#issuecomment-3732320045 by @gamerdad84.

### DIFF
--- a/qt/icon.py
+++ b/qt/icon.py
@@ -17,14 +17,13 @@ import logger
 # If the current theme does not even contain the "document-save" icon
 # try to use another well-known working theme (if it is installed):
 themes_to_try = (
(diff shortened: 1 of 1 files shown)


buhtz:
In context of #1836: I was not able to make my Manjaro VM run again. So I switched to EndevourOS (also Arch-based) with XFCE4. There I can confirm that the icons not longer missing in root mode. Adwaita is used in root-mode.

Another observation on that test setup was that the rood mode does ignore the DEs dark mode. But I can live with that for now.

PR #2365, opened 2026-01-16
downgrade the missing-icons error message to a warning

samo-sk:
As suggested in https://github.com/bit-team/backintime/issues/2262#issuecomment-3761124721. I haven't added a changelog entry; I am not sure this minor change needs it.


buhtz (approved):
Thank you very much. That is fine this way.
I also think no need of a changelog entry.

PR #2378, opened 2026-01-22
chore(qt): Robust screen detection on Wayland/Hyperland and X11

### LINKED ISSUE #2373: No "screen" instance: (was: "AttributeError: 'NoneType' object has no attribute 'availableGeometry'" crash (Arch Linux)")

On Arch Linux, Wayland. Crash on initial launch:

```
backintime-qt

Back In Time
Version: 1.6.0-rc1

Back In Time comes with ABSOLUTELY NO WARRANTY.
This is free software, and you are welcome to redistribute it
under certain conditions; type `backintime --license' for details.


  File "/usr/share/backintime/qt/usermessagedialog.py", line 89, in _center
    self._center_to_screen()
    ~~~~~~~~~~~~~~~~~~~~~~^^
  File "/usr/share/backintime/qt/usermessagedialog.py", line 100, in _center_to_screen
    geom = screen.availableGeometry()
           ^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'availableGeometry'
/usr/bin/backintime-qt: line 26:   152 Aborted                    (core dumped) /usr/bin/python -Es ${APP_PATH}/app.py "$@"
```

The installed package is https://aur.archlinux.org/packages/backintime-git and I replaced the PKGBUILD file with this newer unpublished version:  [PKGBUILD](https://github.com/user-attachments/files/24790426/PKGBUILD.txt). But it also crashes without the modified PKGBUILD file.

Testing it in a minimal container environment, see [backintime-diagnostics.txt](https://github.com/user-attachments/files/24790513/backintime-diagnostics.txt). But the same happens on a normal Arch Linux system.

buhtz:
Fix #2373

@willemw12 Can you check this fix before I merge it?

### DIFF
--- a/qt/qttools.py
+++ b/qt/qttools.py
@@ -463,7 +463,8 @@ def screen_width_in_chars(widget: QWidget, reference_char: str = 'M') -> int:
     char_px = metrics.horizontalAdvance(reference_char)
 
     # Screen width
-    screen = QGuiApplication.screenAt(widget.pos())
+    handle = widget.windowHandle()
+    screen = handle.screen() if handle else QGuiApplication.primaryScreen()
     geom = screen.availableGeometry()
 
     # Screen width in 'em' (number of characters)
--- a/qt/restoreconfigdialog.py
+++ b/qt/restoreconfigdialog.py
@@ -150,7 +150,8 @@ class RestoreConfigDialog(QDialog):
     def _resize_to_full_hight(self):
         """Resize dialog to full height and center it horizontal.
         """
-        screen = QGuiApplication.screenAt(self.pos())
+        handle = self.windowHandle()
+        screen = handle.screen() if handle else QGuiApplication.primaryScreen()
         geom = screen.availableGeometry()
 
         # Determine the height of the dialog's title bar and border. This
--- a/qt/restoreconfigdialog.py
+++ b/qt/restoreconfigdialog.py
@@ -27,7 +27,6 @@ from snapshots import SID
 from PyQt6.QtGui import (QBrush,
                          QColor,
                          QFont,
-                         QGuiApplication,
                          QFileSystemModel,
                          QPalette,
                          QShortcut)
@@ -38,6 +37,7 @@ from PyQt6.QtWidgets import (QDialog,
                              QLabel,
                              QLayout,
                              QPushButton,
+                             QSizePolicy,
                              QToolButton,
                              QTreeView,
                              QVBoxLayout,
@@ -73,16 +73,25 @@ class RestoreConfigDialog(QDialog):
         self.setWindowIcon(icon.SETTINGS_DIALOG)
         self.setWindowTitle(_('Import configuration'))
 
-        layout = QVBoxLayout(self)
+        self.setSizePolicy(
+            QSizePolicy.Policy.Preferred,
+            QSizePolicy.Policy.Expanding
+        )
 
-        self._create_hint(layout, config)
+        main_layout = QVBoxLayout(self)
+
+        top_layout = QVBoxLayout()
+        self._create_hint(top_layout, config)
         self._lbl_spinner, self._spinner, self._btn_scan \
-            = self._create_scan_controls(layout)
+            = self._create_scan_controls(top_layout)
 
         self._btn_scan.clicked.connect(self.start_scanning)
 
+        main_layout.addLayout(top_layout, 0)
         self._tree_view, self._tree_model = self._create_tree()
-        layout.addWidget(self._tree_view)
+        tree_layout = QVBoxLayout()
+        tree_layout.addWidget(self._tree_view)
+        main_layout.addLayout(tree_layout, 1)
 
         # expand users home
         self._expand_with_parents(self._index_from_path(Path.home()))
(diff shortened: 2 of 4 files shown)


willemw12:
Checked. Fixed.

A GUI issue I noticed: the "import" screen on initial launch on Hyprland is unusable ("corrupt" screen), until I toggled the "floating/tiling" property of that window (with a key bind/shortcut). The file tree widget was separate from the main import screen.The other screens I checked were a fine.

buhtz:
Thank you for checking this.

Abou the import window issue. Do you mean this window?

image

Is it possible to reproduce this and give me a screenshot? Or a screen photo?

I still struggle to setup a an Arch VM with full graphic desktop, to reproduce this. Which desktop environment are you using?

I might try one of these pre-build VirtualBox images: https://www.osboxes.org/arch-linux/

EDIT: The RestoreConfigDialog maximize its height (vertical), set its width to 1/3 of the screen and centers horizontal.

willemw12:
Is difficult to capture the mixed overlayed windows in a single screenshot.
But what you see is both screenshot1 and screenshot2 at the same time. When you move the mouse cursor, then it flickers between them. It is as if they are competing for the same GUI resource or (widget) index, maybe.

To test, you can also:

- Install Hyprland (won't conflict with other desktops) and Kitty (the default terminal emulator)
- Login into the Hyrland session
- Press "Super/Window key + q" to launch Kitty

Then run from the terminal:

- For starting for the first time: `rm -rf ~/.config/backintime ~/.local/share/backintime`
- Then: `backintime-qt`

screenshot1

screenshot2

buhtz:
Mhm... Hyprland doesn't seem to be available on Debian 13. I tried i3 as an alternative tilling WM. There it works as expected. I mean I can not reproduce the problematic behavior with i3.

The behavior you describe is that the dialog always try to resize itself to the 1/3 of the screen width but the WM try to expand to to the available space (half screen I am assuming because the other half is taken by the terminal emulator).

Not sure how to approach this problem.

**EDIT**: Can you do some debug printing for me please.

This is `qt/restoreocnfigdialog.py`:

```py
    def _resize_to_full_hight(self):
        """Resize dialog to full height and center it horizontal.
        """
        handle = self.windowHandle()
        screen = handle.screen() if handle else QGuiApplication.primaryScreen()
        geom = screen.availableGeometry()

        # Determine the height of the dialog's title bar and border. This
        # value is unknown or incorrect until the dialg is fully drawn.
        # That is the reason why we use this workaround.
        deco_height = self.frameGeometry().height() - self.geometry().height()
        if deco_height == 0 and self._resize_tries > 0:
            self._resize_tries -= 1
            QTimer.singleShot(1, self._resize_to_full_hight)
            return

```

Can you please add a simple print after that return. I want to see if this method is called multiple times. The behavior you describe indicates this but I can not see how it could be called multiple times.

```py
        if deco_height == 0 and self._resize_tries > 0:
            self._resize_tries -= 1
            QTimer.singleShot(1, self._resize_to_full_hight)
            return

        print('resize-to-full-hight')
```

willemw12:
You would need a recent distro and Hyprland version. See https://wiki.hypr.land/Getting-Started/Installation/

If you have hyprlauncher installed, you can press "Super/Window key + r" to launch BIT without a terminal emulator open.

i3wm runs on Xorg (X11), not Wayland. Does it work in GNOME or KDE on Wayland?

Are you doing something different with this dialog compared to other BIT dialogs? For example, the "Exclude Suggestions" dialog window behaves normally as a tiling window (not as a floating window). The "Import configuration" window seems to behave partially as as floating window (the window is taller than the available workspace below the panel/bar and seems to fill up the monitor height).


`resize-to-full-hight` is only printed once.


Maybe you can create a new issue for this,

buhtz:
Thank you for providing more details. Maybe it is a bit wired what I try to achieve in this dialog. Maybe I just shouldn't manipulate its size and position.

It is interesting that you experience the dialog higher than the screen itself. Something is wrong with this calculation.

willemw12:
Is the height of the dialog only determined by the number of folders in the folder tree widget? No max height and surrounding scroll widget?


The basic idea is that programs in Wayland should only know what is inside a window, nothing outside that (the window size or location on the desktop (geometry), at least not set it. And there is also XWayland.

But I heard about an extra Wayland protocol (extension) and maybe a separate Wayland portal (package) for allowing this. Each Window Compositor (in GNOME, in KDE, Hyprland, ...) have their own set of Wayland protocols and (sometimes shared) implementations.

buhtz:
> Is the height of the dialog only determined by the number of folders in the folder tree widget? No max height and surrounding scroll widget?

No. The screen size is used, the size of decorations (windows title bar) and the panel/taskbar are substracted. Works well on xfce.

But the effort isn't worth the effect. So I removed all that fancy resizing stuff and updated the PR.

Can you check it again please?

willemw12:
Fixed. Hyprland in normal (tiling) mode, works.

But in "tabbed" workspace mode (`windowrule = group set, match:modal off, match:workspace 1`), the main window is not a tabbed window, but fills the whole workspace. Maybe you are setting the size for the main window. Not really an important issue and can be ignored.

buhtz:
All this fancy window managers ... :rofl: 
Tabbed workspace? Never heard of that fancy stuff. :eyes: It seems Hyprland is my new edge case to put on the regular BIT testing protocol. ;)

The main window and some of the dialogs do remember their last position and size and do restore it. See state file `~/.local/state/backintime-qt.json`.

willemw12:
https://wiki.hypr.land/Getting-Started/Master-Tutorial/#vm

Usually those windows geometry "hints" that are saved by GNOME, KDE, etc. are ignored by tiling compositors (Hyprland, ...).

The Niri compositor has "infinite horizontal scrolling workspaces".

PR #2404, opened 2026-02-08
fix(restore): block restore while backup is running

### LINKED ISSUE #1589: Restore may be started while a snapshot of the same profile is taken
## Observation

While documenting the lock file usage of BiT in v1.4.1 (currently released version as at today) I saw
that `snapshot()` checks if a restore is running for the same profile to avoid conflicts (esp.
when restoring directly into the original folders of the snapshot files):

https://github.com/bit-team/backintime/blob/25c2115b42904ec4a4aee5ba1d73bd97cb5d8b31/common/snapshots.py#L719-L720

I could find not a check for a running backup ("take snapshot") in ´restore()` so IMHO it may happen that
- a new snapshot is taken in one process
- a restore for the same profile may be started then in another process

https://github.com/bit-team/backintime/blob/25c2115b42904ec4a4aee5ba1d73bd97cb5d8b31/common/snapshots.py#L449-L458

## Possible Impact

- TODO to be analyzed

PS: I think this may be good issue for new BiT developers (just test by starting two instances and then try to understand the source code of the `restore()` function...

Dodothereal:
## Summary
- check for a running backup instance at the start of `Snapshots.restore()`
- abort restore early with a diagnostic warning when backup is active
- add a regression test ensuring no restore rsync execution starts while a backup PID lock is present

## Testing
- `python3 -m py_compile common/snapshots.py common/test/test_restore.py`
- focused lock-check script to validate early-return behavior when backup PID file is active

Fixes #1589


Dodothereal:
Closing to keep contributions focused on the previously discussed issue only.

PR #2462, opened 2026-04-02
refactor: remote (SSH) not-encrypted profiles 

buhtz:
Remote (SSH) profile without encryption


buhtz:
```
refactor: remote (SSH) not-encrypted profiles

Re-write of remote SSH not-encrypted profiles

- 
```

PR #2554, opened 2026-07-31
fix(ssh): resolve hostname via ssh -G for known_hosts check

### LINKED ISSUE #2550: Make known_hosts check consistent with OpenSSH host name handling
🚫 AI-generated code submissions are not welcome. 🚫
⚠️ If you cannot demonstrate full authorship, understanding, and responsibility for your submitted content, your account might be reported to the GitHub/Microsoft abuse team without warning.

---

https://github.com/bit-team/backintime/blob/02ca168d601ce0f5ec1bd15dc7ee665136613d6c/common/sshsetupvalidator.py#L351

The SSH setup check currently uses the configured host name directly with `ssh-keygen -F` to determine whether a host key is already present in known_hosts.

Example:

    ssh-keygen -F FOOBAR

does not find an existing entry if known_hosts contains the same host as `foobar`. However, OpenSSH itself resolves the configured host name before connecting:

    ssh -G user@FOOBAR

returns:

    hostname foobar

and the actual SSH connection succeeds.

This means Back In Time can report a host as "not trusted" even though OpenSSH would accept the host key and the connection works.

The known_hosts lookup should use the effective host name resolved by OpenSSH instead of the raw configured host name. For example:

    # lookup the config
    ssh -G user@FOOBAR | grep hostname

This would also correctly handle SSH configuration features such as Host/HostName aliases.

Iqbalez:
## Summary
- Resolve the effective SSH hostname with `ssh -G` (same options as profile SSH) before `ssh-keygen -F` known_hosts lookups
- Keeps bracket/port host forms using the resolved hostname and port
- Adds unit tests with mocked `subprocess.run` for alias/case handling, failure paths, and `ssh -G` command construction

Fixes #2550

## Test plan
- [x] `python -m unittest test.test_sshsetupvalidator` (Linux CI; not runnable on Windows due to `os.geteuid` / `syslog` dependencies)
- [ ] Manual: configure `Host FOOBAR` + `HostName foobar`, ensure known_hosts entry for `foobar` passes SSH setup validation

Made with [Cursor](https://cursor.com)

### DIFF
--- a/common/sshsetupvalidator.py
+++ b/common/sshsetupvalidator.py
@@ -348,15 +348,63 @@ class SSHSetupValidator:  # pylint: disable=too-few-public-methods
                 + _('Details:') + f'\n{err.strip()}'
             )
 
+    def _build_ssh_g_command(self) -> list[str]:
+        """Build ``ssh -G`` command mirroring connection options."""
+        cmd = self._build_ssh_command()
+        return [cmd[0], '-G', *cmd[1:]]
+
+    @staticmethod
+    def _parse_ssh_g_output(output: str) -> tuple[str | None, int | None]:
+        hostname = None
+        port = None
+
+        for line in output.splitlines():
+            key, _, value = line.partition(' ')
+            value = value.strip()
+
+            if key == 'hostname':
+                hostname = value
+            elif key == 'port':
+                try:
+                    port = int(value)
+                except ValueError:
+                    pass
+
+        return hostname, port
+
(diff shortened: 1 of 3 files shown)


buhtz:
It seems you have not read any of the information intended to be for new contributors.
This pull request has been closed not only because it appears to contain AI-generated content but the contributor is obviously not willing and/or able to take responsibility for this content.

AI-based submissions cause significant issues in this project and consume valuable
resources from volunteer maintainers and developers.

You are reported to Microsoft's abuse team.

